PR #5, opened 2018-03-19
Import command line arguments from omp

diml:
I forgot to import this from ppx_driver.v0.10.3

### DIFF
--- a/src/driver.ml
+++ b/src/driver.ml
@@ -1122,11 +1122,27 @@ let standalone_args =
   ]
 ;;
 
+let get_args ?(standalone_args=standalone_args) () =
+  let args = standalone_args @ List.rev !args in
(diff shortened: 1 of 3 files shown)


src/driver.ml
```diff
@@ -1178,7 +1194,8 @@ let standalone_run_as_ppx_rewriter () =
     List.map standalone_args ~f:(fun (arg, spec, _doc) ->
       (arg, spec, " Unused with -as-ppx"))
   in
-  let args = List.rev_append !args standalone_args in
+  let args = get_args () ~standalone_args in
```
xclerc: (I would reverse the order of the arguments passed to `get_args`.)

diml: sure

PR #7, opened 2018-03-21
Remove references to type_conv

xclerc:
This PR removes some of the references to `type_conv` from the
code base. Some references are still present, either because they
are referencing the previous project or for backward compatibility.


diml:
The last references are the command line argument and the toplevel transformation name? For the command line argument, we could add new ones and mark the old ones deprecated, and for the transformation name, we could add the possibility to give aliases to `register_transformation`

xclerc:
> The last references are the command line argument and the toplevel transformation name?

Yes; there is also the name of the attribute (currently `"type_conv.deriving" ^ suffix`).
(There are other references in the `test/` directory, but they will be fixed in another PR.)

> For the command line argument, we could add new ones and mark the old ones deprecated, and for the transformation name, we could add the possibility to give aliases to register_transformation

Indeed, that's pretty much what I had in mind. Should we do these changes in the current PR?

diml:
> Indeed, that's pretty much what I had in mind. Should we do these changes in the current PR?

I think we can merge this PR now

PR #8, opened 2018-03-27
Update AST to 4.06

xclerc:
**Note: this is a partial update**
We do not import the changes related to [MPR#7334](https://caml.inria.fr/mantis/view.php?id=7344)
(i.e.  commits fd0dc6a0fbf73323c37a73ea7e8ffc150059d6ff
and c15ad73e56c54663c9cb6f7cac4241bb3c4e3cb8) as they
cause a number of problems, notably related to ppx extensions
(see *e.g.* [1] and [2] for the lwt variant of the issue).
We expect that a non-parser fix to MPR#7334 will be provided
in a future version of OCaml.

[1] https://github.com/ocsigen/lwt/issues/564
[2] https://caml.inria.fr/mantis/view.php?id=7758

### DIFF
--- a/ast/ast.ml
+++ b/ast/ast.ml
@@ -167,7 +167,7 @@ and core_type_desc = Parsetree.core_type_desc =
   | Ptyp_arrow of arg_label * core_type * core_type
   (* T1 -> T2       Simple
      ~l:T1 -> T2    Labelled
-     ?l:T1 -> T2    Otional
+     ?l:T1 -> T2    Optional
   *)
   | Ptyp_tuple of core_type list
   (* T1 * ... * Tn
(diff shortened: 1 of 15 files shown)


diml:
Ok, however we use `-as-ppx` for merlin, is that going to be an issue? I.e. rewriters working but failing when invoked from merlin. /cc @trefis

trefis:
I don't think it's going to be much of an issue.
IIUC the parsers will diverge for this specific construction, so the worst that can happen is that merlin will report a type-error that won't appear during compilation because the annotation isn't there anymore (or vice versa).

If it turns out to annoy people too much we can always modify merlin's parser so that when passed a flag it doesn't copy the annotation.

diml:
Ok.

> If it turns out to annoy people too much we can always modify merlin's parser so that when passed a flag it doesn't copy the annotation.

I have a feeling it's going to annoy people :)

PR #9, opened 2018-03-28
Remove references to type_conv (continued)

xclerc:
(Basically a follow-up to https://github.com/ocaml-ppx/ppxlib/pull/7.)

We might want to factor out some parts dealing with attributes/transforms,
but I am not sure it is worth the trouble.

### DIFF
--- a/src/deriving.ml
+++ b/src/deriving.ml
@@ -468,9 +468,9 @@ let parse_arguments l =
   with Unknown_syntax (loc, msg) ->
     Unknown_syntax (loc, msg)
 
-let mk_deriving_attr context ~suffix =
+let mk_deriving_attr context ~prefix ~suffix =
   Attribute.declare
-    ("type_conv.deriving" ^ suffix)
+    (prefix ^ "deriving" ^ suffix)
     context
     Ast_pattern.(
       let generator_name () =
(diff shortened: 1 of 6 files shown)


src/deriving.ml
```diff
@@ -8,25 +8,37 @@ let do_insert_unused_warning_attribute = ref false
...
+      (["code"; "attribute"],
+       (function
+         | "code"      -> do_insert_unused_warning_attribute := false
+         | "attribute" -> do_insert_unused_warning_attribute := true
+         | _           -> assert false))
+  in
+  Driver.add_arg "-deriving-keep-w32"
+    keep_w32_spec
     ~doc:" Do not try to disable warning 32 for the generated code";
+  Driver.add_arg "-deriving-w32"
```
diml: I find the name of this option not very intuitive, could we use something like `-deriving-disable-w32-method`?

I can't remember why I didn't ask about it when we added `-type-conv-w32`...

src/deriving.ml
```diff
@@ -617,6 +635,7 @@ let expand_sig_type_ext ~loc ~path te generators =
 
 let () =
   Driver.register_transformation "type_conv"
+    ~aliases:["deriving"]
```
diml: I would swap the name and alias, IMO the main name should be the preferred one.

src/deriving.ml
```diff
@@ -655,5 +674,43 @@ let () =
            ; Context_free.Rule.attr_sig_exception_expect
                Attr.Expect.ec
                expand_sig_exception
+
+           (* equivalent but deprecated attributes *)
```
diml: We can just get rid of those. The only two rewriters to interpret `[@@deriving]` attributes are ppxlib and ppx_deriving, and they don't need disambiguation.

PR #10, opened 2018-04-04
Do not use physical equality on strings.

hhugo:


xclerc:
For the record, tests pass with this patch on top of https://github.com/ocaml-ppx/ppxlib/pull/11.

diml:
Ok, that seems fine. I guess the worse that can happen is that a fragment of Ast gets relocated, but even in this case what we would get is a false positive. So I don't think there is a risk of accidentally not reporting an uninterpreted attribute,

PR #12, opened 2018-04-12
Update test expectations now that short paths are used

xclerc:


diml:
Thanks,  I forgot to promote the tests...

PR #13, opened 2018-04-12
Use "relative" line numbers in expect tests

xclerc:
The line numbers are now "relative" to the end of the previous
`%%expect` block (if any, to the beginning of the file otherwise).

It makes tests less fragile; *e.g* adding a test before another
one does no more trigger a renumbering.


diml:
What about also setting `pos_fname` to `//toplevel//`? The locations should be less verbose and not include line numbers

PR #16, opened 2018-04-18
Restore compatibility with OCaml 4.04.x

xclerc:
In order to restore compatibility with OCaml 4.04.x, this PR
imports a copy of `Buffer.add_utf_8_uchar`. It cannot be a
mere copy/paste because the original implementation is
based on the non-exported type definition of `Buffer.t`.

With this PR, ppxlib can be compiled with 4.04.x. However,
it is noteworthy that tests do not pass because the set of
enabled warnings was different back then (fixable), and
because the heuristic of `-short-path` was different (fixable
by explicitly opening some modules?).

### DIFF
--- a/ast/lexer.mll
+++ b/ast/lexer.mll
@@ -99,24 +99,25 @@ let keyword_table =
 ]
 
 module Buffer = struct (* Imported for compatibility with 4.04.x *)
-  include Buffer
+  [@@@warning "-32"]
   let add_utf_8_uchar b u = match Uchar.to_int u with
     | u when u < 0 -> assert false
     | u when u <= 0x007F ->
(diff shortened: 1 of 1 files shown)


ast/lexer.mll
```diff
@@ -98,6 +98,27 @@ let keyword_table =
     "asr", INFIXOP4("asr")
 ]
 
+module Buffer = struct (* Imported for compatibility with 4.04.x *)
+  include Buffer
```
diml: I think the following pattern is slightly better as it uses the upstream functions for newer version of OCaml:

```
[@@@warning "-32"]
let add_utf_8_uchar = ...
include Bufffer
```

xclerc: Sure

Issue #20, opened 2018-04-23
Question about linking actual rewriter with ocamlbuild


Kakadu:
There is a [command](https://github.com/ocaml-ppx/ppxlib#building-a-custom-driver-using-ocamlfind) to build custom rewriter using `ocamlfind` but I currently have some troubles making things work

The problem is that current approach relies on the fact that ppx plugin is built as findlib library. And if I'm developing syntax extension in my project and want to run some tests than I need to compile syntax extension and after that compile tests with adjusted OCAMLPATH variable. I tried to pass `cmxa` manually and remove dependency on findlib package but it seems that the order of `cmxa`'s matters and only way to do every all right is to reference PPX package. With this approach there is also an issue with dependencies: tests should be recompiled when the package changes and this currently doesn't work as I want.

Can you give me a piece of advice without recommending switching to `dune`?

P.S. This is a question mentioned in #18

diml:
You can try replacing `-package ppxlib.runner` by `my-rewriter.cmxa runner.cmx` where runner.ml contains:

```
Ppxlib.Driver.standalone ()
```

then things should be linked in the right order

Kakadu:
@diml It is very helpful. Do you think about adding it to the README?

diml:
that seems fine to me. Would you mind writing a PR? :)

PR #21, opened 2018-04-24
README: compiling rewriter that is not installed 

Kakadu:
Add compilation lines for your rewriter that is not yet installed as findlib package. 

That PR should resolve issue #20 

@diml

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,10 +146,8 @@ right things for you. This section is written having ocamlbuild in mind.
 When developing a new rewriter you are very likely to prepare a few
 tests for it. The compilation line above doesn't suit this task very well
 (because ocamlfind package with your rewriter is not yet installed) 
-and it will be more convenient to specify `.cm[x]a` with your rewriter
-manually. With this approach we sacrifice compilation speed using 
-`-ppx argument` option but it should not be critical because it's very
-likely that few rewriters will be used in the code of tests.
+and it will be more convenient to specify `.cmx[a]` with your rewriter
+manually.
 
 For example, let's suppose that the standalone rewriter (`pp_foo.native`) 
 have this code


README.md
```diff
@@ -274,6 +274,38 @@ package "runtime-lib" ( ... )
 You can check that this META works for all the 4 methods described
 above.
 
+### Building rewriter that you are currently developing
+When developing a new rewriter you are very likely to prepare a few
+tests for it. The compilation line above doesn't suit this task very well
+(because ocamlfind package with your rewriter is not yet installed) 
+and it will be more convenient to specify `.cm[x]a` with your rewriter
+manually. With this approach we sacrifice compilation speed using 
+`-ppx argument` option but it should not be critical because it's very
```
diml: I don't really follow this, where does the slowdown comes from?

README.md
```diff
@@ -274,6 +274,38 @@ package "runtime-lib" ( ... )
...
+
+For example, the standalone rewriter (`pp_foo.native`) will
+have this code
+
+    let () = Ppxlib.Driver.standalone ()
+
+in `pp_foo.ml` and will require a few extra switches to compile 
+(N.B. order matters)
+
+    ocamlfind ... ppx/ppx_foo.cmxa -package ppxlib
```
diml: It might be worth mentioning that `ppx_foo.cmxa` must be linked with `-linkall`, or alternatively use `ppx_foo.cmx`.

README.md
```diff
@@ -274,6 +274,38 @@ package "runtime-lib" ( ... )
 You can check that this META works for all the 4 methods described
 above.
 
+### Building rewriter that you are currently developing
+When developing a new rewriter you are very likely to prepare a few
```
diml: Shouldn't this section come after `Building a custom driver using ocamlfind`? For instance it's not clear to me what `the compilation line above` refers to.

diml:
BTW, doing a bit of advertisement for Dune: with Dune you can use an ocamlfind package that is not yet installed. i.e. if you are developing a package `foo`, you can have a rule that depends on package `foo` and use `ocamlfind` in the action. `ocamlfind` will see package `foo` as if it had been installed, this is very convenient for tests.

README.md
Kakadu: There were a claim that combining all PPXs into single rewriter works faster then using many -ppx compiler switches

Kakadu:
@diml Yeah, I heard about dune.....

README.md
diml: Indeed, however I don't see why this is related to building a rewriter this way given that you can still link external ppxlib based rewriters with the local one.

PR #27, opened 2018-05-08
Travis and appveyor

diml:
This PR sets up travis and appveyor using autoci (thanks @yomimono!).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: c
+install: wget https://raw.githubusercontent.com/ocaml/ocaml-ci-scripts/master/.travis-docker.sh
+script: bash -ex .travis-docker.sh
+services:
+- docker
+env:
+  global:
+  - PINS="ppxlib:."
+  - DISTRO="debian-stable"
+  matrix:
+  - PACKAGE="ppxlib" OCAML_VERSION="4.04.2"
+  - PACKAGE="ppxlib" OCAML_VERSION="4.05.0"
+  - PACKAGE="ppxlib" OCAML_VERSION="4.06.0"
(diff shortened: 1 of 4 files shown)


diml:
Ok, this seems to be working. Some of the tests are failing under older version of OCaml though

metaquot/ppxlib_metaquot.ml
```diff
@@ -39,6 +39,30 @@ module Make(M : sig
       match t.ptyp_desc with
       | Ptyp_extension ({ txt = "t"; _}, _ as ext)-> M.cast ext
       | _ -> super#core_type t
+
```
xclerc: This change should probably not be part of this PR.

.travis.yml
```diff
@@ -0,0 +1,13 @@
+language: c
```
xclerc: I suddenly feel sad. (but OCaml is not listed [here](https://docs.travis-ci.com/user/languages/))

metaquot/ppxlib_metaquot.ml
xclerc: (sorry, I meant the support for the new quotations)

.travis.yml
diml: indeed :(

metaquot/ppxlib_metaquot.ml
diml: removed

diml:
> Would it make sense to run the tests only on the versions that can compile
expect tests, and only check that we can build on previous versions?

Yes, let me check if we can do that

diml:
The last commit seems to do the trick

xclerc:
Excellent!

.travis.yml
yomimono: issue about it here: https://github.com/ocaml/ocaml-ci-scripts/issues/53 ; there is a path for becoming a "community-supported language", which requires at least 3 community contributors interested in maintaining the support.  We have 2 so far!

PR #34, opened 2018-06-06
Attempt to fix AppVeyor issues (end-of-line characters in diff)

xclerc:
See whether passing the `--strip-trailing-cr` command-line
switch to `diff` fixes the AppVeyor issues.

### DIFF
--- a/test/driver/omp-integration/test/jbuild
+++ b/test/driver/omp-integration/test/jbuild
@@ -8,6 +8,7 @@
 
 (alias
  ((name runtest)
-  (action (diff --strip-trailing-cr test.expected test.output))))
+  (deps (test.expected test.output))
+  (action (run diff --strip-trailing-cr test.expected test.output))))
 
 (jbuild_version 1)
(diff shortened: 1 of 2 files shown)


xclerc:
OK, that seems to do the trick. AppVeyor is now happy,
and the Travis error is about opam not being able to
download some packages.

For the record, I tweaked the tests rather than override
the diff command (using Dune's `--diff-command` option)
so that we still benefit from `patdiff` for the other tests.
@diml, are you OK with that?

(note: this PR should be squased.)

diml:
@xclerc yep, that's fine. FTR, we are hoping to sort this issue for Dune 1.0.0 (https://github.com/ocaml/dune/issues/844)

xclerc:
Indeed, it is definitely not a ppxlib-specific issue.
I actually wondered whether it should be fixed in
Dune and/or patdiff. Anyway, as far as ppxlib is
concerned, only two "minor" tests are impacted
and it is a bit annoying to have to consult the
various AppVeyor outputs to ensure that he error
is just the "expected" one.

diml:
agreed

PR #37, opened 2018-08-08
Update AST to 4.07

xclerc:
note: this PR also removes support for the if-then-else check
(however elements, such as the command-line switch, are
kept for compatibility)


diml:
Should we print a warning when this flag is passed?

diml:
(I tweaked a bit the error message)

PR #38, opened 2018-08-08
Remove the `--dev` passed to jbuilder

xclerc:
It is an error to pass this command-line switch to recent
jbuilder versions, leading to build failures.


diml:
It's a bug in dune 1.1.0 actually. It was fixed in master, we are going to do a release today

xclerc:
I have just checked that dune 1.1.1 accepts the flag; hence closing this PR.

PR #41, opened 2018-09-04
Let the tool name pass throught when used as a -ppx

diml:
This is needed for ppx_import

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@ next
 
 - Add `Attribute.declare_with_name_loc` (#33, @diml)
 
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -3,6 +3,8 @@ next
 
 - Add `Attribute.declare_with_name_loc` (#.., @diml)
 
+- Let the tool name pass throught when used as a -ppx (#..., @diml)
```
xclerc: Don't forget to update with the PR name (same two lines up).

diml: Indeed, thanks

Issue #36, opened 2018-07-13
File "src/jbuild", line 18, characters 17-40: # Error: "ocaml-migrate-parsetree" is not a ppx driver during the opam upgrade with dune-1.0


XVilka:
```

#=== ERROR while compiling ppxlib.0.2.1 =======================================#
# context     2.0.0~rc3 | linux/x86_64 | ocaml-system.4.06.0 | file:///home/akochkov/data/tmp/opam-repository
# path        ~/.opam/default/.opam-switch/build/ppxlib.0.2.1
# command     ~/.opam/default/bin/jbuilder build -p ppxlib -j 7
# exit-code   1
# env-file    ~/.opam/log/ppxlib-630120-bf8001.env
# output-file ~/.opam/log/ppxlib-630120-bf8001.out
### output ###
# File "src/jbuild", line 18, characters 17-40:
# Error: "ocaml-migrate-parsetree" is not a ppx driver



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
┌─ The following actions failed
│ λ build ppxlib 0.2.1
└─ 
```

xclerc:
I stumbled upon this one yesterday.

The fix, as provided to me by @diml, is to reinstall
`ocaml-migrate-parsetree` **after** installing the `dune`
package.

diml:
This should be fixed since [this PR](https://github.com/ocaml/opam-repository/pull/12350)

basus:
The referenced PR seems to have been merged, but I'm still getting the original error. I'm running OCaml 4.06.1 and OPAM 1.2.2 on Arch Linux.

diml:
If you do `opam update` followed by `opam upgrade`, does it reinstall ocaml-migrate-parsetree?

neojski:
Doing `opam update && upam upgrade` didn't fix it for me but doing `opam uninstall ocaml-migrate-parsetree && opam install ocaml-migrate-parsetree` did the trick and I was able to install core after that. I'm using ocaml 4.06.0 and opam 1.2.2.

cfcs:
This is broken for me too, latest version of opam, on debian.

cfcs:
perhaps we could update the constraint in the `opam` file to be `= 1.0.11` since `>= 1.0.9` doesn't seem to work?

XVilka:
I am closing it as solved.

PR #42, opened 2018-09-11
make it build with current opam universe

cfcs:
fixes https://github.com/ocaml-ppx/ppxlib/issues/36

### DIFF
--- a/ppxlib.opam
+++ b/ppxlib.opam
@@ -16,7 +16,7 @@ depends: [
   "base"                    {>= "v0.11.0"}
   "jbuilder"                {build & >= "1.0+beta18.1"}
   "ocaml-compiler-libs"     {>= "v0.11.0"}
-  "ocaml-migrate-parsetree" {>= "1.0.9"}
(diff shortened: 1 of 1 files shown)


diml:
Thanks, could you sign your commit as explained in the [CONTRIBUTING.md](https://github.com/ocaml-ppx/ppxlib/blob/master/CONTRIBUTING.md) file?

diml:
?

cfcs:
That document is 2048x larger than my patch, and I'm not a lawyer; if you don't want it, that's fine, I'll just pin it locally and get on with my day.

diml:
To be more precise, the only part you need to read is [the following one](https://github.com/ocaml-ppx/ppxlib/blob/master/CONTRIBUTING.md#signing-commits) . Most of it is the official DCO text from the Linux Foundation, and the rest is a couple of lines telling you to do `git commit -s`. Note that many open source projects use a DCO nowadays, the text itself is quite simple and you don't need to be a lawyer to understand it.

PR #44, opened 2018-09-21
Fix the special_function rule for dotted operators

Octachron:
The rule for special functions, `Context_free.Rule.special_function`,  fails for operators with a dot inside their name (e.g. `+.+`) due to `Longident.parse` converting `+.+` to `Ldot(+,Lident +)` .  This small PR proposes to fix this issues by adding a function `Longident.parse_with_operator` that can also parse identifiers of the form `A_1.A_2....A_n.(anything)` to `Ldot( ..., anything)` .

### DIFF
--- a/src/context_free.ml
+++ b/src/context_free.ml
@@ -126,7 +126,7 @@ module Rule = struct
 
   let special_function id f =
     T (Special_function, { name   = id
-                         ; ident  = Longident.parse_with_operator id
+                         ; ident  = Longident.parse id
                          ; expand = f
                          })
   ;;
--- a/src/context_free.ml
+++ b/src/context_free.ml
@@ -126,7 +126,7 @@ module Rule = struct
 
   let special_function id f =
     T (Special_function, { name   = id
-                         ; ident  = Longident.parse id
+                         ; ident  = Longident.parse_with_operator id
                          ; expand = f
                          })
   ;;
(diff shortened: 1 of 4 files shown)


diml:
Thanks, I suggest to just change the behaviour of `Longindent.parse` instead of adding a new function, given that the current behaviour is not very useful.

BTW, in this case, wouldn't it be simpler to make `special_function` take a `Longident.t` directly? I can't remember why I originally used a string here, but it looks like a mistake.

diml:
Ok, let's change `special_function` to take a `Longident.t` then. I don't believe it is used a lot, so it seems fine to change its signature.

Octachron:
After a glance at `ppx_jane` dependencies, there are at least three ppx rewriters (ppx_custom_printer, ppx_fail, ppx_pipebang) that uses `special_function`. They would be easy enough to fix, but is it fine to break them?

Moreover, I have a look at the use of `Longident.parse` in `ppxlib` and there is the same problem with `Ast_builder.evar`  which use `Longident.parse` on the input identifier and thus requires this PR's patch to work with `"+.+"`.  

Maybe it would be fine to have a string and longident variants for both functions?

diml:
Regarding `evar`, it's more a convenient function that is used quite a bit, so it's better to keep using a string. There is already a `Longident.t` variant: `pexp_ident`.

Regarding `special_function`, I thought ppx_custom_printf was the last user. We are planning a few breaking changes for the 1.0.0, I guess we can change the signature of `special_function` as part of this. In the meantime I'll merge this PR. Thanks!

PR #45, opened 2018-09-29
No bash and port to dune

rgrinberg:
I've noticed that ppxlib depends on bash [1].

[1] http://cvsweb.netbsd.org/bsdweb.cgi/pkgsrc/devel/ocaml-ppxlib/Makefile.diff?r1=1.3&r2=1.4


rgrinberg:
The port is complete, but I'm not sure how to port a particular test. omp-integration, has the following rule:

```
(executable
 ((name test)
  (flags (:standard -safe-string))
  (preprocess (pps (ppxlib_driver_omp_test_ppx -no-check ppxlib.runner)))))
```

But we no longer really have the ability to select a driver in dune files. The correct way to run this test would be to somehow hide ppxlib from the current environment.

rgrinberg:
The tests are failing because they seem to be requiring findlib:

```
-  module N = Ppxlib_private.Name
-  [%%expect{|
- +Unknown directive `require'.
- +Unknown directive `require'.
- +Line _, characters 5-9:
- +Error: Unbound module Base
- +Line _, characters 5-10:
- +Error: Unbound module Stdio
-  module N = Ppxlib.Ppxlib_private.Name
-  |}]
```

Was this test dependency omitted by mistake?

diml:
Yh, it's well possible about the missing findlib dependency. The omp package used to depend on it even though it wasn't using it, so it's possible it was just available by chance.

diml:
For the ppx_driver_omp_test, it's testing that when ppxlib is the driver, then transformations registered with omp directly are still active. The best way to fix this test is to add a ppxlib based rewriter and make sure both the omp and ppxlib rewriters are active.

rgrinberg:
Okay, @diml I've added a ppxlib based rewriters and used both rewriters (omp based and ppx based) in the test.

test/driver/omp-integration/test/dune
```diff
@@ -0,0 +1,12 @@
...
+ (flags (:standard -safe-string))
+ (preprocess (pps ppxlib_driver_omp_test_ppx ppxlib_ppx)))
+
+(rule
+ (with-stdout-to test.output (run ./test.exe)))
+
+(alias
+ (name runtest)
+ (deps test.expected test.output)
+ (action (diff test.expected test.output)))
```
diml: Could these be merged into a single `test` stanza?

diml:
The testsuite is still failing BTW, does it passes on your machine?

rgrinberg:
> The testsuite is still failing in Travis BTW, does it passes on your machine?

Yes, they pass for me. It's still that `ocamlfind` missing issue:

```
- +Unknown directive `require'.
```

I'm just wondering why it works on 4.04.

> Could these be merged into a single test stanza?

Yeah, quite likely. I will look into that.

diml:
Regarding ocamlfind, does adding `ocamlfind` as `{test}` dependency solves the problem? If yes, let's just do it.

diml:
Ok, this looks ready to me and CI is passing.

diml:
Thanks!

PR #46, opened 2018-10-01
Ignore all attributes starting with _

### LINKED ISSUE #40: Ignore more machine generated attributes
We currently whilelist merlin attributes as they can safely be dropped. Reason needs something similar, cf https://github.com/janestreet/ppx_inline_test/issues/15.

To be more generic, we can simply whitelist all attributes with a name starting with `_`. We should also whitelist specific existing reason attributes to make things simpler.

diml:
Fixes #40

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+next
+----
+
+- Do not report errors about dropped or uninterpreted attributes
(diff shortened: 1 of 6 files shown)


diml:
Indeed, I added them as reserved namespaces.

PR #51, opened 2018-10-19
Fix attribute extraction for Otag/Rtag

xclerc:
The library fails on declarations such as:

    type t1 = < >
    type t2 = < t1 >

This handles the `Oinherit`/`Rinherit` cases by just returning the
empty list for `get` and doing nothing for `set`. Another option
would be to perform the operation on the nested `core_type`
value, but I am not sure it makes sense.

### DIFF
--- a/src/attribute.ml
+++ b/src/attribute.ml
@@ -80,15 +80,6 @@ module Context = struct
     | Psig_extension (e, l) -> (e, l)
     | _ -> failwith "Attribute.Context.get_psig_extension"
 
-  let get_rtag : row_field -> _ = function
-    | Rtag (lbl, attrs, can_be_constant, params_opts) ->
-      (lbl, attrs, can_be_constant, params_opts)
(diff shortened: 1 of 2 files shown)


src/attribute.ml
```diff
@@ -150,11 +149,17 @@ module Context = struct
     | Psig_extension ->
       { x with psig_desc = Psig_extension (get_psig_extension x |> fst, attrs) }
     | Rtag                   ->
-      let (lbl, _, can_be_constant, params_opts) = get_rtag x in
-      Rtag (lbl, attrs, can_be_constant, params_opts)
+      begin match x with
+      | Rtag (lbl, _, can_be_constant, params_opts) ->
+        Rtag (lbl, attrs, can_be_constant, params_opts)
+      | Rinherit _ -> x
```
diml: Should we assert that the list of attributes is empty here? Otherwise it is effectively silently dropped

Issue #61, opened 2019-02-21
Getting tool_name from rewriter and/or derivers


NathanReb:
Following https://github.com/ocaml/dune/issues/1861 I'm trying to add a specific behaviour when my `ppxlib` based deriver is used in an `ocamldep` context. `ocaml-migrate-parsetree` seems to make it available through a config argument but I can't seem to find how I could access it from `ppxlib`.

I worked out an ugly solution to come around that limitation for now but if I'm indeed not missing something, it might be useful to find a way to pass it on to the actual `ast -> ast` function.

NathanReb:
It turns out my ugly workaround doesn't even work so my current solution is to always silently ignore abstract types. What I'd really want is to only ignore them in an `ocamldep` context and produce an error otherwise. The current behaviour obviously isn't satisfactory as it would silently fail without giving any feedback to the user.

diml:
Indeed, we don't forward this information in ppxlib. Adding the forwarding of this information might be invasive, but given that this is a global property of the running instance, maybe we can get away with a global setting. For instance, we could add a function like this to `driver.mli`:

```
val tool_name : unit -> string
```

NathanReb:
I'm usually not too fond of such globals but if you feel like that's the way to go that definitely works for me!
I guess it's the least invasive non API-breaking way of exposing `tool_name` here.
In the long run I assume it could be part of some abstract `context` value passed instead of `loc` and `path` to `ast -> ast` functions, as you already suggested in #60. Such a config could then easily be enriched with new fields without breaking the API.

Actually what would you think about adding such a `context` type and new versions of `Extension.declare` and `Deriving.Generator.make` now?

Let me know which one you'd prefer and I'll have a go at it!

diml:
Passing a context around does look better indeed. If you are happy to have a go then I'm all for it!

NathanReb:
Ok great! I'm not too familiar with `ppxlib`'s internals yet so if you have any advice/suggestion I'd gladly take it!

diml:
You can start by passing the context to the various functions in `Transform.t` in src/driver.ml. You don't need to touch the `rules` field since it is turned into the generic `impl` or `intf` field before the transformations are applied. Then by following type errors you should find the places that need to be updated. Feel free to ask if you need more details about some part of the code.

NathanReb:
[...]

Issue #60, opened 2019-02-05
Pass attribute location to Deriving.Generator


NathanReb:
Currenlty it seems that the location passed to the generator function
```ocaml
loc: Location.t -> path: string ->  `input_ast -> `output_ast
```
at least in the `str_type_decl` case is the loc for the whole `Pstr_type` node. This is a bit annoying because if you then use that `loc` for the generated nodes (which I assume is what most `ppxlib` users do) that means that any error in there will be reported on the whole type declaration. That can be a bit confusing to the end user of the deriving plugin.

What I'd like to be able to do instead is locate it on the attribute node for `[@@deriving ...]` which I believe would make more sense. It might even be worth locating it on the right part of the attribute payload, ie the one corresponding to the registered deriver.

I don't think there's a nice way, if there's even one, to get that location from the input ast nodes.

Would you consider changing what is passed as `loc` or at least adding a separate argument for the different relevant locations?

hhugo:
What version are you using. I think it was fixed 2 months ago. https://github.com/ocaml-ppx/ppxlib/pull/55

NathanReb:
I'm using `0.5.0`.

NathanReb:
Let me know if you want me to provide a minimal example.

diml:
I usually use the location of the element I'm processing. i.e. if I see `x` and need to produce a reference to `sexp_of_x`, then the location of `sexp_of_x` will be the same as `x`. That works relatively well in practice.

I suppose passing the location of the deriver name to the generator would be fine. Since this requires an API change, we should add a more future proof version of the `make` function, for instance one that take a context as argument containing the various locations and the path.

NathanReb:
I tend to find it a little unsettling when errors for generated code are located on code that I wrote myself, especially when the error is highlighted by merlin in my editor. I can imagine it being even more disturbing to ocaml newcomers. That might be completely subjective though.

I also feel like it would look more consistent with how ppxlib's extension work.

From that same subjective point of view I think that'd be great to have such a context value passed to the generator. On the other hand, I wonder if consistency across ppxlib based plugins wouldn't be preferable. Having such a context with multiple locations to chose from might result in every plugin having its own conventions.

NathanReb:
> when errors for generated code are located on code that I wrote myself

Let me clarify that a bit since it will inevitably be located on code that I wrote. What I mean is I prefer the error to be located on the part of the code responsible for the code generation, ie in the case of a deriver, here:
[...]

PR #67, opened 2019-03-21
Expose an abstract Code_path.t type to replace expanders' string path argument

### LINKED ISSUE #65: Context dependent extensions
I'm currently trying to write a ppx extension that would be rewritten to the fully qualified path of the value it appears in (see https://github.com/NathanReb/ppx_there).

It seems that the right way to do that currently is to use `Ppxlib.Driver.register_transformation ~impl:my_expand_function` because I need some context to actually rewrite the extension node.

I looked into `Extension` though and found out about `Extension.Expert`. I'm not quite sure I got this right but I have the impression that for now the only thing it allows me to do is to verify the payload fits the declared `Ast_pattern.t` and if so get the payload node.

I was wondering if there were more helpers for such ppx plugins provided by `ppxlib` as I feel like I'm gonna redo a lot of the work that's already done here.

NathanReb:
Fixes #65 

This is PR adds a new `Code_path` module to represent code path in an abstract and improved way.
It exposes an abstract type, a smart constructor and a few accessors.

It also adds an `Ast_traverse.map_with_code_path` class similar to the current `map_with_path` which uses the new representation instead of a string.

This is now used by `Driver`, `Deriving` and `Extension` in order to expose V2 variants that pass a `Code_path.t` to expanders. V1 functions are just rewritten on top of them.

### DIFF
--- a/src/ast_traverse.ml
+++ b/src/ast_traverse.ml
@@ -69,6 +69,25 @@ class map_with_path = object
     super#module_type_declaration (enter mtd.pmtd_name.txt path) mtd
 end
 
+class map_with_code_path = object
+  inherit [Code_path.t] map_with_context as super
+
+  method! module_binding path mb =
+       super#module_binding (Code_path.enter mb.pmb_name.txt path) mb
+
+  method! module_declaration path md =
+    super#module_declaration (Code_path.enter md.pmd_name.txt path) md
+
+  method! module_type_declaration path mtd =
+    super#module_type_declaration (Code_path.enter mtd.pmtd_name.txt path) mtd
+
+  method! value_description path vd =
+    super#value_description (Code_path.enter vd.pval_name.txt path) vd
+
+  method! value_binding path vb =
+    super#value_binding path vb
+end
+
 class sexp_of = object
   inherit [Sexp.t] Ast.lift
 
--- a/src/ast_traverse.ml
+++ b/src/ast_traverse.ml
@@ -69,6 +69,16 @@ class map_with_path = object
     super#module_type_declaration (enter mtd.pmtd_name.txt path) mtd
 end
 
+let var_names_of = object
+  inherit [string list] fold as super
+
+  method! pattern p acc =
+    let acc = super#pattern p acc in
+    match p.ppat_desc with
+    | Ppat_var {txt; _} -> txt :: acc
+    | _ -> acc
+end
+
 class map_with_code_path = object (self)
   inherit [Code_path.t] map_with_context as super
 
@@ -84,16 +94,17 @@ class map_with_code_path = object (self)
   method! value_description path vd =
     super#value_description (Code_path.enter vd.pval_name.txt path) vd
 
-  method! value_binding path ({pvb_pat; pvb_expr; pvb_attributes; pvb_loc} as vb) =
-    match pvb_pat.ppat_desc with
-    | Ppat_var v ->
-      let new_path = Code_path.enter v.txt path in
-      let pvb_pat = self#pattern path pvb_pat in
-      let pvb_expr = self#expression new_path pvb_expr in
-      let pvb_attributes = self#attributes new_path pvb_attributes in
-      let pvb_loc = self#location path pvb_loc in
-      { pvb_pat; pvb_expr; pvb_attributes; pvb_loc }
-    | _ -> super#value_binding path vb
+  method! value_binding path {pvb_pat; pvb_expr; pvb_attributes; pvb_loc} =
+    let all_var_names = var_names_of#pattern pvb_pat [] in
+    let var_name = Base.List.last all_var_names in
+    let in_binding_path =
+      Base.Option.fold var_name ~init:path ~f:(fun path var_name -> Code_path.enter var_name path)
+    in
+    let pvb_pat = self#pattern path pvb_pat in
+    let pvb_expr = self#expression in_binding_path pvb_expr in
+    let pvb_attributes = self#attributes in_binding_path pvb_attributes in
+    let pvb_loc = self#location path pvb_loc in
+    { pvb_pat; pvb_expr; pvb_attributes; pvb_loc }
 end
 
 class sexp_of = object
--- a/src/ast_traverse.ml
+++ b/src/ast_traverse.ml
@@ -83,22 +83,24 @@ class map_with_code_path = object (self)
   inherit [Code_path.t] map_with_context as super
 
   method! module_binding path mb =
-       super#module_binding (Code_path.enter mb.pmb_name.txt path) mb
+       super#module_binding (Code_path.enter ~loc:mb.pmb_loc mb.pmb_name.txt path) mb
 
   method! module_declaration path md =
-    super#module_declaration (Code_path.enter md.pmd_name.txt path) md
+    super#module_declaration (Code_path.enter ~loc:md.pmd_loc md.pmd_name.txt path) md
 
   method! module_type_declaration path mtd =
-    super#module_type_declaration (Code_path.enter mtd.pmtd_name.txt path) mtd
+    super#module_type_declaration (Code_path.enter ~loc:mtd.pmtd_loc mtd.pmtd_name.txt path) mtd
 
   method! value_description path vd =
-    super#value_description (Code_path.enter vd.pval_name.txt path) vd
+    super#value_description (Code_path.enter ~loc:vd.pval_loc vd.pval_name.txt path) vd
 
   method! value_binding path {pvb_pat; pvb_expr; pvb_attributes; pvb_loc} =
     let all_var_names = var_names_of#pattern pvb_pat [] in
     let var_name = Base.List.last all_var_names in
     let in_binding_path =
-      Base.Option.fold var_name ~init:path ~f:(fun path var_name -> Code_path.enter var_name path)
+      Base.Option.fold var_name
+        ~init:path
+        ~f:(fun path var_name -> Code_path.enter ~loc:pvb_loc var_name path)
     in
     let pvb_pat = self#pattern path pvb_pat in
     let pvb_expr = self#expression in_binding_path pvb_expr in
--- a/src/ast_traverse.ml
+++ b/src/ast_traverse.ml
@@ -82,6 +82,9 @@ end
 class map_with_code_path = object (self)
   inherit [Code_path.t] map_with_context as super
 
+  method! expression path expr =
+    super#expression (Code_path.enter_expr path) expr
+
   method! module_binding path mb =
        super#module_binding (Code_path.enter_module ~loc:mb.pmb_loc mb.pmb_name.txt path) mb
 
--- a/src/ast_traverse.ml
+++ b/src/ast_traverse.ml
@@ -69,7 +69,7 @@ class map_with_path = object
     super#module_type_declaration (enter mtd.pmtd_name.txt path) mtd
 end
 
-class map_with_code_path = object
+class map_with_code_path = object (self)
   inherit [Code_path.t] map_with_context as super
 
   method! module_binding path mb =
@@ -84,9 +84,15 @@ class map_with_code_path = object
   method! value_description path vd =
     super#value_description (Code_path.enter vd.pval_name.txt path) vd
 
-  method! value_binding path vb =
-    match vb.pvb_pat.ppat_desc with
-    | Ppat_var v -> super#value_binding (Code_path.enter v.txt path) vb
+  method! value_binding path ({pvb_pat; pvb_expr; pvb_attributes; pvb_loc} as vb) =
+    match pvb_pat.ppat_desc with
+    | Ppat_var v ->
+      let new_path = Code_path.enter v.txt path in
+      let pvb_pat = self#pattern path pvb_pat in
+      let pvb_expr = self#expression new_path pvb_expr in
+      let pvb_attributes = self#attributes new_path pvb_attributes in
+      let pvb_loc = self#location path pvb_loc in
+      { pvb_pat; pvb_expr; pvb_attributes; pvb_loc }
     | _ -> super#value_binding path vb
 end
 
--- a/src/ast_traverse.ml
+++ b/src/ast_traverse.ml
@@ -85,7 +85,9 @@ class map_with_code_path = object
     super#value_description (Code_path.enter vd.pval_name.txt path) vd
 
   method! value_binding path vb =
-    super#value_binding path vb
+    match vb.pvb_pat.ppat_desc with
+    | Ppat_var v -> super#value_binding (Code_path.enter v.txt path) vb
+    | _ -> super#value_binding path vb
 end
 
 class sexp_of = object
--- a/src/ast_traverse.ml
+++ b/src/ast_traverse.ml
@@ -83,16 +83,18 @@ class map_with_code_path = object (self)
   inherit [Code_path.t] map_with_context as super
 
   method! module_binding path mb =
-       super#module_binding (Code_path.enter ~loc:mb.pmb_loc mb.pmb_name.txt path) mb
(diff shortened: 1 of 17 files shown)


NathanReb:
This is very raw but I wanted your feedback on a few things at this point. Especially:
1. So far I just modified the existing v2 variants in `Extension` but I imagine it would be worth adding a `Deriving.Generator.V2` module as well so that deriver plugins authors can also use that new code path representation.
2. I took the liberty to modify `Rule.attr_group_inline` and `Rule.attr_inline` types as they are to be removed. This seemed more convenient than to do the `Code_path.t -> string` wrapping in `Deriving` and also would make it easier to do `1`. Not sure if that's ok.
3. Dealing with value_bindings is slightly more complex as we have to deal with patterns. A good v1 would be to only update the path when the pattern is just a variable. There doesn't seem to be a straightforward way of dealing with more complex patterns, for example in:
```ocaml
let a, b = [%ext left], [%ext right]
```
to pass `X.a` as the path to the expander for `[%ext left]` and `X.b` to `[%ext right]`.

Finally, I can't get the tests to run properly on my machine. I'm getting a bunch of these:
```
$ dune runtest
File "test/base/dune", line 5, characters 14-52:
5 |   (glob_files %{project_root}/src/.ppxlib.objs/*.cmi))
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Warning: Directory src/.ppxlib.objs doesn't exist.
```
followed by a few:
```
 expect_test alias test/driver/attributes/runtest (exit 2)
(cd _build/default && test/expect/expect_test.exe test/driver/attributes/test.ml) &> /dev/null
```

Does that look familiar to you? I'll try to figure it out on my own as I'm gonna add tests for `Code_path` anyway.

Let me know what you think!

src/extension.mli
```diff
@@ -151,13 +151,13 @@ module V2 : sig
     :  string
     -> 'context Context.t
     -> (payload, 'a, 'context) Ast_pattern.t
-    -> (loc:Location.t -> path:string -> 'a)
+    -> (loc:Location.t -> path:Code_path.t -> 'a)
```
diml: We should probably mint a `V3` module here

diml:
> So far I just modified the existing v2 variants in Extension but I imagine it would be worth adding a Deriving.Generator.V2 module as well so that deriver plugins authors can also use that new code path representation.

Modifying the existing `V2` module doesn't seem right to me; versioned modules should be stable. We should mint an `Extension.V3` module instead. Adding `Deriving.Generator.V2` seems good.

> I took the liberty to modify Rule.attr_group_inline and Rule.attr_inline types as they are to be removed. This seemed more convenient than to do the Code_path.t -> string wrapping in Deriving and also would make it easier to do 1. Not sure if that's ok.

Seems good to me.

> Dealing with value_bindings is slightly more complex as we have to deal with patterns. A good v1 would be to only update the path when the pattern is just a variable. There doesn't seem to be a straightforward way of dealing with more complex patterns, for example in:

I'd suggest to use the first variable in the pattern. It might look slightly odd, but at the end of the day it should still help the developer quickly identify what the path is referring to.

> Finally, I can't get the tests to run properly on my machine. I'm getting a bunch of these:

Ah, it's because of a recent change in Dune. These paths are not part of the spec and so are not stable :/ Try adding `byte/` before the `*.cmi`

NathanReb:
> Modifying the existing V2 module doesn't seem right to me; versioned modules should be stable. We should mint an Extension.V3 module instead.

Hmm ok, I get it. I guess I thought of `V2` more as an `Unstable`. I'll move them to `V3`.
Just curious though, what's the reason behind `V2`? I'm wondering cause it is just an alias for `declare` and `declare_inline` for now.

> I'd suggest to use the first variable in the pattern. It might look slightly odd, but at the end of the day it should still help the developer quickly identify what the path is referring to.

Ok that sound good to me. Is there a foreseeable future where we might have a more precise path as described in my example?

I was also wondering if it wasn't preferable to add the variable name to the path only when descending into the expression. For instance in the following example:
```ocaml
let a, [%some_ext] = 0, 0
```
We probably don't want `[%some_ext]`'s expander to receive `X.a` as the path.

> Ah, it's because of a recent change in Dune. These paths are not part of the spec and so are not stable :/ Try adding byte/ before the *.cmi

Great, thanks! I'll try that!

diml:
> Hmm ok, I get it. I guess I thought of V2 more as an Unstable. I'll move them to V3.
Just curious though, what's the reason behind V2? I'm wondering cause it is just an alias for declare and declare_inline for now.

It was meant to be temporary. At some point I did a breaking change in `declare` and I proceeded that way: I added `V2.declare`, deprecated `declare` and waited for a while. Then I replaced `declare` by the new `V2.declare` and I was planning to eventually deprecate `V2` and then delete it altogether.

I now think this is too complicated. Not only you have to remember to go through all these steps, but if a user of ppxlib jumps a few versions of ppxlib at once, they won't see the intermediate steps and it might be hard to understand what's happening. It's much simpler to use fully backward compatible versioned modules.

BTW, I remember that we talked once about using an abstract context rather than passing the location and path explicitly. Maybe we should take the opportunity to do this change? The API will be more future proof that way.

> Ok that sound good to me. Is there a foreseeable future where we might have a more precise path as described in my example?

I haven't thought much about it, though it does seem like a hard problem. I don't even know if there is always a solution.

> We probably don't want [%some_ext]'s expander to receive X.a as the path.

Agreed. I was more thinking of cases like this:

```ocaml
let a, b, [%some_ext] = [%some_other_ext]
```

if the path before this is `p`, then the patch in `[%some_ext]` is `p` and the path in `[%some_other_ext]` is `Code_path.enter "a" p`. Does that look reasonable to you?

NathanReb:
I just pushed improved variable name selection + incremented code path for expression but not for pattern.
I also choose to pass the incremented path when descending into the attributes as I felt it would make sense but I don't have a strong opinion about it.

> BTW, I remember that we talked once about using an abstract context rather than passing the location and path explicitly. Maybe we should take the opportunity to do this change? The API will be more future proof that way.

Yeah it probably is the right time. I'm not sure what's the best approach here, I tried refreshing my memory by looking at #61 but I think I need to have a go at it first, see what the issues are with this.

NathanReb:
I'm still having issues running the tests. I fixed the errors that I introduced but the tests still don't run properly and I get a whole bunch of:
```diff
 let _ = convert_longident "Base.( + )"
 [%%expect{|
-- : string * longident =
-("Base.( + )", Ppxlib.Longident.Ldot (Ppxlib.Longident.Lident "Base", " + "))
+Line _, characters 8-25:
+Error: Unbound value convert_longident
```
Adding `byte/` to the `.cmi` test dependency did fix the warning but running the tests still fails with those undound values and unbound modules, either on master or on my branch. I'm not super familiar with expect tests unfortunately.

NathanReb:
> I haven't thought much about it, though it does seem like a hard problem. I don't even know if there is always a solution.

Indeed there's no good behaviour for stuff like:
```ocaml
let a, b = List.split [(0, 0)]
```

Probably not worth bothering about that too much although it might be nice to have a way for the expander to know about that potential inaccuracy. For instance in `ppx_there` I'd be interested to know about that so I can eventually raise in such cases.

NathanReb:
I pushed a new commit that defines an `Expansion_context` module and uses that instead of separated `loc` and `path` arguments for the new `Extension` and `Deriving` APIs.

For now all the work's done in `Context_free.map_top_down` where an `Expansion_context.t` is build to be passed to `Deriving` and `Extensions` functions.

At first I was wondering about a `Ast_traverse.map_with_expansion_context` but it seems tedious and unnecessary to deal with the locations in there. I guess in the future, as `Expansion_context` is extended with more stuff it might be worth extracting the "location free" parts into a type and eventually introduce some kind of `map_with_location_free_ctxt`.

Besides the test I think this is pretty much ready for review so I'll undraft it so I can start adressing issues while I'm figuring out how to get the test to run properly on my machine!

NathanReb:
Finally figured out the issue with the tests, I had to update the `Topdir` paths in `expect_test.mll` as well.

I pushed two new commits, one contains the actual expect tests for the `Code_path` module while the other contains the required updates to get the tests to run on my machine. I'm not sure whether we want this last one merged or not. The whole test setup seems pretty fragile so let me know If you'd prefer that I write those tests differently.

diml:
Ah, I just saw your last message about  the tests. I merged a fix in master.

src/ast_traverse.ml
```diff
@@ -69,6 +69,44 @@ class map_with_path = object
...
+  method! module_declaration path md =
+    super#module_declaration (Code_path.enter md.pmd_name.txt path) md
+
+  method! module_type_declaration path mtd =
+    super#module_type_declaration (Code_path.enter mtd.pmtd_name.txt path) mtd
+
+  method! value_description path vd =
+    super#value_description (Code_path.enter vd.pval_name.txt path) vd
+
+  method! value_binding path vb =
```
diml: We should deconstruct `vb` in the pattern and without using `; _`. This way we are sure to process all the fields.

src/code_path.ml
```diff
@@ -0,0 +1,34 @@
...
+let fully_qualified_path t = 
+  String.concat "." (t.main_module_name :: (val_path t))
+
+let enter mod_or_val_name t =
+  {t with val_path = mod_or_val_name :: t.val_path}
+
+let to_string_path t =
+  let val_path = val_path t in
+  let sub_module_path =
+    let is_module_name s = s <> "" && Base.(Char.is_uppercase (String.get s 0)) in
```
diml: We open `Import` in all modules, so we should do the same here. It's easier to reason about the code if everything uses the same stdlib.

NathanReb:
Great, just rebased on top of your changes and removed my commit fixing the tests!

diml:
In `Code_path`, the components of the path are currently plain strings. I suggest  to use a more precise type, such as:

```
type t =
  | Module of string
  | Variable of string
  | ...
```

and only convert the components to strings at the very end. It seems better to keep the extra information when working with the abstract path. Maybe we should keep the location of individual components as well?

BTW, regarding the expander context, I'm wondering if we should use the same one for derivers and extension expanders. In particular, we might want to add fields that are specific to derivers in the future. It seems slightly more future proof to use different types. What do you think?

src/ast_traverse.ml
NathanReb: Indeed that's much cleaner, I changed it!

src/code_path.ml
NathanReb: Good point, I'll add that!

NathanReb:
Indeed, I'll clean up Code_path a little but since the variable should always be final it might be a better idea to change it to:
```ocaml
type t =
  { file_path : string
  ; main_module_name : string
  ; module_path : string list
  ; variable_name : string option
  }
```

> Maybe we should keep the location of individual components as well?

That could prove useful in the future indeed. Not sure what the best solution is here, we could reuse the ast `'a loc` type although that doesn't seem really clean.

> BTW, regarding the expander context, I'm wondering if we should use the same one for derivers and extension expanders. In particular, we might want to add fields that are specific to derivers in the future. It seems slightly more future proof to use different types. What do you think?

I agree we might want to have different information passed to derivers and extensions. They will also probably share a lot. I guess we could either use two different types, built on top of a common part or the other way around, parametrize the expansion context by a `deriver` or `extension`.

I'm not sure what's best here, I guess the two different types is slightly better as we can pass the common part around thanks to `Ast_traverse.map_with_expansion_base_ctxt` and use smart constructors to build the right context in `Context_free.map_top_down` before applying the transformation.

src/code_path.ml
NathanReb: That triggers a warning because of the use of stdlib's `Filename`. Should I add `core_kernel` as a dependency or simply use `Caml.Filename` here?

diml: Let's just use `Caml.Filename`

diml:
Ah, the representation of the code path makes me think that maybe we have a slightly different idea of what the code path exactly is. For instance, I thought that the components of a code path could alternate between modules and variables because of first class modules, and it seems to me that this is indeed what is implemented by the current `map_with_code_path`. Do you agree?

> That could prove useful in the future indeed. Not sure what the best solution is here, we could reuse the ast 'a loc type although that doesn't seem really clean.

What about representing the path as a `(Loc.t * Component.t) list`, or a `Component.t loc list`?

> I'm not sure what's best here, I guess the two different types is slightly better as we can pass the common part around thanks to Ast_traverse.map_with_expansion_base_ctxt and use smart constructors to build the right context in Context_free.map_top_down before applying the transformation.

Agreed, that seems good to me

NathanReb:
> For instance, I thought that the components of a code path could alternate between modules and variables because of first class modules, and it seems to me that this is indeed what is implemented by the current map_with_code_path. Do you agree?

I looked into but I don't think that's actually the case. The only nodes for which the path can be updated are `value_binding`, `value_description`, `module_binding`, `module_declaration` and `module_type_declaration`. I don't think what you suggest can happen without dealing with `module_expr` right? I might be missing something here as I'm not too familiar with the first class module parts of the AST.
Would alternating between variables and module names in the path be something that we would even want? From the top off my head I can't think of a case where that would make much sense. If you have an example of a case where that would be desirable behaviour that would probably help.

> What about representing the path as a (Loc.t * Component.t) list, or a Component.t loc list?

Ok I'll use `loc` in that case as I'm personally not too fond of tuples when records can be used.

NathanReb:
Pushed the split of expansion context into deriver and extension specific contexts.

For the `Code_path` fixes I think we need to figure out whether module and values names can/should indeed alternate to decide which abstraction's the best.

NathanReb:
I also added locations to the current code path components.

Forgot to mention it but I took the liberty to give explicit names to the expansion contexts' `loc` and changed them to `extension_point_loc` and `derived_item_loc` so that it is clearer what they are and it makes easier to add different locations to the context such as the deriving plugin invocation location for instance (as mentioned in #60). Let me know what you think about that!

diml:
> Ok I'll use loc in that case as I'm personally not too fond of tuples when records can be used.

Seems good. I prefer records too.

> Forgot to mention it but I took the liberty to give explicit names to the expansion contexts' loc and changed them to extension_point_loc and derived_item_loc so that it is clearer what they are and it makes easier to add different locations to the context such as the deriving plugin invocation location for instance (as mentioned in #60). Let me know what you think about that!

That seems good to me!

Regarding the module/variable alternance, I added a test just to check the current behaviour. I find the result slightly surprising. I think we should either not include any of the module names after `s.`, or on the contrary include them all. Both choices are acceptable I guess. If we want to include all the modules, then we have to deal with `Pexp_letmodule` constructors.

NathanReb:
Yeah I just saw the test and the current behaviour is indeed far from ideal. I think I need to give it some thoughts to understand the implications of both approaches here.

I might be a bit biased as I hope to use this feature for `ppx_there` but it might be a good idea to split the path in two: a part until a value binding and a part for whatever comes after as I think they're slightly different.

diml:
That seems reasonable to me.

NathanReb:
I pushed some changes to improve the handling of the "toplevel" part of the code path, ie before descending into expressions.

I'll have a look into handling the rest but I think we could merge this as a first version and handle the in-expression path in an API compatible way later if we want to.

I also moved some `Code_path` function to the undocumented section. I was wondering whether we could go further and not expose them to the end user by constraining `Ppxlib`'s interface in a `ppxlib.mli`.
That can obviously be done separately!

diml:
Thanks, that looks good to me.

> I also moved some Code_path function to the undocumented section. I was wondering whether we could go further and not expose them to the end user by constraining Ppxlib's interface in a ppxlib.mli.
That can obviously be done separately!

That might cause quite a lot of duplication though. We can start with lighter solutions such as grouping private functions in `Private` or `Expert` sub-modules and see how it goes.

diml:
BTW, I let you merge the PR :)

NathanReb:
Great thank!

PR #70, opened 2019-03-28
Add a metaquot section to the manual

NathanReb:
This PR adds a manual section for PPX authors with a single sub section documenting `metaquot` usage.

### DIFF
--- a/doc/ppx-for-plugin-authors.rst
+++ b/doc/ppx-for-plugin-authors.rst
@@ -1,0 +1,101 @@
+**********************
+PPX for plugin authors
+**********************
+
+This section describes how to use ``ppxlib`` for PPX plugin authors.
+
+Metaquot
+------------
+
+``metaquot`` is a PPX plugin to help you write PPX plugins. It lets you write AST node values
+using the actual OCaml syntax instead of building them with the more verbose AST types or
+``Ast_builder``.
+
+To use metaquot you need to add it to the list of preprocessor for your PPX plugin:
+
+.. code:: scheme
+
+          (library
+           (name my_plugin_lib)
+           (preprocess (pps ppxlib.metaquot)))
+
+Metaquot exposes different extensions. The extension you should use depends on the type of AST node
+you're trying to write. You can use the following extensions with the following syntax:
+- ``expr`` for ``Parsetree.expression``: ``[%expr 1 + 1]``
+- ``pat`` for ``Parsetree.pattern``: ``[%pat? ("", _)]``
+- ``type`` for ``Parsetree.core_type``: ``[%type: int -> string]``
+- ``stri`` for ``Parsetree.structure_item``: ``[%stri let a = 1]``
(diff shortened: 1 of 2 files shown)


NathanReb:
For now it only contains documentation for the expression extension. I'd like to improve it to document the use of `metaquot` to write AST node patterns but since this is my first contribution to the manual I wanted to hear your feedback first as I don't know if the style fits what you expect.

I guess the pattern part could always be added later since this is already an improvement over the current situation but my guess is that it should probably be integrated into the current section rather than added as a separate part before or after.

doc/ppx-for-plugin-authors.rst
```diff
@@ -0,0 +1,101 @@
...
+          }
+
+For this to compile you need the AST types to be in the scope so you should always use metaquot
+where ``Ppxlib`` is opened.
+You'll also note that the generated node expects a ``loc : Location.t`` value to be available. The
+produced AST node value and every other nodes within it will be located to ``loc``. You should make
+sure ``loc`` is the location you want for your generated code when using metaquot.
+
+Using only these extensions you can only produce constant/static AST nodes. Metaquot as a solution
+for that as well: quasi-quotation. You can use quasi-quotation to insert any expression representing
```
diml: It is the first time I see the term quasi-quotation. Is is standard in this context? I always called these antiquotations

diml:
Thanks, the style looks good to me.

doc/ppx-for-plugin-authors.rst
NathanReb: I also usually use anti-quotation but I think I read the term quasi-quotation in this [whitequark's article](https://whitequark.org/blog/2014/04/16/a-guide-to-extension-points-in-ocaml/) the other day and got confused.

By re-reading it now it seems clear that quasiquotation is in fact used to designate `metaquot` as a whole which makes much more sense.

I'll change that!

NathanReb:
Added mentions of the pattern extension and a few improvements.

Let me know what you think!

doc/ppx-for-plugin-authors.rst
```diff
@@ -0,0 +1,150 @@
...
+``metaquot`` can be used both to write expressions of some of the AST types or to write patterns to
+match over those same types. The various extensions it exposes can be used in both contexts,
+expressions or patterns.
+
+The extension you should use depends on the type of AST node you're trying to write or to
+pattern-match over. You can use the following extensions with the following syntax:
+- ``expr`` for ``Parsetree.expression``: ``[%expr 1 + 1]``
+- ``pat`` for ``Parsetree.pattern``: ``[%pat? ("", _)]``
+- ``type`` for ``Parsetree.core_type``: ``[%type: int -> string]``
+- ``stri`` for ``Parsetree.structure_item``: ``[%str let a = 1]``
```
xclerc: `[%str` -> `[%stri`

NathanReb: Good catch, fixed!

PR #72, opened 2019-04-02
Add omp_config to expansion context

### LINKED ISSUE #61: Getting tool_name from rewriter and/or derivers

NathanReb:
Fixes #61 

This PR makes it possible to recover the `Ocaml_migrate_parsetree.Driver` config from the expansion context.

This is done by extracting the location independent parts of the context into `Expansion_context.Base` and replacing `Ast_traverse.map_with_code_path` with a `map_with_expansion_ctxt_base`.

### DIFF
--- a/src/ast_traverse.mli
+++ b/src/ast_traverse.mli
@@ -54,7 +54,7 @@ end
 
 class map_with_path : [string] map_with_context
 
-class map_with_code_path : [Code_path.t] map_with_context
+class map_with_expansion_context : [Expansion_context.Base.t] map_with_context
 
 class virtual ['res] lift : object
   inherit ['res] Ppxlib_traverse_builtins.lift
--- a/src/expansion_context.mli
+++ b/src/expansion_context.mli
@@ -39,7 +39,7 @@ module Extension : sig
   (** Undocumented section *)
 
   (** Build a new expansion context with the given extension point location and base context *)
-  val make : extension_point_loc:Location.t -> base:Base.t -> t
+  val make : extension_point_loc:Location.t -> base:Base.t -> unit -> t
 end
 
 module Deriver : sig
@@ -62,5 +62,5 @@ module Deriver : sig
   (** Undocumented section *)
 
   (** Build a new expansion context with the given item location and code path *)
-  val make : derived_item_loc:Location.t -> base:Base.t -> t
+  val make : derived_item_loc:Location.t -> base:Base.t -> unit -> t
 end
--- a/src/expansion_context.mli
+++ b/src/expansion_context.mli
@@ -1,3 +1,24 @@
+module Base : sig
+  (** Type for the location independent parts of the expansion context *)
+  type t
+
+  (**/*)
+  (** Undocumented section *)
+
+  (** Build a new base context at the top level of the given file with the given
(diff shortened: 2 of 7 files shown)


NathanReb:
This is an early version but I'd love to hear your feedback.

There's some places where I'm unsure I did the right thing, especially in `Driver.print_passes` where it was unclear what the config should be as I haven't taken the time to look into it and figure out exactly what that function was used for.

I was also wondering whether `map_with_expansion_ctxt_base` should be documented. I put everything from `Expansion_context.Base` into an undocumented section as it's quite likely to change, especially the smart constructor. That means that if one wants to use `map_with_expansion_ctxt_base` they have to they will have to use "private" API to build a initial base context, which obviously isn't great.

src/ast_traverse.mli
```diff
@@ -54,7 +54,7 @@ end
 
 class map_with_path : [string] map_with_context
 
-class map_with_code_path : [Code_path.t] map_with_context
+class map_with_expansion_ctxt_base : [Expansion_context.Base.t] map_with_context
```
diml: I suggest to call this `map_with_expansion_context`. I'm not sure the `_base` suffix is adding much and I find it slightly odd to mix full words and abbreviations.

src/expansion_context.mli
```diff
@@ -28,12 +52,15 @@ module Deriver : sig
...
+
   (** Wrap a [fun ~loc ~path] into a [fun ~ctxt] *)
   val with_loc_and_path : (loc:Location.t -> path:string -> 'a) -> (ctxt:t -> 'a)
 
   (**/**)
   (** Undocumented section *)
 
   (** Build a new expansion context with the given item location and code path *)
-  val make : derived_item_loc:Location.t -> code_path:Code_path.t -> t
+  val make : derived_item_loc:Location.t -> base:Base.t -> t
```
diml: If we want to make this a bit more future proof, we could add a unit argument. This way, we could add optional arguments to this function in the future.

diml:
`print_passes` is used to implement the `-print-passes` argument, so using a dummy value is fine. BTW, I'm starting to see your point about having an explicit public API. It's true that we have quite a lot of functions in ppxlib for which we don't want to provide long term guarantees, and trying to find the right design now is a bit annoying.

BTW, I've been thinking about these stability problems recently and I started writing a design document for a general mechanism: https://github.com/diml/api-stabiliser

We could try this idea by providing an explicit API as a separate ppxlib-v1 package and see how it goes.

src/ast_traverse.mli
NathanReb: I agree, I'll fix that!

src/expansion_context.mli
NathanReb: Good point, I'll add that!

PR #73, opened 2019-04-02
Small fixes to Metaquot manual section

NathanReb:
I've been a little prompt with the merge and didn't check how it was rendered by sphinx. This PR fixes some of the issues I introduced.

### DIFF
--- a/doc/ppx-for-plugin-authors.rst
+++ b/doc/ppx-for-plugin-authors.rst
@@ -25,6 +25,7 @@ expressions or patterns.
 
 The extension you should use depends on the type of AST node you're trying to write or to
 pattern-match over. You can use the following extensions with the following syntax:
+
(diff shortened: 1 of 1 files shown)


NathanReb:
I noticed a few details that might be worth fixing.

The ocaml syntax coloration used for the online version seems a little out of date, my local version is slightly better. I don't know exactly how it's deployed but it might be worth updating the `sphinx` or `sphinx_rtd_theme` versions.

The generated documentation is pretty thin and makes a lot of things wrap, including some uninterpreted code blocks which obviously isn't great. Do you if and how we could work around that?

diml:
I'm not too sure actually. I basically just copied the setup @rgrinberg wrote for dune. To upgrade the theme, we'd have to see what is supported by the readthedocs website.

NathanReb:
Ok I'll try to have a look at it!

PR #74, opened 2019-04-18
Update the list of attributes

xclerc:


diml:
Could you add `ocaml.immediate64` while you are at it? We will need that quite soon

PR #76, opened 2019-04-26
Add ocamlformat to the list of reserved namespace.

hhugo:
ocamlformat handles attributes of the following form 
```ocaml
[@@@ocamlformat "name=value"]
```


diml:
Thanks

PR #79, opened 2019-05-08
Remove the copy of the parser

xclerc:
This pull request remove the copy of the parser from ppxlib.

At the time of removal, there was only one difference with the
upstream parser: this copy did not rewrite `let x : t = e` to
`let x : t = (e : t)`.

note: `ppx_let` has been undoing the rewrite for some time,
but this change is quite probably a breaking one.


diml:
Well, if it breaks too much stuff, we could always undo the rewriting as a pre-pass in ppxlib. Should we merge this before #78? I feel like #78 would be smaller after this patch

xclerc:
Regarding the merge order, you are right that it would make the
diff a tad smaller, however I am pondering whether it would make
sense to:

- merge #78;
- make a release;
- merge #79;
- make a release.

My reasoning is that #78 is not breaking but #79, so it might be
nice for users/opam constraints to be able to use ppxlib with
OCaml 4.08.

That being said, I should have been more explicit. This is theoretically
a breaking change, but I guess only projects like `Async` and `Lwt`
are affected in practice. The former uses `ppx_let` and the latter
is not based on ppxlib (I think).

diml:
Ack, you are right that the order you propose is better for users. I'm going to review #78

xclerc:
In the spirit of multiple releases sorted by probability
of breaking things, I am polishing a third PR to switch
the selected AST from 4.07 to 4.08.

diml:
@xclerc let me know when it's ready for the next round of review

xclerc:
@diml I think it is. Since your initial review, I have only
pushed https://github.com/ocaml-ppx/ppxlib/pull/79/commits/591eec99e203567380fa6148c5d5a23b59d37d0e, to fix a problem I overlooked: once
we do not embed a parser anymore, we are in a situation
were the functions exposed for parsing are these from
the current compiler while the AST is fixed. Practically,
this means that one cannot *e.g.* use the provided traversal
functions on the result of `Parse.implementation`.

The aforementioned commit simply wrap the functions
from `Parse` and `Parser` to convert their result to the
fixed AST.

(note: this is exercised during the build of ppxlib itself,
so having a green CI means that the issue is actually
fixed.)

diml:
Ah, of course! That makes sense

masahi:
Hi, not sure my issue is related to this PR, but after ppxlib 0.8.0, I'm no longer able to install core_kernel with MetaOCaml switch. Until ppxlib 0.7.0 there was no issue. 

The problem is, core_kernel defines an infix operator ">." which is not compatible with metaocaml quotation. After I upgraded ppxlib to the latest one (0.8.1), ppx.exe gives a syntax error when processing files that contain such operator.

Is this PR the cause of my issue? If so, is there a walkaround?

diml:
The parser copy was essentially the upstream parser with just a couple of modifications. I don't recall anything about `>.` though.

Does MetaOCaml use a modified parser? That would explain why removing the parser copy in ppxlib broke building core_kernel with MetaOCaml.

trefis:
Yes, MetaOCaml uses a different parser.

There isn't much we can do here.

masahi:
ok, I understand this is a price to pay for using a compiler fork.

Thanks anyway. I'm new to OCaml and I'm blown away by its excellent tooling support.

PR #81, opened 2019-07-07
Some small tuning on top of select-ast-408

hhugo:
@xclerc, this is an experiment.  Let me know what you think..


xclerc:
This is quite probably better than the current implementation.

(The only downside being the changes against what is expected
to be a copy-paste from the compiler code base; it would be nice
to upstream `location_stack = location list` and its
consequences in the top part of `ast.ml`, eventually.)

PR #83, opened 2019-07-17
Set compiler's error-style and color through env vars in standalone mode

NathanReb:
OCaml 4.08 introduced to new flags allowing users to configure the error reporting style and use of colors.

Those can also be set via the env variables `OCAML_COLOR` and `OCAML_ERROR_STYLE`.

When using a standalone `ppxlib` driver, the appropriate flags weren't set according to those env variables and therefore the error reporting wasn't impacted.

A nice follow up could be to re-expose the command line flags from the standalone driver as well.

### DIFF
--- a/src/driver.ml
+++ b/src/driver.ml
@@ -1257,6 +1257,7 @@ let standalone_run_as_ppx_rewriter () =
 ;;
 
 let standalone () =
+  Compiler_specifics.read_clflags_from_env ();
   try
     if Array.length Caml.Sys.argv >= 2 &&
        match Caml.Sys.argv.(1) with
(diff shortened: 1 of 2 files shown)


NathanReb:
I realized as I was writing the PR text that this won't work as is since the compiler lib from older compilers don't expose those flags.

NathanReb:
From a quick look it seems like updating `ast/gen-compiler_specifics` is the way to go here.

NathanReb:
Cool, let's merge! Should we cut a patch release?

diml:
Seems good to me! Are you happy to take care of it? :)

PR #85, opened 2019-07-19
Opam file fixes

NathanReb:


NathanReb:
I also made some changes to the CI so we can catch those issues early on.

ppxlib.opam
```diff
@@ -11,12 +11,12 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 run-test: [
-  ["dune" "runtest" "-p" name "-j" jobs] { ocaml >= "4.06" }
+  ["dune" "runtest" "-p" name "-j" jobs] { ocaml:version >= "4.06" $ ocaml:version < "4.08" }
```
xclerc:
```suggestion
  ["dune" "runtest" "-p" name "-j" jobs] { ocaml:version >= "4.06" & ocaml:version < "4.08" }
```

NathanReb: Ouh yeah, typo! Thanks!

xclerc:
If were are changing the versions for Travis, we should
probably do the same for Appveyor. The changes to the
opam file should also probably be propagated to the
opam repository (actually, `build` is already gone).

NathanReb:
> if were are changing the versions for Travis, we should probably do the same for Appveyor. 

It doesn't actually change the ocaml versions we're using. The docker images for `ocaml/opam2` have pre-installed opam switches pointing to the latest bugfix release for every ocaml release from `4.04`. The `4.07` switch is a `4.07.1` one for example. I 'm not super familiar with the appveyor CI scripts but I'll take a look to see if we can achieve something similar there, although I doubt it since it's not using docker.

> The changes to the opam file should also probably be propagated to the opam repository

Agreed, I'll give it a go.

xclerc:
(I am deeply sorry - I edited your comment by mistake;
I did not even know it was possible.)

xclerc:
> It doesn't actually change the ocaml versions we're using. The docker images for `ocaml/opam2` have pre-installed opam switches pointing to the latest bugfix release for every ocaml release from `4.04`. The `4.07` switch is a `4.07.1` one for example. I 'm not super familiar with the appveyor CI scripts but I'll take a look to see if we can achieve something similar there, although I doubt it since it's not using docker.

I am not positive I understand what you mean, but my point
was: "if Travis tests 4.08 then AppVeyor should probably test
4.08".

NathanReb:
I added the 4.08 appveyor build and it all seems to work well!

The opam-repo PR brought up an issue with topfind in our tests in a system compiler switch. I'll take a look at that but I think it can be fixed separately.

NathanReb:
Is this good to go?

PR #86, opened 2019-07-25
do not produce a suprious empty correction in deriving_inline

aalekseyev:
Do not produce a suprious empty correction when deriving_inline expands into an extension that undergoes further expansion.

Empty corrections arise in particular when `deriving_inline` can see that the generated AST differs from the one in the source file, but after passing it through the styler it becomes the same.

I tried writing a test, but it doesn't seem to demonstrate the issue.

Apparently behavior is different between the two modes the driver is invoked:
If `diff_command = "-"` then we just "exit 0" after producing the corrected file (so the AST difference goes undetected).
If `diff_command <> "-"` then we "exit 1" so a very confusing error is reported (empty diff).

If I just run inline tests with dune, I get the former behavior so it's impossible to observe the improvement. The way we run it in Jane Street, `diff_command` is not `"-"` so we can see the difference.

Signed-off-by: Arseniy Alekseyev <aalekseyev@janestreet.com>

### DIFF
--- a/src/context_free.ml
+++ b/src/context_free.ml
@@ -599,13 +599,14 @@ class map_top_down ?(expect_mismatch_handler=Expect_mismatch_handler.nop)
         if not in_generated_code then
           Generated_code_hook.insert_after hook Signature_item item.psig_loc
             (Many extra_items);
+        let original_rest = rest in
         let rest = loop rest ~in_generated_code in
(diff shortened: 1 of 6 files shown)


diml:
Backported to ocaml-ppx/ppx

Issue #14, opened 2018-04-12
Remove `Context_free.Rule.attr*`


diml:
These are only used for `[@@deriving ...]` and it's unlikely they'll ever be used for something else. Now that `[@@deriving]` is part of ppxlib, I think removing them and specializing the implementation to only support `[@@deriving ...]` would make the code simpler.

NathanReb:
I'd like to give this a try in order to make my life easier with #61.

I'm trying to understand what the scope of this actually is.

My understanding so far is that we want to get rid of the `attr_*` smart constructor and replace them with something more in the line of `Context_free.Rule.extension` that would only allow you to declare a rule for the `deriving` attribute.
Removing these smart constructors is gonna break the API but I assume we don't consider this as a problem since they're probably only used in `Deriving`.

These smart constructors appear to be called a single time, on all the registered generators at once. I'm not sure exactly what the new API should look like.

I'm also guessing this is the occasion to leak a bit less about the internal types in `Context_free.Rule`.

diml:
> My understanding so far is that we want to get rid of the attr_* smart constructor and replace them with something more in the line of Context_free.Rule.extension that would only allow you to declare a rule for the deriving attribute.

I was thinking to completely remove them and make `Context_free` depend on `Deriving`. i.e. context_free.ml would directly call functions from `Deriving`.

> Removing these smart constructors is gonna break the API but I assume we don't consider this as a problem since they're probably only used in Deriving.

Agreed.

BTW, I remember that I tried this last year. My attempt is in the branch `simplify-context-free`. I don't remember exactly what's the status of this branch is, but it might be worth having a look. I couldn't easily get rid of the circular dependency between Context_free, Deriving and Driver, which was annoying and required forward references, but I don't know if we can do better though.

kwshi:
Hmm-- I'm working on a project that requires `[@@deriving]`-like features, except on _module declarations_.  As far as I can tell, `[@@deriving]` only currently supports types.  If there is no plan to develop the `attr`-libraries and use only `[@@deriving]`, what shall I do?

diml:
I know ppx_deriving allows this. It seems fine to add the same support in ppxlib.

kwshi:
[...]

PR #92, opened 2019-08-20
Simplify rule generation for deriving extensions

rgrinberg:
Previously, the attributes, rules, and the generators would all be
created separately and then composed. That's a bit error prone as there
too many places where one needs to update the codebase to add a new type
of deriving plugin. Now one simply needs to specify the type, and the
various rule generators, and code expanders. It's easier to get this all
right in one place.

I'm also wondering if it makes sense to clean this code up further with cinaps. WDYT?

### DIFF
--- a/src/deriving.ml
+++ b/src/deriving.ml
@@ -496,20 +496,6 @@ let mk_deriving_attr context ~prefix ~suffix =
     (fun x -> x)
 ;;
 
-module Attr = struct
-  let suffix = ""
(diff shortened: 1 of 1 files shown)


diml:
After merging, could you also apply this patch in ocaml-ppx/ppx?

BTW, one thing I wanted to do in the past is specialise `Context_free` for `Deriving`. Currently, `Context_free` is agnostic of the `Deriving` module, but that makes its code more complicated for no benefit. (This is discussed in #14)

PR #94, opened 2019-08-22
Add support for deriving on module types

### LINKED ISSUE #93: [deriving] add support for module types
To achieve feature parity with ppx_deriving, we should add this feature to ppxlib: https://github.com/ocaml-ppx/ppx_deriving/pull/94

rgrinberg:
This PR adds support for deriving generators on module type declarations. For example:

```ocaml
module type Foo = sig .. end [@@deriving bar]
```

It is necessary to achieve feature parity with ppx_deriving.

Fix #93 

cc @kwshi who was interested in this feature.

### DIFF
--- a/test/deriving/test.ml
+++ b/test/deriving/test.ml
@@ -78,6 +78,5 @@ end = struct
   let y = 42
 end
 [%%expect{|
-Line _, characters 42-42:
-Error: ppxlib: [@@@deriving.end] attribute missing
+module Y : sig module type X = sig  end val y : int end
 |}]
--- a/test/deriving/test.ml
+++ b/test/deriving/test.ml
@@ -10,6 +10,8 @@ let foo =
   Deriving.add "foo"
     ~str_type_decl:(Deriving.Generator.make_noarg
                       (fun ~loc ~path:_ _ -> [%str let x = 42]))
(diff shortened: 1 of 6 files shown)


rgrinberg:
My last commit adds a test which demonstrates an error:

```ocaml
module Y : sig
  module type X = sig end [@@deriving mtd]
end = struct
  module type X = sig end [@@deriving mtd]
end
[%%expect{|
Line _, characters 42-42:
Error: ppxlib: [@@@deriving.end] attribute missing
|}]
```

Can anyone give a hint why this is happening?

rgrinberg:
Actually, this seems to be an issue with the signature using tests.

test/deriving/test.ml
```diff
@@ -40,3 +54,29 @@ type nonrec int = int [@@deriving foo, bar]
...
 |}]
+
+module Foo_sig : sig
+  type t [@@deriving foo]
+end = struct
+  type t
+end
+[%%expect{|
+Line _, characters 25-25:
+Error: ppxlib: [@@@deriving.end] attribute missing
```
NathanReb: Isn't it a bit weird that this error persists? I'd expect an error saying the module doesn't provide the `val y : int` required by the signature.

We should probably have a look at that separately

PR #95, opened 2019-08-23
Add a quoting API inspired by Ppx_deriving

rgrinberg:
This PR introduces a hygiene API in the spirit of Ppx_deriving.quoter.
The main differences are:

* We can't use metaquot
* There are no optional arguments
* We don't wrap stuff with modules as this can be easily handled
  separately


rgrinberg:
It seems like I'm unable to `#install_printer` on the `Pprintast` printers. Any idea as to why this isn't working?

src/quote.mli
```diff
@@ -0,0 +1,9 @@
+open Import
```
diml: This mli needs documentation, otherwise it's not easy to understand what it does without looking at the implementation

diml:
Not sure about the install_printer problem. We should switch all these tests to ppx_expect at some point anyway

src/quote.mli
rgrinberg: Yup, I will write some documentation today.

rgrinberg:
I've written up some docs. Shall I rename the module to `Quoter`? I'm not sure if `Quote` or `Quoter` is more natural here.

diml:
`Quoter` seems like a better name indeed. Appart from that, it looks good to merge.

rgrinberg:
@diml I renamed it.

diml:
And for people who might be watching this PR: we are adding this module to make it easier to port plugins from `Ppx_deriving` to `Ppx.Deriving`

rgrinberg:
Thanks for the review, I'll prepare a corresponding PR to ppx.

PR #97, opened 2019-08-29
update dune code in the doc of @@deriving_inline

Khady:
The syntax used for the `lint` part of the `library` stanza was not working with a recent version of dune.

Signed-off-by: Louis Roché <louis@louisroche.net>


rgrinberg:
Could you apply this patch to https://github.com/ocaml-ppx/ppx as well? (if it's applicable)

PR #98, opened 2019-09-13
Fix a bug in Longident.parse

NathanReb:
This PR fixes a bug in `Longident.parse` and adds a couple regression tests.

`Longident.parse` doesn't not support Lapply-type longident (ie `A.B(C)`). That is fine but it used to return inconsistent values in such cases instead of raising an exception.

I fixed the bug and also made it clear in the documentation that these aren't handled as this function is part of the public API of `ppxlib`.

### DIFF
--- a/src/longident.ml
+++ b/src/longident.ml
@@ -57,18 +57,20 @@ let parse_simple s =
 
 (* handle ["A.B.(+.+)"] or ["Vec.(.%.()<-)"] *)
 let parse s =
+  let invalid () =
+    invalid_arg (Printf.sprintf "Ppxlib.Longident.parse: %S" s)
+  in
   match String.index s '(' with
   | None -> parse_simple  s
   | Some l -> match String.rindex s ')' with
-    | None -> invalid_arg "Ppxlib.Longident.parse"
+    | None -> invalid ()
     | Some r ->
-      if Int.( r <> String.length s - 1 ) then
-        invalid_arg "Ppxlib.Longident.parse";
+      if Int.( r <> String.length s - 1 ) then invalid ();
       let group = if Int.(r = l + 1) then "()" else
           String.strip (String.sub s ~pos:(l+1) ~len:(r-l-1))
       in
       if Int.(l = 0) then Lident group
-      else if Char.(s.[l - 1] <> '.') then invalid_arg "Ppxlib.Longident.parse"
+      else if Char.(s.[l - 1] <> '.') then invalid ()
       else
         let before = String.sub s ~pos:0 ~len:(l-1) in
         match String.split before ~on:'.' with
--- a/src/longident.ml
+++ b/src/longident.ml
@@ -60,11 +60,10 @@ let parse s =
   let invalid () =
     invalid_arg (Printf.sprintf "Ppxlib.Longident.parse: %S" s)
   in
-  match String.index s '(' with
-  | None -> parse_simple  s
-  | Some l -> match String.rindex s ')' with
-    | None -> invalid ()
-    | Some r ->
+  match String.index s '(', String.rindex s ')' with
+  | None, None -> parse_simple  s
+  | None, _ | _, None -> invalid ()
+  | Some l, Some r ->
       if Int.( r <> String.length s - 1 ) then invalid ();
       let group = if Int.(r = l + 1) then "()" else
           String.strip (String.sub s ~pos:(l+1) ~len:(r-l-1))
--- a/src/longident.ml
+++ b/src/longident.ml
@@ -65,9 +65,12 @@ let parse s =
       if Int.( r <> String.length s - 1 ) then
         invalid_arg "Ppxlib.Longident.parse";
       let group = if Int.(r = l + 1) then "()" else
-          String.strip (String.sub s ~pos:(l+1) ~len:(r-l-1)) in
-      if Int.(l = 0) then Lident group else
+          String.strip (String.sub s ~pos:(l+1) ~len:(r-l-1))
+      in
+      if Int.(l = 0) then Lident group
+      else if Char.(s.[l - 1] <> '.') then invalid_arg "Ppxlib.Longident.parse"
+      else
         let before = String.sub s ~pos:0 ~len:(l-1) in
         match String.split before ~on:'.' with
-        | [] -> Lident group
+        | [] -> assert false
         | s :: l -> Ldot(unflatten ~init:(Lident s) l, group)
--- a/test/base/dune
+++ b/test/base/dune
@@ -3,8 +3,10 @@
  (deps
   (:test test.ml)
   (glob_files %{project_root}/src/.ppxlib.objs/byte/*.cmi))
- (action (chdir %{project_root}
-          (progn
-           (ignore-outputs
-            (run %{project_root}/test/expect/expect_test.exe %{test}))
-           (diff? %{test} %{test}.corrected)))))
+ (action
+  (chdir %{project_root}
+   (progn
+    (setenv OCAMLRUNPARAM ""
+     (ignore-outputs
+      (run %{project_root}/test/expect/expect_test.exe %{test})))
+    (diff? %{test} %{test}.corrected)))))
--- a/test/base/dune
+++ b/test/base/dune
@@ -6,7 +6,6 @@
  (action
   (chdir %{project_root}
    (progn
-    (setenv OCAMLRUNPARAM ""
-     (ignore-outputs
-      (run %{project_root}/test/expect/expect_test.exe %{test})))
+    (ignore-outputs
+     (run %{project_root}/test/expect/expect_test.exe %{test}))
     (diff? %{test} %{test}.corrected)))))
--- a/src/longident.mli
+++ b/src/longident.mli
@@ -11,5 +11,11 @@ include Comparable.S with type t := t
 
 val flatten_exn : t -> string list
 val last_exn : t -> string
+
+(** Parses the given string as a longident, properly handling infix operators
+    which may contain '.'.
+    Note that it does not parse [Lapply _] longidents and will raise
+    [Invalid_argument _] if passed values such as ["A(B)"]. *)
 val parse : string -> t
+
 val name : t -> string
--- a/test/base/test.ml
+++ b/test/base/test.ml
@@ -113,6 +113,11 @@ let _ = convert_longident "A.B(C)"
 Exception: Invalid_argument "Ppxlib.Longident.parse: \"A.B(C)\"".
 |}]
 
+let _ = convert_longident ")"
+[%%expect{|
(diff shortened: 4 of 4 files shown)


NathanReb:
You'll note that the test aren't great as we're getting the stacktrace in there and it's likely to change so we should probably tweak `expect.ml` or make sure we don't get backtraces in the tests.

NathanReb:
Well, looks like I'm gonna need to fix that backtrace issue anyway.

src/longident.ml
```diff
@@ -65,9 +65,12 @@ let parse s =
       if Int.( r <> String.length s - 1 ) then
         invalid_arg "Ppxlib.Longident.parse";
       let group = if Int.(r = l + 1) then "()" else
-          String.strip (String.sub s ~pos:(l+1) ~len:(r-l-1)) in
-      if Int.(l = 0) then Lident group else
+          String.strip (String.sub s ~pos:(l+1) ~len:(r-l-1))
+      in
+      if Int.(l = 0) then Lident group
+      else if Char.(s.[l - 1] <> '.') then invalid_arg "Ppxlib.Longident.parse"
```
rgrinberg: Would it make sense to include the string in the exception error message?

NathanReb: I believe so! Good catch! I'll change that.

test/base/dune
```diff
@@ -3,8 +3,10 @@
...
   (glob_files %{project_root}/src/.ppxlib.objs/byte/*.cmi))
- (action (chdir %{project_root}
-          (progn
-           (ignore-outputs
-            (run %{project_root}/test/expect/expect_test.exe %{test}))
-           (diff? %{test} %{test}.corrected)))))
+ (action
+  (chdir %{project_root}
+   (progn
+    (setenv OCAMLRUNPARAM ""
```
rgrinberg: I personally prefer to include things like this in the code. I.e:

```
let () = Printexc.record_backtrace false
```

However, this is a matter of taste so you can have the final word here.

diml:
> Also, doesn't compiler-libs provide an API for this?

Eventually, we want to detach ourselves completely from compiler-libs in order to be shielded from the unstable nature of the compiler libraries. astlib will be the only link. So it's better to not rely on functions from compiler-libs.

test/base/dune
NathanReb: That's what I tried at first but couldn't get it working for some reasons, I'll try again!

NathanReb:
> Looks good to me. I added a test case for an invalid identifier that seems to be recognized as such. I don't think this PR introduces this bug so it's up to you if you want to fix it or not.

Good catch! We might as well fix it now while we're at it, I'll take a look!

NathanReb:
Pushed a fix for the `")"` case! I think it also slightly simplifies the code but I agree it's still quite complex and hard to follow!

gasche:
I thought this was an amusing problem so I wrote a simple parser, without using `String.index` which make the parsing logic fragile.

```ocaml
type ident =
  | Ident of string
  | Dot of ident * string
  | Apply of ident * ident

let parse : string -> ident = fun str ->
  let len = String.length str in
  (* the indices returned by all functions are the next position
     strictly *after* the end of the lexeme they parsed
     
     (consume_op parses the *inside* of an operator "(...)", so
      it starts after the "(" returns the position of the ")").)
   *)
  let rec parse i =
    let w, i = parse_word i in
    parse_after (Ident w) i
  and parse_word i =
    assert (i < len);
    match str.[i] with
      | 'A'..'Z' | 'a'..'z' ->
         let after = consume_alphanum (i + 1) in
         String.sub str i (after - i), after
      | '(' ->
         let closing = consume_op (i + 1) in
         assert (str.[closing] = ')');
         let after = closing+1 in
         String.sub str i (after - i), after
      | c ->
         Printf.ksprintf invalid_arg "parse: invalid word start %S"
           (String.make 1 c)
  and consume_alphanum i =
    if i = len then i
    else match str.[i] with
      | 'a'..'z' | 'A'..'Z' | '0'..'9' | '\'' | '_' -> 
         consume_alphanum (i+1)
      | _ -> i
  and consume_op i =
    (* precondition: we are after an opening paren,
       postcondition: we stop at the corresponding closing paren *)
    if i = len then i
    else match str.[i] with
      | ')' -> i
      | '(' ->
         let closing_paren = consume_op (i+1) in
         consume_op (closing_paren + 1)
      | _ -> consume_op (i + 1)
  and parse_after before i =
    if i = len then before, i
    else match str.[i] with
      | '(' ->
         let i = i+1 in
         let arg, i = parse i in
         assert (str.[i] = ')');
         let i = i+1 in
         parse_after (Apply(before, arg)) i
      | '.' ->
         let i = i+1 in
         let w,i = parse_word i in
         parse_after (Dot(before, w)) i
      | c -> before, i
  in
  let ident, i = parse 0 in
  assert (i = len);
  ident

let () =
  assert (parse "A" = Ident("A"));
  assert (parse "A.B.C" = Dot(Dot(Ident "A", "B"), "C"));
  assert (parse "A(B)" = Apply(Ident("A"),Ident("B")));
  assert (parse "A.B(C.D).E" =
            Dot(Apply(Dot(Ident("A"), "B"), Dot(Ident "C", "D")), "E"));
  assert (parse "A.(+)" = Dot(Ident "A", "(+)"));
  assert (parse "A.(.())" = Dot(Ident "A", "(.())"));
  assert (parse "A(B).(.())" =
            Dot((Apply(Ident "A", Ident "B"), "(.())")));
  assert (parse "A(B).(.()).C" =
            Dot(Dot(Apply(Ident "A", Ident "B"), "(.())"), "C"));
  assert (parse "A(B).(.())(C)" =
            Apply(Dot(Apply(Ident "A", Ident "B"), "(.())"), Ident "C"));
  ()
```

test/base/dune
NathanReb: I moved it to the test file although what I really wanted to do was to disable backtraces for all expect tests in `expect_test.mll` but I couldn't find a better solution than using `Toploop.execute_phrase` and that's not really satisfying.

Now that I think about it I'm not even sure we want to disable backtraces for all expect test so I'll stick to that for now!

NathanReb:
@gasche's function indeed works better but I agree with @rgrinberg that at this point using a lexer would probably make more sense.

gasche:
Be careful that you need a lexer *and* a parser (or a token-less parser), given that the grammar has recursive structure. (You can use a yacc-like tool for the parser or write it by hand.) The end result would feel less manual, but (1) the mix of tools gives an impression of complexity/verbosity and (2) you have more build-system plumbing to deal with. I'm not fully convinced that would really feel more convenient -- but it's worth trying!

(If you want to avoid bugs, I would recommend testing whichever implement you choose using property-based testing (a quicheck-like tool, for example [QCheck](http://c-cube.github.io/qcheck/)): it's easy to generate longidents and print them to string, you can check that parsing them back gives the expected result. It's a half-an-hour exercise that would give strong confidence.)

diml:
Personally, I'm not too worried about the build system plumbing (it's one line in the dune file) or the mix of a lexer + a hand written parser given that the grammar is simple (that's what we do in dune for s-expressions for instance). However, I am a bit more concerned about performances. IIRC, we call this function quite often and using a lexer requires setting up a lexing buffer, copying the string, etc... This can be costly for such a simple function. It feels like using an elephant to crunch a fly.

So personally I'd prefer the hand-written parsing function + tests.

NathanReb:
Ok, sounds good to me, I'll use @gasche's implementation then!

How would you feel about me adding a `fuzz/` folder and some `crowbar` tests there? We obviously wouldn't run those for every PR.

diml:
If it's a 5 minute job, I'd say go for it. Otherwise, let's keep it in mind for later. Progress on the ppx project is more important.

In any case, we should add @gasche's tests as expectation tests to the testsuite.

NathanReb:
Ok sure, let's get this merged first, we can work on that later!

NathanReb:
I just tried @gasche's implementation and it contains a few tiny bugs. I'll work on fixing those but in the meantime I think it makes sense to merge this as it is as it fixes existing bugs and add a bit of documentation!

diml:
Sounds good to me!

gasche:
I did write a small random generator to check, and the one bug it found is that it fails to parse identifiers starting with an underscore (`_foo`). To fix this, `parse_word` should be modified to allow `_` in the first case.

My QCheck generator is as follows. (Initially I wrote a Crowbar generator, but I didn't manage to get Crowbar to shrink counter-examples, so I moved back to QCheck):

```ocaml
let rec bprint_ident buf =
  let print fmt = Printf.bprintf buf fmt in
  function
  | Ident str -> print "%s" str
  | Dot (path, id) -> print "%a.%s" bprint_ident path id
  | Apply (f, arg) -> print "%a(%a)" bprint_ident f bprint_ident arg

let string_of_ident id =
  let buf = Buffer.create 10 in
  bprint_ident buf id;
  Buffer.contents buf

let ident_gen_qcheck =
  let open QCheck.Gen in
  let choose = oneof in
  let between c c' =
    let c, c' = if int_of_char c <= int_of_char c' then c, c' else c', c in
    let low = min (int_of_char c) (int_of_char c') in
    let high = max (int_of_char c) (int_of_char c') in
    map char_of_int (int_range low high) in
  let upper = between 'A' 'Z' in
  let lower = between 'a' 'z' in
  let number = between '0' '9' in
  let underscore = pure '_' in
  let alphanum =
    choose [
        upper;
        lower;
        number;
        underscore;
        pure '\'';
      ] in
  let make_word (first, rest) =
    first::rest
    |> List.map (String.make 1)
    |> String.concat ""
  in
  let lident = map make_word (pair (choose [lower; underscore]) (list alphanum)) in
  let uident = map make_word (pair upper (list alphanum)) in
  let operator =
    oneofl [
        "(+)";
        "(*)";
        "( * )";
        "(!)";
        "(.())";
        "(.()<-)";
        "(mod)";
        "(let!)";
      ] in
  let lident_or_op = choose [lident; operator] in
  let simple_ident = choose [uident; lident_or_op] in
  let rec ident st =
    choose [
        map (fun id -> Ident id) simple_ident;
        map (fun (path, id) -> Dot(path, id)) (pair ident simple_ident);
        map (fun (f, arg) -> Apply(f, arg)) (pair ident ident);
    ] st in
  ident


let test_qcheck () =
  let ident_arb = QCheck.make ~print:string_of_ident ident_gen_qcheck in
  let test =
    QCheck.Test.make ~name:"parse" ident_arb @@ fun ident ->
    parse (string_of_ident ident) = ident
  in ignore (QCheck_runner.run_tests ~verbose:true [ test ])
```

This generates identifiers that are not actually valid OCaml longidents, such as `foo.bar` (in the Dot construct, the left-hand-side must be a module path, not an arbitrary identifier). It would be easy to restrict the generator into the valid-OCaml subset, but the parser does accept more than that.

PR #99, opened 2019-10-07
Update ppxlib to use cinaps clause and helpers library.

ceastlund:
Upstreaming this from code written at janestreet (adapted for dune rather than jenga).

### DIFF
--- a/src/cinaps/dune
+++ b/src/cinaps/dune
@@ -1,4 +1,3 @@
 (library
  (name ppxlib_cinaps_helpers)
- (public_name ppxlib.cinaps_helpers)
  (libraries re))
--- a/src/cinaps/dune
+++ b/src/cinaps/dune
@@ -1,0 +1,4 @@
+(library
+ (name ppxlib_cinaps_helpers)
+ (public_name ppxlib.cinaps_helpers)
+ (libraries re))
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,10 @@ env:
   - PACKAGE="ppxlib"
   matrix:
   - OCAML_VERSION="4.04"
+    TESTS=false
   - OCAML_VERSION="4.05"
+    TESTS=false
   - OCAML_VERSION="4.06"
+    TESTS=false
   - OCAML_VERSION="4.07"
   - OCAML_VERSION="4.08"
--- a/dune-project
+++ b/dune-project
@@ -1,2 +1,3 @@
 (lang dune 1.0)
 (name ppxlib)
+(using cinaps 1.0)
--- a/ppxlib.opam
+++ b/ppxlib.opam
@@ -22,6 +22,7 @@ depends: [
   "ppx_derivers"            {>= "1.0"}
   "stdio"                   {>= "v0.11.0"}
   "ocamlfind"               {with-test}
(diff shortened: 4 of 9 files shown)


src/cinaps/dune
```diff
@@ -0,0 +1,4 @@
+(library
+ (name ppxlib_cinaps_helpers)
+ (public_name ppxlib.cinaps_helpers)
```
diml: Why is this library public?

ceastlund: Hah! Sorry, I was just used to writing dune files that way.

diml:
I just submitted a new release of cinaps with the staged mode: https://github.com/ocaml/opam-repository/pull/15015. Once it's accepted we can restart Travis and the build should succeed

ceastlund:
Weird, it worked for some versions of ocaml and not for others. @diml, do you have any insight into the errors?

diml:
The cinaps dependency is missing for tests. I tried adding it. I don't know why it succeed at all on some of the versions :confused:

diml:
I restarted the build now that the new cinaps has been released. Let see how it goes. Could you sign your commits BTW?

ceastlund:
Re-pushed with signed-off commits.

diml:
Good to merge as soon as CI passes

diml:
I added you to the repository so you should be able to merge this PR

ceastlund:
Still failing. Looks like some package constraint problems. Is this just due to the cinaps pin we've been doing?

diml:
I found the issue: https://github.com/ocaml/opam-repository/pull/15209

ceastlund:
Do we need to wait for an opam release to pick up the fix?

diml:
Yh, the opam PR has been merged. I restarted Travis, hopefully it should succeed this time!

diml:
And we can ignore the AppVeyor failures for now, they seem unrelated

diml:
Ok, now Travis is happy except for older OCaml versions. Indeed, cinaps doesn't build with OCaml < 4.07...

diml:
hmm, this is kind of annoying. I feel like the simplest choices are:

1. drop compat for OCaml < 4.07 in ppxlib
2. adapt cinaps so that it works with OCaml >= 4.04

1 obviously requires much less work than 2. What do others think? /cc @trefis @xclerc

ceastlund:
I mean a lot of the new ppx compatibility work is aimed at making ppx compatible over many versions, so it seems worth some effort to keep 4.04+ compatibility, doesn't it?

diml:
Yes, well after you can also say that if you use an older compiler then you use an older ppxlib. We just need to keep a window of reasonable size. That'll be the same for ppx, initially we should support a wide range of version but eventually will start dropping support for older versions of the compiler.

ceastlund:
I do not appear to have write access, so I can't merge this.

diml:
github says that you haven't accepted the invitation to join the project yet.

Just had a chat with @trefis and @xclerc, the easiest way to keep the CI green without more work is to only do the build with older compilers in Travis and build&run the tests only with newer compilers.

diml:
Travis is happy now. Merging

PR #102, opened 2019-11-19
Specify relative order between `rules` and `impl`

aalekseyev:
Specify relative order between `rules` and `impl` in the preprocessing pipeline.

### DIFF
--- a/src/driver.mli
+++ b/src/driver.mli
@@ -81,8 +81,11 @@ end
     [lint_impl] and [lint_intf] are applied to the unprocessed source. Errors they return
     will be reported to the user as preprocessor warnings.
 
-    [preprocess_impl] and [preprocess_intf] are applied after linters,
-    but before other transformations.
+    Rewritings are applied in the following order:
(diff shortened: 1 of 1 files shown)


src/driver.mli
```diff
@@ -81,8 +81,11 @@ end
     [lint_impl] and [lint_intf] are applied to the unprocessed source. Errors they return
     will be reported to the user as preprocessor warnings.
 
-    [preprocess_impl] and [preprocess_intf] are applied after linters,
-    but before other transformations.
+    Rewritings are applied in the following order:
+    - linters ([lint_impl], [lint_intf])
+    - preprocessing ([preprocess_impl], [preprocess_intf])
+    - contest-independent rules ([rules], [extensions])
```
trefis: Context*?

aalekseyev: fixed

aalekseyev: Uh, it was merged before I "fixed" it. I'll make another PR.

PR #106, opened 2019-11-22
Preserve module-level attributes

ceastlund:
Currently `~enclose_impl` and `~enclose_intf` transformations can break module-level attributes like `@@@alert` because they move them away from the top of the signature/structure, where the compiler looks for them.

This patch changes `~enclose_intf` and `~enclose_impl` to leave attributes at the start of a structure/signature if they started there.

### DIFF
--- a/src/driver.ml
+++ b/src/driver.ml
@@ -163,8 +163,13 @@ module Transform = struct
           hook.f context loc (Many l));
       (header, footer)
     in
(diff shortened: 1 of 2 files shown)


diml:
That looks good to me, but can this be documented in the mli?

ceastlund:
Updated mli, and propagated the match change too.

PR #109, opened 2019-12-03
Support OCaml 4.10

xclerc:
This pull request contains minor changes to adapt to the
upcoming OCaml 4.10.

This is work in progress because we need a 4.10-compatible
of Base in order to link.


xclerc:
`ppxlib` can be build on 4.10 using https://github.com/janestreet/base/tree/410.

Issue #108, opened 2019-11-29
Appveyor fails on Sys_error("C:\cygwin64\tmp\cinapsf14ff5.mli: Permission denied")


XVilka:
```
+ C:\cygwin64\home\appveyor\.opam\ocaml-variants.4.07.1+mingw32c\bin\dune.exe "runtest" "-p" "ppxlib" "-j" "2" (CWD=C:/cygwin64/home/appveyor/.opam/ocaml-variants.4.07.1+mingw32c/.opam-switch/build/ppxlib.0.9.0)
-       cinaps alias src/runtest (exit 2)
- (cd _build/default/src && .cinaps/cinaps.exe -diff-cmd -)
- Fatal error: exception Sys_error("C:\cygwin64\tmp\cinapsf14ff5.mli: Permission denied")
-       cinaps alias src/runtest (exit 2)
- (cd _build/default/src && .cinaps/cinaps.exe -diff-cmd -)
- Fatal error: exception Sys_error("C:\cygwin64\tmp\cinapsf14ff5.mli: Permission denied")
[ERROR] The compilation of ppxlib failed at "C:\\cygwin64\\home\\appveyor\\.opam\\ocaml-variants.4.07.1+mingw32c\\bin\\dune.exe runtest -p ppxlib -j 2".
#=== ERROR while compiling ppxlib.0.9.0 =======================================#
# context              2.0.5 | win32/x86_64 | ocaml-variants.4.07.1+mingw32c | pinned(file://C:/projects/ppxlib)
# path                 ~/.opam/ocaml-variants.4.07.1+mingw32c/.opam-switch/build/ppxlib.0.9.0
# command              C:\cygwin64\home\appveyor\.opam\ocaml-variants.4.07.1+mingw32c\bin\dune.exe runtest -p ppxlib -j 2
# exit-code            1
# env-file             ~/.opam/log/ppxlib-992-9b8556.env
# output-file          ~/.opam/log/ppxlib-992-9b8556.out
### output ###
#       cinaps alias src/runtest (exit 2)
# (cd _build/default/src && .cinaps/cinaps.exe -diff-cmd -)
# Fatal error: exception Sys_error("C:\cygwin64\tmp\cinapsf14ff5.mli: Permission denied")
<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+- The following actions failed
| - build ppxlib 0.9.0
+- 
- No changes have been performed
```

[...]

PR #110, opened 2019-12-18
Add full code path to deriving expansion context

matthewelse:
A PPX deriver that I'm developing could benefit from knowing the name of the enclosing module, which it uses to suggest a name for the derived structure.

This PR includes:
1. An addition to Code_path to keep track of the current parent module
2. An additional API call to create a Deriver.Generator that takes [Code_path.t]s, rather than [string]s.

Similarly, we could also do the same with extensions, but that isn't included currently in this PR.

### DIFF
--- a/src/code_path.ml
+++ b/src/code_path.ml
@@ -6,6 +6,7 @@ type t =
   ; submodule_path : string loc list
   ; value : string loc option
   ; in_expr : bool
+  ; parent_module : string
   }
 
 let top_level ~file_path =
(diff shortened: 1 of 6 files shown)


matthewelse:
I think that the build failure on appveyor is unrelated to my changes.

XVilka:
AppVeyor failure is unrelated, see https://github.com/ocaml-ppx/ppxlib/issues/108

NathanReb:
Please note that the latest versions of ppxlib already expose what you need. I think what you want is just a new function in `Code_path` that returns the enclosing module name. That doesn't require any addition to the record type.

If you use either `Extension.V3.declare` or `Deriving.Generator.V2.make` the expand function will receive an `Expansion_context` from which you can get the `Code_path.t`.

Just to clarify, if you consider the following `a.ml` file:
```ocaml
module B = struct
  module C = struct
    [%here]
  end
end
```
the enclosing module name should be `C`, am I correct?

You could write that function like this:
```ocaml
let enclosing_module t =
  match t with
  | {submodule_path = []; main_module_name; _} -> main_module_name
  | {submodule_path = m::_; _} -> m
```

I believe that should be all you need but please let me know if I'm missing something.

Also note that until this gets released you can already implement that function in your deriver!

matthewelse:
Pretty sure this solves my problem! Thanks

PR #111, opened 2019-12-22
Add tests for parsing longident's with .

rgrinberg:
There are some identifiers that have the `.` character. We test cases with `+.`
and `*.`

It seems like the `+.` isn't handled correctly when there's no module prefixing.

### DIFF
--- a/src/longident.ml
+++ b/src/longident.ml
@@ -51,10 +51,14 @@ let parse s =
   let invalid () =
     invalid_arg (Printf.sprintf "Ppxlib.Longident.parse: %S" s)
   in
-  match (String.index_opt s '(', String.rindex_opt s ')') with
-  | None, None -> parse_simple s
-  | None, _ | _, None -> invalid ()
-  | Some l, Some r -> (
+  if String.length s < 1 then invalid ();
+  let open_par = String.index_opt s '(' in
+  let close_par = String.index_opt s ')' in
+  match (s.[0], open_par, close_par) with
+  | ('A' .. 'Z' | 'a' .. 'z' | '_'), None, None -> parse_simple s
+  | _, None, None -> Lident s (* This is a raw operator, no module path *)
+  | _, None, _ | _, _, None -> invalid ()
+  | _, Some l, Some r -> (
       if Int.(r <> String.length s - 1) then invalid ();
       let group =
         if Int.(r = l + 1) then "()"
--- a/test/base/test.ml
+++ b/test/base/test.ml
@@ -121,6 +121,11 @@ let _ = convert_longident "+."
 ("( + ).", Ppxlib.Longident.Ldot (Ppxlib.Longident.Lident "+", ""))
 |}]
 
+let _ = convert_longident "(+.)"
+[%%expect{|
+- : string * longident = ("( +. )", Ppxlib.Longident.Lident "+.")
+|}]
+
 let _ = convert_longident "Foo.(+.)"
 [%%expect{|
 - : string * longident =
--- a/test/base/test.ml
+++ b/test/base/test.ml
@@ -114,11 +114,9 @@ let _ = convert_longident ")"
 Exception: Invalid_argument "Ppxlib.Longident.parse: \")\"".
 |}]
 
-(* FIXME this is a bug *)
 let _ = convert_longident "+."
 [%%expect{|
-- : string * longident =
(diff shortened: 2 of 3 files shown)


rgrinberg:
@gasche, i'm a bit curious, why is the implementation of `Longident.parse` in compiler-libs not complete? I'm being bit by a bug in it in the context of merlin as well.

gasche:
`parse` is a misnomer for this function that is used in very restricted ways inside the compiler codebase, only to parse paths of the form `Foo.Bar.baz`. It was never designed for external usage.

The robust code to parse longidents is in the parser, but it separates in many smaller categories (`mod_longident` similarly does not allow module applications, `mod_ext_longident` does, and both of those only parse module identifiers (the last component is uppercased) and get turned into various non-module-identifiers (for values, types, etc.)) that are finer grained than the type `Longident.t` represents. We could propose to expose this parser logic, but:
- I don't know what would be a good API for it. (We could expose the various kind of longidents, one for each namespace: modules, types and module types, values, classes and class types.)
- We cannot expose this from `Longident.parse`, because it would create a cyclic dependency between the parser and `Longident`. So users would have to access `Parse.*_longident` instead, and build a lexing buffer themselves.

Finally, we could also try to use a hand-written parser for `Longident` -- for example the one I proposed in https://github.com/ocaml-ppx/ppxlib/pull/98#issuecomment-532662179 -- but it seems weird to duplicate logic that we already have someplace else, with inferior technology, for something we don't need inside the compiler ocdebase.

src/longident.ml
```diff
@@ -51,10 +51,14 @@ let parse s =
   let invalid () =
     invalid_arg (Printf.sprintf "Ppxlib.Longident.parse: %S" s)
   in
-  match (String.index_opt s '(', String.rindex_opt s ')') with
-  | None, None -> parse_simple s
-  | None, _ | _, None -> invalid ()
-  | Some l, Some r -> (
+  let open_par = String.index_opt s '(' in
+  let close_par = String.index_opt s ')' in
+  let first_char_is_normal = is_normal_ident_char s.[0] in
```
panglesd, line 56: Since we have the `invalid` function, it makes sens to test whether `s` has length at least 1.

test/base/test.ml
```diff
@@ -114,6 +114,29 @@ let _ = convert_longident ")"
 Exception: Invalid_argument "Ppxlib.Longident.parse: \")\"".
 |}]
 
+(* FIXME this is a bug *)
```
panglesd, line 117: What is the bug exactly? It would make sense to add more details here!

panglesd (approved):
About the "add tests" part: Let's merge, it's good to have more tests!

About the "fix a longident bug about unparenthesized operators": that looks good and fixing a bug. I don't know enough the possible syntax of longidents to be confident there aren't other bugs though!

test/base/test.ml
NathanReb: Ouh my bad, I forgot to remove that comment, I added the fix to the PR!

NathanReb: To answer your question, the bug discovered by the initial test addition was that `+.` was parsed as:
`Ldot (Lident "+", "")` instead of `Lident "+."`.

I added a fix for this.

In the long run we might want to make this function more robust, as was mentioned in https://github.com/ocaml-ppx/ppxlib/pull/98.

src/longident.ml
NathanReb: Good point, just fixed it!

src/longident.ml
```diff
@@ -51,10 +51,15 @@ let parse s =
...
     invalid_arg (Printf.sprintf "Ppxlib.Longident.parse: %S" s)
   in
-  match (String.index_opt s '(', String.rindex_opt s ')') with
-  | None, None -> parse_simple s
-  | None, _ | _, None -> invalid ()
-  | Some l, Some r -> (
+  if String.length s < 1 then invalid ();
+  let open_par = String.index_opt s '(' in
+  let close_par = String.index_opt s ')' in
+  let first_char_is_normal = is_normal_ident_char s.[0] in
```
panglesd, line 57: Also, `is_normal_ident_char` allows the `'` character, but I don't think a longident can start with `'`...

NathanReb: Fixed it!

PR #124, opened 2020-04-08
Add Deriving.of_name for capturing external derivers in aliases

mrmr1993:
This PR adds a mechanism to refer to existing derivers defined outside of the current deriver library, so that they may be used in aliases.

For example, in the [Coda project](https://github.com/CodaProtocol/coda) we would like to wrap `bin_io`, but this is currently only possible by abusing `Obj.magic` and similar. It would be preferable to have a safe interface.

This can be used as
```ocaml
let local_deriver1 = Deriving.add "local_deriver1" ...

let local_deriver2 = Deriving.add "local_deriver2" ...

let local_deriver_all =
  Deriving.add_alias "local_deriver_all"
    [local_deriver1; local_deriver2; Deriving.of_name "external_deriver"]
```


diml:
How are you using `Obj.magic`?

Regarding the patch, I seem to remember that somewhere in the code we make an assumption about the fact that all existing `Deriver.t` where created by `Deriver.add*`. So the code would be checked carefully.

tmattio:
It seems to me that it should be using `Ppx_derivers.lookup` instead of `Fn.id` to make sure the deriver actually exists.

Seeing the PR is three years old I'm going to close this for now, but feel free to re-open if needed.

PR #130, opened 2020-06-09
Bump internal AST to 4.10

### LINKED ISSUE #129: Bump the Selected_ast to 410

NathanReb:
Closes #129

Besides the straightforward changes, there are a couple things worth noting:
- Location checks needed a bit of adjustement for the functors relocation functions. I talked with @trefis to make sure I wasn't messing things up so that should be good.
- The `map_with_path` and `map_with_ctxt` classes now need to deal with module bindings and module declarations without ids, i.e. things like `module _ = ...`. For now I choosed to add the underscore to the path in both cases so that the paths would differ between the inside and the outside of an unnamed module, ie in:
```
let x = ...

module _ = struct
  let y = ...
end
```
x and y would have different paths. That means that now one could end up with e.g., `A.B._.c` as a path. That's not perfect and I'm happy to change that but I think paths need to carry that information somehow. Maybe we could try to identify them using their location in the file somehow, something like:
`A.B.<unnamed module at line x>.c`.

As a follow up I'll look into our tests and make sure they are ran with the latest versions of OCam rather than with 4.06 and 4.07 as they currently are! I ran the test locally and all seems fine! If we improve the situation with unnamed modules as mentioned above, might be worth adding tests for it though.


jeremiedimino:
@ceastlund are you happy with using `_` for paths? Inside JS, I remeber that `ppx_sexp_conv` uses that. It's for error messages IIRC.

NathanReb:
I just pushed a fix, the CI was broken for OCaml < 4.08

NathanReb:
The appveyor builds seem to fail to install cinaps for some reason.

jeremiedimino:
My guess is that it relies on a behaviour of the stdlib that was broken before 4.08. Let's drop AppVeyor testing for OCaml < 4.08.

NathanReb:
Maybe it's worth including https://github.com/ocaml-ppx/ppxlib/pull/148 into this PR then!

jeremiedimino:
Seems good

NathanReb:
Just to make sure we're all on the same page, the tests are now only run for 4.10. As I mentioned in the last commit's message, our current test suite can't be compatible with 4.09 or lower since there are differences in how ASTs get printed.

jeremiedimino:
Yes. It seems too much of a pain to keep the testsuite compatible with multiple versions of the compiler, especially with the compiler libs behaviour changing all the time.

NathanReb:
One last thing to fix and then we should be good to go: the 4.10 builds on appveyor seem to fail when running cinaps, something about a permission denied:
```
C:\cygwin64\home\appveyor\.opam\ocaml-variants.4.10.0+mingw64c\bin\dune.exe "runtest" "-p" "ppxlib" "-j" "2" (CWD=C:/cygwin64/home/appveyor/.opam/ocaml-variants.4.10.0+mingw64c/.opam-switch/build/ppxlib.0.13.0)
       cinaps alias src/cinaps (exit 2)
 (cd _build/default/src && .cinaps/cinaps.exe -diff-cmd -)
 Fatal error: exception Sys_error("C:\cygwin64\tmp\cinaps5ab341.mli: Permission denied")
       cinaps alias src/cinaps (exit 2)
 (cd _build/default/src && .cinaps/cinaps.exe -diff-cmd -)
 Fatal error: exception Sys_error("C:\cygwin64\tmp\cinaps5ab341.mli: Permission denied")
[ERROR] The compilation of ppxlib failed at "C:\\cygwin64\\home\\appveyor\\.opam\\ocaml-variants.4.10.0+mingw64c\\bin\\dune.exe runtest -p ppxlib -j 2".
```

jeremiedimino:
Can we disable cinaps on appveyor? We only need it to run once.

NathanReb:
That means disabling the tests on appveyor though, not sure there's a better approach to this unfortunately.

jeremiedimino:
It is what it is. ppxlib doesn't do much that is system dependent so it's not the end of the world. And it's only until we can figure out what's going on with cinaps on Windows.

PR #147, opened 2020-06-15
[WIP] announce of omp 2.0.0

jeremiedimino:
Not to merge. This is just preparing the announce of omp 2.0.0.

### DIFF
--- a/ann.md
+++ b/ann.md
@@ -126,11 +126,13 @@ ppx rewriters. So we don't know if using Dune to upgrade all ppx
 rewriters at once will be a good enough story to make writing and
 maintaining ppx rewriters painless.
 
-However, because it is technically simpler than our original plan, we
-want to give this idea a proper try. @NathanReb is going to try this
-process for the upcoming 4.10 upgrade of ppxlib. Depending on the
-result, we might try this process once more for the 4.11 upgrade or
-decide sooner to pursue our original plan.
+However, because it leads to a world that is technically simpler and
+easier to reason about, we want to give this idea a proper
+try. @NathanReb is going to try this process for the upcoming 4.10
+upgrade of ppxlib. Depending on how this goes, we will decide whether
+to continue with our original plan or diverge and follow a plan based
+on the ocaml-migrate-parsetree technology rather than the dynamic AST
+we have been working on.
 
 [omp]: https://github.com/ocaml-ppx/ocaml-migrate-parsetree/
 [ppxlib]: https://github.com/ocaml-ppx/ppxlib/
--- a/ann.md
+++ b/ann.md
@@ -25,8 +25,30 @@ to maintain.
 ### What is the replacement?
 
 At the moment, ocaml-migrate-parsetree offers one way of writing ppx
-rewriters. Following the simplification, ocaml-migrate-parsetree
-itself will no longer be enough.
+rewriters. For instance, one might write the following to create a
+ppx rewriter with ocaml-migrate-parsetree:
+
+```ocaml
+open Migrate_parsetree
+open Ast_404
+
+let structure = ...
+let signature = ...
+
+let () =
+  Driver.register ~name:"ppx_cstruct" Versions.ocaml_404
+    (fun _config _cookies -> {default_mapper with structure; signature})
+```
+
+Following the simplification, the above code will break for at least
+two reasons:
+
+- the `Migrate_parsetree.Driver` will be gone
+- the `default_mapper` and `mapper` type coming from
+  `Ast_404.Ast_mapper` will be gone
+
+As a result, ocaml-migrate-parsetree itself will no longer be enough
+to write ppx rewriters.
 
 We recommend that ppx rewriters using the ocaml-migrate-parsetree API
 switch to [ppxlib][ppxlib].  The purpose of ppxlib is to provide a
--- a/ann.md
+++ b/ann.md
@@ -63,7 +63,8 @@ with the release of ocaml-migrate-parsetree 2.0.0 and a compatible
 ppxlib release.  At this point, most exsiting reverse dependencies of
 ocaml-migrate-parsetree will be marked as incompatible with
 ocaml-migrate-parsetree 2.0.0 and we will continue porting existing
-projects to ppxlib.
+projects to ppxlib to gradually make the world compatible with
+ocaml-migrate-parsetree 2.0.0.
 
 ### Consequences for other ppx libraries
 
--- a/ann.md
+++ b/ann.md
@@ -10,7 +10,7 @@ will go away. This will have an impact for projects that are currently
 using these functionalities and this post describes what we think such
 projects should do and how we plan to help.
 
-The motiviation for this change is to simplify ocaml-migrate-aprsetree
+The motivation for this change is to simplify ocaml-migrate-parsetree
 and make it easier to maintain. The core functionality it provides is
 valuable and worth maintaining, however right now the project is way
 too big and difficult to maintain. Going through compiler upgrades has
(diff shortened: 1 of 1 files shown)


ann.md
```diff
@@ -0,0 +1,86 @@
...
+We are preparing a 2.0.0 release of [ocaml-migrate-parsetree][omp].
+This new major release will simplify the project to its core purpose,
+which is to provide migration functions between the various versions
+of the parsetree types. Everything else, such as the driver or the
+per-version copy of various modules from the compiler distribution
+will go away. This will have an impact for projects that are currently
+using these functionalities and this post describes what we think such
+projects should do and how we plan to help.
+
+The motiviation for this change is to simplify ocaml-migrate-aprsetree
```
CraigFe, line 13:
```suggestion
The motivation for this change is to simplify ocaml-migrate-parsetree
```

ann.md
```diff
@@ -0,0 +1,86 @@
...
+and make it easier to maintain. The core functionality it provides is
+valuable and worth maintaining, however right now the project is way
+too big and difficult to maintain. Going through compiler upgrades has
+been unpleasant and difficult. And because of this, bugs of the
+segfault kind have accidentally been introduced and detected much
+later.
+
+Once we have droppped everything that doesn't fit in the purpose of
```
CraigFe, line 21:
```suggestion
Once we have dropped everything that doesn't fit in the purpose of
```

ann.md
```diff
@@ -0,0 +1,86 @@
...
+corresponding ticket from the list that you are looking at porting
+this project so that we know someone is working on it. If you would
+like to port a project that is not on this list, you should feel free
+to open a new ticket and mention that you are porting this project. In
+this case, please make sure to add the label `port-to-ppxlib` to the
+issue.
+
+Once we have successfully converted a few projects, we will proceed
+with the release of ocaml-migrate-parsetree 2.0.0 and a compatible
+ppxlib release.  At this point, most exsiting reverse dependencies of
```
CraigFe, line 63:
```suggestion
ppxlib release.  At this point, most existing reverse dependencies of
```

ann.md
```diff
@@ -0,0 +1,86 @@
...
+The following ticket is the meta issue tracking the preparation of the
+2.0.0 release:
+
+https://github.com/ocaml-ppx/ocaml-migrate-parsetree/issues/101
+
+To make this transition smoother, we are planning to help with the
+porting effort. We have prepared [a list of ppx rewriters to port to
+ppxlib][toport].  This list is not exhaustive and in particular
+excludes ppx rewriters that have no reverse dependencies in opam in
+order to keep this work maneageable for us.  In any case, help porting
```
CraigFe, line 50:
```suggestion
order to keep this work manageable for us.  In any case, help porting
```

gasche:
For the record (reacting to the part of the announce, which I think has now been edited away, that discusses versioning schemes), I prefer the current o-m-p versioning story to the one proposed by default with ppxlib, which is to use `Ppxlib.Current` as the AST, to have my code break in a few ppxlib releases. With o-m-p (and I think it is possible to use ppxlib this way as well), I can specify a specific AST version in my ppx code. This has two major advantages:
- If I want to, I can upgrade to new OCaml releases faster than ppxlib's "current" decides to move (ppxlib current is still stuck to 4.08, while 4.10 has been released a few months ago; because upgrades may break all users, they are not done very frequently).
- I remain compatible with all future versions of o-m-p that still support this AST version (which is not that costly maintenance-wise). In particular, I get bugfixes without having to change my code. With Ppxlib's approach, I can only use versions of Ppxlib that keep the same Current version, and I miss the bugfixes of all following releases.

I don't see a clear advantage to the Ppxlib versioning scheme, and I don't know where this choice has been documented and justified.

(Now that I think about it, one is that it forces most ppx authors to stick to the same Parsetree version, and thus it reduces the need to convert between parsetrees at preprocessing time, which is good for performance.)

jeremiedimino:
@CraigFe thanks for the typo fixes!

@gashe I removed it because it felt like I was talking about two different subjects and it was making the post more difficult to digest. I'm thinking of writing a dedicated blog post for the versioning story and our duniverse experiment. But the fact is, I did talk several times about the advantages of ppxlib in various blog posts or emails, and we even talked about it at ICFP last year. It seems that such arguments always end up lost at sea. If we really wanted to describe ppxlib and why it is valuable in a lasting way, it seems to me that one of us would need to sit down and write a paper about it. However, writing papers is not in our job description so here we are.

That said, it's much easier and pleasant to have such conversations over live videoconf, so I'm happy to schedule a time to do that.

gasche:
Apologies for being a bit unclear; I didn't mean to criticize the edit (I found the previous version interesting, but I think the new version is fine as well), and I'm not trying to criticize the work either. I'm just saying that, as someone that is only working on ppx on a very irregular basis, it is hard to find documentation on these design choices -- and I think those are natural questions, at least they are the first questions that came to my mind when I was explained ppxlib usage after the fact. I wish they would be documented in the ppxlib documentation, for example, which would be a natural place to have them. (I perfectly understand that you, @jeremiedimino, is fairly time-constrained, same boat here, but maybe some other contributors could try to do this? @NathanReb: this is a subtle hint.)

(I can't find a discussion of this specific aspect of the ppxlib design in the [abstract](https://icfp19.sigplan.org/details/ocaml-2019-papers/8/The-future-of-OCaml-PPX-towards-a-unified-and-more-robust-ecosystem) or [slides](https://github.com/NathanReb/future-of-ppx-talk) of the OCaml'19 workshop talk, which I enjoyed very much seeing at the time.)

I'm happy to schedule a video call with you, but I'm also worried that I will have forgotten about some of the details next time I need to work on ppx_deriving in 6 months. Ideally there would be live discussions *and* written documentation in the software repository.

jeremiedimino:
Yh, it's true that from the ppxlib's landing page or the manual it's not easy to figure all this out. The manual is really in its infancy. We were planning to improve it but then all our focus went to the ppx refactoring project. BTW, we adopted a transparent way of working, with [meeting notes](https://github.com/ocaml-ppx/ppx/wiki) published on the web so that everyone can follow what's happening and what's on our stack.

 FTR, [this blog post](https://blog.janestreet.com/ppx_core-context-free-rewriters-for-better-semantic-and-faster-compilation/) explains the most important design choices of ppxlib. Perhaps you can have a look and then we can discuss it if you have questions. In any case, I agree on written documentation but live discussion are also helpful to understand what are the grey areas from an external point of view and guide writing the documentation.

jeremiedimino:
Announce posted: https://discuss.ocaml.org/t/ocaml-migrate-parsetree-2-0-0/5991/2

jeremiedimino:
I was thinking again about this discussion, and there is actually a point to be made about documentation: if ppxlib is to become the standard API to write ppx rewriters, then it should be properly documented. It's as important as the library being good technically. I'll make sure to integrate writing documentation (README and manual) in our plan.

gasche:
I had a discussion with @jeremiedimino and @NathanReb about ppxlib
design and the choice of (non)versioning the API. Here is what I
understand now.

- The main reason why Ppxlib does not provide a versioned API by
  default is that the maintainers cannot afford to maintain the Ppxlib
  feature set (helpers to build and inspect the AST fragments) for
  several different AST versions. This would be more work and make the
  system more complex and harder to maintain.

- It is possible for users who want a versioned API to use
  ocaml-migrate-parsetree explicitly (converting between their
  preferred version and ppxlib's version), but then the Ppxlib helpers
  are not available for the version they are working with. This is
  similar to the situation of ppx preprocessors that do not use Ppxlib
  with the extra feature that ppxlib is aware of which part of the AST
  need to be sent to their preprocessor, and only converts those
  parts.

- The current plan for ppxlib is to update Ppxlib's AST to the most
  recent OCaml release on a frequent basis (shortly after each OCaml
  release). The ppxlib maintainers built tools to easily check and
  patch a large number of ppxlib-using ppx preprocessors, and submit
  PRs upstream as they update their AST version.

- We can always wait a few years and evaluate this choice with some
  experience of a few OCaml releases with ppxlib update. (In
  particular, if someone wanted to move to a versioned AST, it would
  be just as easy to do it in a couple years.)

PR #149, opened 2020-06-26
Don't leak OMP's config through ppxlib's API

NathanReb:
This removes `Migrate_parestree.Driver.config` from ppxlib's interface. We still expose the `tool_name` as it's the only part of it that was used as far as I can tell. This change is required ahead of the future omp.2.0.0 release.

### DIFF
--- a/src/driver.ml
+++ b/src/driver.ml
@@ -134,7 +134,7 @@ module Transform = struct
       Some { first with loc_end = last.loc_end }
   ;;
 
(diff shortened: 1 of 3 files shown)


NathanReb:
This is extremely weird but I pushed an extra commit to my branch with the removal of the `ocamldoc` mentions and for some reasons it doesn't show up here.

jeremiedimino:
Github is having issues today.  The CI results are going to be the same after adjusting the comment, so you can commit&push to master directly.

PR #150, opened 2020-07-01
Update expect tests

jeremiedimino:
In a similar way to what was done in the ppx project.

/cc @ceastlund


jeremiedimino:
Everything seems to be working now.

PR #151, opened 2020-07-06
Remove base dependency

ceastlund:
Removes ppxlib's dependency on base and stdio. Reimplements the parts ppxlib still needs in a new helper library Stdppx.

### DIFF
--- a/ppxlib.opam
+++ b/ppxlib.opam
@@ -21,6 +21,9 @@ depends: [
   "ppx_derivers"            {>= "1.0"}
   "sexplib0"
   "ocamlfind"               {with-test}
+  "cinaps"                  {with-test}
+  "base"                    {with-test}
+  "stdio"                   {with-test}
 ]
 synopsis: "Base library and tools for ppx rewriters"
 description: """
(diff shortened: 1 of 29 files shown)


ppxlib.opam
```diff
@@ -15,14 +15,16 @@ run-test: [
...
-  "base"                    {>= "v0.11.0"}
   "dune"                    {>= "1.11"}
   "ocaml-compiler-libs"     {>= "v0.11.0"}
   "ocaml-migrate-parsetree" {>= "1.5.0"}
   "ppx_derivers"            {>= "1.0"}
-  "stdio"                   {>= "v0.11.0"}
+  "sexplib0"
+  "stdlib-shims"
   "ocamlfind"               {with-test}
-  "cinaps"                  {with-test & >= "v0.12.1"}
```
NathanReb, line 25: Do you know why the version constraint is not necessary anymore?

ceastlund: Not sure. Not even sure it's not required. May have just gotten lost in the shuffle. I'll look into it.

jeremiedimino: In doubt, we might as well keep it. Less chances for things to go wrong.

PR #157, opened 2020-07-20
Add a Getting Started section to the manual

NathanReb:
This is a first step towards improving the manual. It doesn't cover everything but it will help people get started by providing basic explanations and a skeleton for extension rewriters and derivers.

It's hard to quickly explain how to write the actual logic so it doesn't cover that either but I was thinking that adding a couple of examples and pointing to them instead would be better. I was thinking about a small ppx_getenv for extensions but I'm a bit shot on ideas for a deriver so I'd welcome anything that's remotely meaningful and that can be fully written in a couple dozen lines.

I was thinking that a good thing to cover next is the AST. I don't think we should write a huge section about this but at least the basics + a few pointers. The ones I had in my original blog post are a bit outdated, except `Parsetree.mli`, it might make sense to bring some tools back to life in ppxlib such as `dumpast`.


NathanReb:
I think this is ready for review/merge. Adding dumpast will require a bit more work than I expected!

jeremiedimino:
@gasche would you be happy to do a pass of review on this?

gasche:
In an absolute sense, yes, but in practice I'm leaving for holidays in two days and I am not confident I will have time. (I looked at an earlier version of the PR and I found it nice, but I didn't do a precise review at the time.) cc @Octachron and @thierry-martinez who may also be interested.

jeremiedimino:
Ack, have a nice holiday then!

@Octachron and/or @thierry-martinez, if you could have a look that would be nice :)

Octachron:
I am on holidays until the 10 of August. I would be happy to do a review when I am back to work.

NathanReb:
Ok let's merge this for now and I'll address feedback later, please feel free to submit your reviews here when you get back from holidays!

PR #158, opened 2020-07-23
strip janestreet-specific CR comments

ceastlund:
CR comments are a janestreet-ism. There's no particular reason we should export them to upstream code.


trefis:
I don't think simply removing the comment is the right approach here.
Why not make them "TODO"s?

ceastlund:
Okay, done.

PR #161, opened 2020-08-03
`register_transformation`: add opt `~instrument`

pitag-ha:
Before, the way to register an instrumentation PPX was to pass the implementation file transformation to the `~impl`-argument of `Driver.register_transformation`. That way, the position at which it was applied was determined as follows: a deriver registered through `~impl` is a whole file transformation. All whole file transformations get applied after `~rules` rewriters get applied. Among whole file transformations, the order is alphabetic with respect to the name they are registered with. So the only way to influence when an instrumentation PPX got applied was through its name.

This commit adds an optional `~instrument`-argument to `Driver.register_transformation`, that allows to specify if one wants the instrumentation PPX to be applied before or after all other whole file transfomrations and `rules` rewriters get applied.

Signed-off-by: Sonja Heinze <sonjaleaheinze@gmail.com>

### DIFF
--- a/src/driver.ml
+++ b/src/driver.ml
@@ -60,6 +60,14 @@ module Cookies = struct
     List.iter !post_handlers ~f:(fun f -> f t)
 end
 
+module Instrument = struct
+  type pos = Before | After
+
+  type t = { transformation :  Parsetree.structure -> Parsetree.structure ; position : pos}
+
+  let make transformation ~position = { transformation ; position }
+end
+
 module Transform = struct
   type t =
     { name            : string
@@ -72,6 +80,7 @@ module Transform = struct
     ; preprocess_intf : (Parsetree.signature -> Parsetree.signature) option
     ; enclose_impl    : (Location.t option -> Parsetree.structure * Parsetree.structure) option
     ; enclose_intf    : (Location.t option -> Parsetree.signature * Parsetree.signature) option
+    ; instrument      : Instrument.t option
     ; rules           : Context_free.Rule.t list
     ; registered_at   : Caller_id.t
     }
@@ -89,7 +98,7 @@ module Transform = struct
 
   let register ?(extensions=[]) ?(rules=[]) ?enclose_impl ?enclose_intf
         ?impl ?intf ?lint_impl ?lint_intf ?preprocess_impl ?preprocess_intf
-        ?(aliases=[]) name =
+        ?instrument ?(aliases=[]) name =
     let rules =
       List.map extensions ~f:Context_free.Rule.extension @ rules
     in
@@ -113,6 +122,7 @@ module Transform = struct
       ; preprocess_impl
       ; preprocess_intf
       ; lint_intf
+      ; instrument
       ; registered_at = caller_id
       }
     in
@@ -230,11 +240,27 @@ module Transform = struct
       ; preprocess_intf = None
       ; enclose_impl
       ; enclose_intf
+      ; instrument = None
       ; rules
       ; registered_at = Caller_id.get ~skip:[]
       }
 
   let partition_transformations ts =
+    let before_instrs, after_instrs, rest = 
+      List.fold_left ts ~init:([], [], []) ~f:(fun (bef_i, aft_i, rest) t -> 
+        let reduced_t = 
+          { t with
+            lint_impl = None
+          ; lint_intf = None
+          ; preprocess_impl = None
+          ; preprocess_intf = None
+          } in
+        let f instr = (instr.Instrument.position, instr.Instrument.transformation) in
+        match Option.map t.instrument ~f with
+        | Some (Before, transf) -> { reduced_t with impl = Some transf }::bef_i, aft_i, rest
+        | Some (After, transf) -> bef_i, { reduced_t with impl = Some transf }::aft_i, rest
+        | None -> bef_i, aft_i, reduced_t::rest)
+    in
     (`Linters
        (List.filter_map ts ~f:(fun t ->
           if Option.is_some t.lint_impl || Option.is_some t.lint_intf then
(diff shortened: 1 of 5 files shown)


src/driver.ml
```diff
@@ -230,11 +232,27 @@ module Transform = struct
...
       ; enclose_intf
+      ; instrument = None
       ; rules
       ; registered_at = Caller_id.get ~skip:[]
       }
 
   let partition_transformations ts =
+    let before_instrs, after_instrs, rest = 
+      List.fold_left ts ~init:([], [], []) ~f:(fun (bef_i, aft_i, rest) t -> 
+        let red_t =
```
jeremiedimino, line 243: What does `red_t` stands for?

pitag-ha: I meant `reduced_t` with it. I've written it out now.

src/instrument.mli
```diff
@@ -0,0 +1,14 @@
...
+type pos = Before | After
+
+type t = { transformation :  Parsetree.structure -> Parsetree.structure ; position : pos}
+
+(** Define an instrumentation that can be passed to [Driver.register_transformation] to 
+    instrument an implementation file. [transformation] is the transformation that will
+    be applied to the AST; [position] specifies if it should be applied before or after
+    rewriters defined through [rules], [impl] or [intf] are applied.*)
+val make 
+  : transformation:(Parsetree.structure -> Parsetree.structure)
```
jeremiedimino, line 12: `transformation` feels a tad long here, and when combined with `register_transformation` it feels a bit redundant:

```ocaml
Driver.register_transfirmation ~instrument:(Instrument.make ~transformation ...)
```

Skipping the label entirely seems fine here.

test/instrument/test.ml
```diff
@@ -0,0 +1,43 @@
...
+
+let () =
+  let transform = Float (Float.abs) |> apply_operation in
+  Driver.register_transformation ~impl:transform#structure "d"
+
+let x, y = -1, -1.0
+[%%expect{|
+val x : int = 0
+val y : float = 2.
+|}]
```
jeremiedimino, line 43: This test seems good, however for readers to convince themselves that the test is passing they have to carefully look at the various transformations and apply them in their head. It would be better if things where completely obvious. One way to do it would be to extend a list of strings with the transformation name. For instance, the expected result would be:

```ocaml
["instrumentation pos=Before"; "impl"; "instrumentation pos=After"]
```

Then a reader can immediately see whether the test is passing or not.

src/driver.ml
jeremiedimino: Ack, yh `reduced` is definitely clearer !

src/driver.ml
```diff
@@ -322,21 +337,21 @@ let get_whole_ast_passes ~hook ~expect_mismatch_handler ~tool_name =
         List.find !Transform.all ~f:(fun (ct : Transform.t) ->
           Transform.has_name ct name))
   in
-  let (`Linters linters, `Preprocess preprocess, `Rest cts) =
+  let (`Linters linters, `Preprocess preprocess, `Before_instrs b_instrs, `After_instrs a_instrs, `Rest cts) =
```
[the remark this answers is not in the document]
pitag-ha: Yes, that's true. I've done changed that.

src/instrument.mli
pitag-ha: Thanks. I've also changed that.

src/instrument.mli
```diff
@@ -0,0 +1,14 @@
+open Import
+
+type pos = Before | After
+
+type t = { transformation :  Parsetree.structure -> Parsetree.structure ; position : pos}
```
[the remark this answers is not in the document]
pitag-ha: The type is abstract in `ppxlib.ml`. It's not abstract here since I need to access the fields in `driver.ml`. 

As an alternative, I could not have a `instrumentor.mli`-file at all and only define the signature of it in `ppxlib.ml`. Or I could also make `t` abstract in `instrumentor.mli` and define getters there that I wouldn't expose in `ppxlib.ml`. Does any of those sound better to you?

test/instrument/test.ml
pitag-ha: That's a very nice idea, thanks! I've added the "a"-, "b"- and "c"-prefixes to the names to stress that the alphabetic order doesn't influence. But if you find that distracting, I can get rid of them.

src/instrument.mli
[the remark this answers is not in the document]
pitag-ha: I could also include this very small module into the driver module and have the type abstract from the beginning on and include it into `ppxlib.ml` as `module Instrumentor = Driver.Instrumentor`.

test/instrument/test.ml
jeremiedimino: Looks good to me :)

jeremiedimino: BTW, the point about using the letter prefixes to stress that the alphabetic order doesn't influence the result is the kind of thing that is worth explaining in a comment. The intent is not necessarily obvious when reading the code, and if it's not obvious to a future maintainer they might simplify it away.

src/instrument.mli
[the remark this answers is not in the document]
jeremiedimino: When we can, it is better to keep the internal and external interfaces the same. It makes it easier to understand the code as we don't need to remember that `ppxlib.ml` changes the API. It also plays better with tools as well, such as documentation generators or merlin. Here, if someone just to the `Instrumentor.t` type via merlin they'll reach the version in ppxlib.ml which is not documented.

In this case, either adding getters or as a sub-module of `driver.mli` seems fine. Given that it is a small module that is likely to only be used at the registration stage, I would personally put it in `Driver` and not even add the `module Instrumentor = Driver.Instrumentor` in `ppxlib.ml`. This is in line with what is done for `Cookies` for instance.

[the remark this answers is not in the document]
pitag-ha: Sounds convincing to me. I've made a sub-module of `Driver` out of it.

test/instrument/test.ml
pitag-ha: I've made a comment explaining that.

PR #164, opened 2020-08-06
moved old OMP versioning code into ppxlib

ceastlund:
This can be better integrated into ppxlib rather than "vendored" as a helper library, but for now it's a starting point that works.


jeremiedimino:
Since we are about to delete all this code right after, it seems better to me to go directly to the better integrated version. WDYT?

ceastlund:
Works for me, I just wasn't sure if you were in a hurry to get a working version out the door.

jeremiedimino:
Ack. I'd like that we release omp 2 and the new ppxlib quite soon to keep the momentum and so that we can move on to the next phase. However, there is no hurry so waiting until Thursday for the final version (final for the current milestone) of ppxlib seems fine to me.

ceastlund:
Okay, the OMP code is now integrated into ppxlib_ast.

jeremiedimino:
Alright, I trimmed a bit more the code and setup cinaps for the new modules.

AppVeyor is failing as omp 2 doesn't seem to be availble in the Windows repo yet.

PR #173, opened 2020-09-06
Add accessors for code_path and tool_name to Expansion_context.Base

jberdine:
For code implemented using
e.g. `Ast_traverse.map_with_expansion_context`, there is a base
expansion context available, but no accessors for e.g. the `code_path`
within. A hacky workaround is to write e.g.
```ocaml
let code_path base =
  Expansion_context.Extension.code_path
    (Expansion_context.Extension.make
       ~extension_point_loc:Location.none ~base ())
```
This patch adds accessors to directly obtain these values without such
a detour.

### DIFF
--- a/src/expansion_context.ml
+++ b/src/expansion_context.ml
@@ -8,6 +8,9 @@ module Base = struct
     let code_path = Code_path.top_level ~file_path in
     {tool_name; code_path}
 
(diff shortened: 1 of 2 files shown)


NathanReb (approved):
This looks good to me, thanks for adding those!

PR #181, opened 2020-09-15
Fix ppxlib_ast preprocessing

NathanReb:
The `deriving_inline` code generation was broken by the preprocessing action and this PR fixes it + remove some code that belonged in OMP but doesn't in ppxlib anymore.

Previously running `dune build @lint` would result in the following error:
```
Error: ppxlib_driver: the rewriting contains parts from another file.
It is too complicated to reconcile it with the source
```


jeremiedimino (approved):
Looks good

NathanReb:
Ouch travis is having a hard time right now it seems :/

NathanReb:
It seems to consistently fail at fetching the sources of our dependencies :/

PR #182, opened 2020-09-15
Make ppxlib compatible with 4.11 and add 4.11 builds to the CI

NathanReb:
Note that this PR does not bump the internal AST to 4.11.

For now I kept `Longident.parse` and discarded the warning but I'm happy to add a compat function that would
be equal to `Parse.longident (Lexing.from_string x)` for 4.11+ and default to `Longident.parse` otherwise but I'm just not too sure where to put it.

It might also be a good idea to use it to replace our `Longident.parse` buggy implementation in `src/` but I'm guessing that a proper fix for all versions would be better suited there. Actually it might be better suited anywhere!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,3 +21,4 @@ env:
   - OCAML_VERSION="4.09"
     TEST=false
   - OCAML_VERSION="4.10"
+  - OCAML_VERSION="4.11.1"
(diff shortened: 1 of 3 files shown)


jeremiedimino (approved):
Looks good

PR #183, opened 2020-09-16
Add `cases` to Ast_traverse

pitag-ha:
In Ast_mapper there is a `cases` field which didn't have a corresponent method in Ppxlib's Ast_traverse classes. This commit adds that support to Ast_traverse. With that, all fields from Ast_mapper will be supported in Ast_traverse.

It's convenient to have that `cases` support since some PPX libraries filter out some cases from the list or modify all of them in the same way.

### DIFF
--- a/ast/ast.ml
+++ b/ast/ast.ml
@@ -30,6 +30,7 @@ open Import
    - adding a type longident_loc = longident loc and replacing all the occurences of the
    latter by the former. This is so that we can override iteration an the level of a
    longident loc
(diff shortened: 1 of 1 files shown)


NathanReb (commented):
This won't be enough, you also need to replace all occurences of `case list` by `cases` in the other type definitions.

As you can see this method is never called by other methods here!

NathanReb (approved):
Looks good thanks!

PR #184, opened 2020-09-17
Remove unused dependency on result

kit-ty-kate:
The `result` library does not seem to be used anywhere.


kit-ty-kate:
This seems to have been added in ppxlib.0.16.0 by https://github.com/ocaml-ppx/ppxlib/pull/164

NathanReb (approved):
Looks good to me, thanks!

PR #186, opened 2020-09-22
Add a check to make sure changelog is updated when necessary

NathanReb:
Keeping the changelog up to date on master should ease the release process.


jeremiedimino (approved):
Looks good. Seems like this could be useful for Dune as well

PR #187, opened 2020-09-22
Astlib

pitag-ha:
This commit adds the new module Astlib to ppxlib. Astlib on one hand includes the various AST versions and conversions between them that used to be in the external library OMP. On the other hand, Astlib provides a minimal API of everything used inside ppxlib from the compiler; ppxlib only uses compiler-libs through Astlib now. To make sure of that, this commit also disables the use of implicit transitive dependencies.
    
The goal is to upstream Astlib into the compiler. Astlib in the compiler will
1. already contain the AST of the new compiler version and the conversions between it and the older versions as soon as a new compiler is released.
2. maintain the Astlib API stable - and with that, everything used in ppxlib from the compiler.


ast/dune
```diff
@@ -6,12 +6,9 @@
  (name        ppxlib_ast)
  (public_name ppxlib.ast)
  (libraries
-  ocaml-compiler-libs.shadow
-  ocaml-compiler-libs.common
-  compiler-libs.common
-  ocaml-migrate-parsetree
+  ppxlib_astlib
   stdlib-shims)
- (flags (:standard -open Ocaml_shadow -safe-string) -w -9-27-32)
```
jeremiedimino, line 14: This open was for the developers of ppxlib, so we should keep it

stdppx/dune
```diff
@@ -2,7 +2,6 @@
  (name stdppx)
  (public_name ppxlib.stdppx)
  (libraries
-  ocaml-compiler-libs.shadow
   sexplib0
   stdlib-shims)
- (flags (:standard -open Ocaml_shadow -safe-string)))
```
jeremiedimino, line 8: same here about `-open Ocaml_shadow`

jeremiedimino (changes requested):

jeremiedimino (changes requested):
Looks good appart from the two removal of `-open Ocaml_shadow`. Next steps are:

- add mli files to the various aliased modules from the compiler libs in astlib. The goal is to be able to compile the repo with `(implicit_transitive_deps false)`. These mli files should be minimal and only expose what's needed by ppxlib
- then, look at the overall resulting API and work on it until it is an API we like

ast/dune
pitag-ha: Ok, thanks! I've put it back.

stdppx/dune
pitag-ha: I've also put this one back. Now, that I have `(implicit_transitive_deps false)`, do I still need it?

jeremiedimino: Nope, you shouldn't need it.

astlib/location.mli
```diff
@@ -0,0 +1,53 @@
...
+    [error] value containing a location, a message, and optionally sub-messages
+    (each of them being located as well). *)
+
+val raise_errorf :
+  ?loc:t -> ('a, Format.formatter, unit, 'b) format4 -> 'a
+
+val report_exception : Format.formatter -> exn -> unit
+(** Reraise the exception if it is unknown. *)
+
+exception Error of Location_error.t
```
pitag-ha, line 51: Is there a better name we could come up for for this exception? `Error` is not very explicit. That way, I could also rename the `Location.Location_error` module simply to `Location.Error`.

jeremiedimino: Not sure, this is used for various kinds of errors. Us there something preventing you from renaming `Location.Location_error` to `Location.Error` at the moment.

pitag-ha: I thought it might be confusing to have both an exception called `Error` and a module called `Error` inside `Location`. Do you think it's not?

jeremiedimino: I think it's fine.

pitag-ha:
Here's a summary of the changes this PR makes:
- the former independent library OMP is now part of the new internal library Astlib.
- the use of `compiler-libs.common` and `compiler-libs.bytecomp` is dropped everywhere except in `Astlib`. Every library/executable in this project that wants to use compiler-libs has to do that via Astlib now. Concretely, that affects the libraries `Ppxlib`, `Ppxlib_ast` and the executables `gen_ast_pattern` and `gen_ast_builder`.
- to guarantee the previous point, `(implicit_transitive_deps false)` has been added to the dune-project. So now all `dune` files contain all dependencies explicitly. Also, `stdlib-shimes` has been necessary in several dune files.
- the `-open Ocaml_shadow` flag is removed. It used to prevent the ppxlib devs from using compiler-libs unintentionally. That protection is now given by `(implicit_transitive_deps false)`. There is a [similar protection](https://github.com/ocaml-ppx/ppxlib/blob/master/src/ppxlib.ml#L5-L23) for ppxlib users which is kept.
- some functions are added to Astlib that are lighter than the original compiler-libs behaviour, e.g. `Location.set_input_name`.
- all the logic dealing with differences depending on the compiler version is moved to Astlib. So for example, `src/gen-compiler-specifics` and the preprocessing logic in `Ppxlib_ast.Location_error` are dropped. Therefore, some preprocessing logic is added to `Astlib.Location.Error` (which is completely new), `Astlib.Compmisc` and `Astlib.Longident`.
- the preprocessor rules in astlib/pp/pp_rewrite.mll and in ast/pp/pp_rewrite.mll are adapted to the previous point.
- the OMP exception `Migrate_parsetree_def.migration_error` as well as the compiler-libs exception `Syntaxerr.Error` have been turned into `Location.Error` exceptions to thin down the API.

pitag-ha:
There are two parts of the Astlib API that stand out at the moment:
- the types `Longident.t` as well as `Location.t` and `Location.loc` are concrete. Those three types have to be equivalent to the corresponding types in `Ppxlib` (via an equivalence chain over `Ppxlib_ast`). Making the types in `Ppxlib` abstract would mean breaking lots of downstream users.
- the compiler-libs `Parsetree` types are implicitly exposed by the Astlib API due to the type equivalences of the `Ast_n.Parsetree` types with them. Those equivalences are only needed by Astlib internally. But distinguishing the internal Astlib API from the external one isn't easy: an `mli`-file for it would be huge and doing dune tricks e.g. with private modules would make Astlib more complex and upstreaming it to the compiler more difficult.

pitag-ha:
Here's a summary of Astlib's content:
- the `Ast_n` and `Migrate_n_{n±1}` modules contain the Ast definitions and migration information between them that used to be in OMP
- the folders cinaps, config and pp contain code that allows preprocessing inside Astlib
- `Ast_mapper`: the four `ppx_context` functions and the two `cookie` functions are necessary for ppxlib to interact correctly with the compiler cookies. `tool_name` is necessary to give the information which tool is being used to the ppx rewriter (via the expansion context)
- `Compmisc` contains a function to read command line flags from an environment variable. I'm not sure why we support that. But it concerns end users, so I can't check in the ppx-universe if removing that feature would break users.
- `Config` contains the magic numbers
- `Location`:
  * why the types are concrete is explained [here](https://github.com/ocaml-ppx/ppxlib/pull/187#issuecomment-831932321)
  * the values `none`, `register_error_of_exn`, `raise_errorf` and `report_exception` as well as the exception are needed by ppxlib users
  * the submodule `Error` includes a compiler-version-agnostic API for everything needed for the `Ppxlib.Location.Error` module
  * `set_input_name` sets the compiler reference that keeps track of the input name (such as the file name) and is needed by the ppxlib driver when loading the input
- `Longident`: why `t` is concrete is explained [here](https://github.com/ocaml-ppx/ppxlib/pull/187#issuecomment-831932321). `flatten` is needed to pretty print a long identifier and `parse` is needed for `Ast_pattern_generated` and `Ast_builder_generated`
- `Parse` and `Parser` contain the parse functions

I've thinned this down from the 14 modules with extensive compiler content in the first version I did to these 7 relatively thin modules with compiler content.

pitag-ha:
There's still one thing we need to decide: what to do about `Pprintast` and `Ast_helper_lite`. `Pprintast` is an extended copy of the compiler `Pprintast`. We have a copy of it, because the compiler module is broken on old compilers. `Ast_helper_lite` seems to be a copy of the compiler `Ast_helper`. Is `Ast_helper` also broken on old compilers? Should I move both of them into Astlib, where I'd use our copy on old compilers and the compiler implementation on new compilers? That way, at some point in the future when dropping the old compiler support, we can drop the copies. But it would mean for the compiler to have the maintenance burden for the whole of those modules. 

If I'm not forgetting anything, that's the only decision left now :)

jeremiedimino:
Printing the AST seems like a feature Astlib could provide. What do you think?

For `Ast_helper`, agreed that it seems too much for `Astlib`. IIRC, we looked in the past and found that removing it would break a lot of users. I propose that we keep the status quo for now and keep it as it is. Since it is redundant with `Ppxlib.Ast_builder`, we could deprecate `Ppxlib.Ast_helper` so that we can eventually remove it.

astlib/astlib.ml
```diff
@@ -0,0 +1,73 @@
...
+module Migrate_407_408 = Migrate_407_408
+module Migrate_408_407 = Migrate_408_407
+module Migrate_408_409 = Migrate_408_409
+module Migrate_409_408 = Migrate_409_408
+module Migrate_409_410 = Migrate_409_410
+module Migrate_410_409 = Migrate_410_409
+module Migrate_410_411 = Migrate_410_411
+module Migrate_411_410 = Migrate_411_410
+module Migrate_411_412 = Migrate_411_412
+module Migrate_412_411 = Migrate_412_411
```
jeremiedimino, line 63: That's doesn't seem super useful to have in the API of Astlib. Seems like having `Ppxlib_ast.Versions` in `Astlib` would be more useful.

astlib/compmisc.mli
```diff
@@ -0,0 +1 @@
+val read_clflags_from_env : unit -> unit
```
jeremiedimino, line 1: It feels like this shouldn't be exposed in the API of Astlib. What about adding a function `Astlib.init` that calls this under the hood?

astlib/parser.mli
```diff
@@ -0,0 +1,24 @@
+(* The type of tokens. *)
```
jeremiedimino, line 1: This comment should give an overview of the module.

astlib/astlib.ml
jeremiedimino: Actually scratch that. These are stable, so need to grow astlib.

jeremiedimino (approved):
I did a high-level review of the changes. It looks good to me overall, nice work @sonja!

I would just remove the `Astlib.Compmisc` module, which seems a bit random. `Comp` is for "compilation", which is not what `Astlib` is about.

pitag-ha:
> Printing the AST seems like a feature Astlib could provide. What do you think?

I've just had a look and have noticed that what I said about moving `Pprintast`  into Astlib didn't make much sense: we need `Pprintast` for the AST we're using internally, not for the compiler AST.

> Since it (`Ppxlib.Ast_helper`) is redundant with `Ppxlib.Ast_builder`, we could deprecate Ppxlib.Ast_helper so that we can eventually remove it.

Yes, I agree. I've just opened in issue for that to not forget: #245

astlib/compmisc.mli
pitag-ha: I agree that `Astlib.Compmisc.read_clflags_from_env` is a bad idea. Why do you think `Astlib.init` would be better than `Astlib.read_clflags_from_env`? The latter would give a hint about what the function does. Do you prefer the former in order to keep open what it does so that it's behavior can be changed in the future?

jeremiedimino: > Do you prefer the former in order to keep open what it does so that it's behavior can be changed in the future?

That's what I was thinking. Since Astlib is a library, it's not clear why it cares about reading command line flags from the environment.

And to be fair, it would improve reproducibility if we didn't have to do that, but that's another story.

pitag-ha: Yes, needing that functionality in Astlib is really a bit unfortunate.

I've called it `Astlib.init` now as you've suggested. Do you think it would be good to add a brief documentation about it saying what it does? Or would documentation get in the way of hiding the behavior and leaving an open door to change it? The reason why I would add it is that without documentation, from an Astlib user's point of view, `init` might be a bit confusing: why does a library need to be initialized? Is it always necessary/advisable to run `Astlib.init` to use `Astlib`? The answer should be "no", but without context the user wouldn't know that.

Also, is removing `read_clflags_from_env` from the ppxlib driver something we could consider?

astlib/parser.mli
pitag-ha: Good catch! For now, I've simply removed that comment. I've had a look at the ppx-universe and we don't even need this module! I'll open a separate PR removing it. I will also write and open another separate PR adding a bit more of documention to Astlib in general.

astlib/compmisc.mli
pitag-ha: I like the idea of giving it the more explicit name `init_error_reporting_style_using_env_vars`. I've done that now.

> We could change Dune to pass a command line flag to the ppx driver instead.

Are you sure the ppxlib driver provides those command line flags? I've had a look and couldn't find them. There's a flag called `-no-color`, but that only seems to impact the printing of the diffs that are printed when there are mismatches.
Also, looking at the ppx-universe: there are over 5 downstream users that set `OCAML_COLOR` and/or `OCAML_ERROR_STYLE` in their tests (so probably there are also quite some end users who do the same). So removing this would have more impact than I expected.

jeremiedimino: > Are you sure the ppxlib driver provides those command line flags?

It doesn't. These would need to be added as well.

> Also, looking at the ppx-universe: there are over 5 downstream users that set OCAML_COLOR and/or OCAML_ERROR_STYLE in their tests (so probably there are also quite some end users who do the same). So removing this would have more impact than I expected.

Indeed.

PR #190, opened 2020-10-06
The ppxlib package is always only valid for a limited set of OCaml compilers

kit-ty-kate:
Supported versions are listed in `ast/supported_version/supported_version.ml`

If someone could also merge https://github.com/ocaml-ppx/ocaml-migrate-parsetree/pull/106 at the same time if this is accepted, that'd be really appreciated.


NathanReb (approved):
Good catch, thanks! It's a shame we have to explicitly add support for newer versions this way, hopefully we'll get it fixed with Astlib.

PR #192, opened 2020-10-22
Silence a warning 16 spotted by ocaml 4.12

NathanReb:
There must have been a bug in the previous versions of the compiler which failed to detect this warning. The function is deprecated so there isn't really any point fixing it until we delete it from ppxlib.


hhugo:
https://github.com/ocaml/ocaml/pull/9783

jeremiedimino (approved):
LGTM

PR #194, opened 2020-11-14
Expose Ppxlib.Driver.map_signature

kit-ty-kate:
This is required for the port of `ppx_deriving` to OMP2 / ppxlib >= 0.16. See https://github.com/ocaml-ppx/ppx_deriving/pull/243 for more detail.

Would it be possible to get a quick point release (`0.19.1`) once this is merged?


kit-ty-kate:
Issue in travis fixed in https://github.com/ocaml/ocaml-ci-scripts/pull/350

Would it make sense to switch to ocaml-ci anyway? cc @avsm

jeremiedimino:
> Would it make sense to switch to ocaml-ci anyway? cc @avsm

@NathanReb looked at this in the past, but they were some difficulties. In the long term, we should indeed aim to switch to ocaml-ci.

pitag-ha:
We are making it a minor release instead of a point release since it adds a new feature.

PR #197, opened 2020-12-16
Re-implement standalone_run_as_ppx_rewriter

pitag-ha:
Before, the `standalone_run_as_ppx_rewriter` implementation was based on compiler-lib's Ast_mapper. With this commit, it reuses the adequate parts of the internal implementation of the main standalone. That has three advantages:
- it takes away a burden for Astlib
- it makes it possible to access the source file name from within the driver module
- it makes it possible to implement that ppxlib preserves the compiler version, also when using standalone_run_as_ppx_rewriter or run_as_ppx_rewriter.

This commit also fixes the following: now, `standalone_run_as_ppx_rewriter` takes into account the `loc-filename` argument when provided. Since that argument forms part of `shared_args`, that's the expected behaviour.

When comparing the new implementation with the old one, note that in the new one the ppx context gets restored through `extract_cookies` and the cookies get updated through `Intf_or_impl.to_ast_io` with `~add_ppx_context:true`.

### DIFF
--- a/src/driver.ml
+++ b/src/driver.ml
@@ -655,6 +655,14 @@ let relocate_mapper = object
       pos
 end
 
+let read_ast ic =
+  match Ast_io.read ic with
+  | Ok (input_name, ast) ->
+      let ast = Intf_or_impl.of_ast_io ast in
+      let kind = Intf_or_impl.kind ast in
+      Ok (ast, input_name, kind)
+  | Error e -> Error e
+
 (* Set the input name globally. This is used by some ppx rewriters
    such as bisect_ppx. *)
 let set_input_name name =
@@ -662,13 +670,12 @@ let set_input_name name =
 
 let load_input (kind : Kind.t) fn input_name ~relocate ic =
   set_input_name input_name;
-  match Ast_io.read ic with
-  | Ok (ast_input_name, ast) ->
-    let ast = Intf_or_impl.of_ast_io ast in
-    if not (Kind.equal kind (Intf_or_impl.kind ast)) then
+  match read_ast ic with
+  | Ok (ast, ast_input_name, ast_kind) ->
+    if not (Kind.equal kind ast_kind) then
       Location.raise_errorf ~loc:(Location.in_file fn)
         "File contains a binary %s AST but an %s was expected"
-        (Kind.describe (Intf_or_impl.kind ast))
+        (Kind.describe ast_kind)
         (Kind.describe kind);
     if String.equal ast_input_name input_name || not relocate then begin
       set_input_name ast_input_name;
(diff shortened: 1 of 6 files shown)


pitag-ha:
The ocaml-ci is green apart from the old compiler version and the arm32 jobs. Those failures aren't related to this PR.

test/driver/standalone_run_as_ppx/dune
```diff
@@ -0,0 +1,13 @@
+(library
```
NathanReb, line 1: We can probaly merge the library into the executable here, what do you think? Not that important but it makes the test setup a little bit simpler, fewer files to browse, etc...

src/driver.ml
```diff
@@ -708,6 +708,38 @@ let load_input (kind : Kind.t) fn input_name ~relocate ic =
     | Impl -> input_name, Impl (Parse.implementation lexbuf)
 ;;
 
+let load_input_run_as_ppx ~input_fn ic =
```
NathanReb, line 711: Out of curiosity, why can't we use `load_input` here? Don't they at least share a common part that we could extract?

NathanReb: Ok from a further reading it seems there wouldn't be too much to share besides the call to `Ast_io.read` and the extraction of the `ast` and `kind` so it's not clear there's any real benefit here. What do you think?

NathanReb (commented):
Looks great! Thanks for the great cram tests!

test/driver/standalone_run_as_ppx/dune
pitag-ha: That's definitely better. Thanks!

src/driver.ml
pitag-ha: I tried to re-use `load_input` at the beginning, but there's really not much in common. 

I'm not really sure if I understand how we could extract the call to `Ast_io.read`, since that's already an independent function which we call in both `load_input` and `load_input_run_as_ppx`, right? So that function call is all the two `load_input` functions have in common and that's already extracted. Or were you meaning something different? `kind` is treated differently in the two functions: in `load_input` we check if `kind` matches the extension of the file given as input, which isn't possible in `load_input_run_as_ppx` since the file given as input is a temporary file. Also the way the `Ast_io.read` return result is handled is different: if `Ast_io.read` returns an `Error Not_a_binary_ast`, in `load_input` we want to treat the input as Ocaml code input, whereas in `load_input_run_as_ppx`, we want to return an error since there we expect a binary AST.

NathanReb: I think I was suggesting we extract something like:
```ocaml
let read_ast ic =
  Ast_io.read ic >>= fun (ast_input_name, ast) ->
    let ast = Intf_or_impl.of_ast_io ast in
    let kind = Intf_or_impl.kind ast in
    Ok (ast_input_name, ast, kind))
```
but again, probably not worth it!

pitag-ha: Ooh, now I see. I like the idea, thanks! I've implemented it (in `Result`, we don't have the map/bind infix operators, that's why I've pattern matched).

PR #199, opened 2020-12-17
Improve ocaml constraints for the test so ocaml-ci understands them

NathanReb:


NathanReb:
Ok this won't do, closing!

PR #201, opened 2020-12-28
Disable tests on older compiler using dune

NathanReb:
This fixes ocaml-ci for 4.07+ by disabling incompatible tests via dune. The higher the ocaml version the more tests are run.

We still need to investigate the arm32 build failure which might be related to cinaps/the cinaps stanza somehow.

For versions older than 4.07, cinaps is not available. At the moment there's no good way to handle this in ocaml-ci meaning we'll have to live without older builds for now.

### DIFF
--- a/examples/simple-deriver/ppx_deriving_accessors.ml
+++ b/examples/simple-deriver/ppx_deriving_accessors.ml
@@ -32,23 +32,25 @@ let accessor_intf ~ptype_name (ld : label_declaration) =
 
 let generate_impl ~ctxt (_rec_flag, type_declarations) =
   let loc = Expansion_context.Deriver.derived_item_loc ctxt in
-  List.concat_map type_declarations
+  List.map type_declarations
     ~f:(fun (td : type_declaration) ->
       match td with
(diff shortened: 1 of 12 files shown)


dune
```diff
@@ -1,4 +0,0 @@
-(env
- (_
-  (binaries
-   (test/expect/expect_test.exe as expect-test))))
```
pitag-ha, line 4: Why is it better / necessary not to make the executable available as "expect-test"?

pitag-ha (commented):
Great that we'll be able to use ocaml-ci. Thanks!

dune
NathanReb: I was getting error from dunes on older compiler since the stanza responsible for building `expect_test.exe` was disabled.

It might be the case that the error was only triggered because some rules were requiring `expect-test`, I'll give it a spin just to make sure.

NathanReb: Well it turns out it wasn't required and was indeed only triggered if an active rule depended on `expect-test` so I reverted this part!

Good catch!

NathanReb:
@pitag-ha I fixed the env thing, does this look good to you?

pitag-ha:
It does, thanks!

Issue #169, opened 2020-09-05
Expansion context for more rules?


jberdine:
I am considering a case where `Context_free.Rule.special_function` could be used, if its expand function were given an expansion context similar to `Expansion_context.Extension.t`. Would that, or some other means to obtain the relevant `Code_path.t`, make sense and be reasonably easily implementable?

NathanReb:
Yeah I think we should widen the use of `Expansion_context` to all context free rules indeed. If you'd like to give this a try I'm happy to point you in the right direction, otherwise I'll take a look myself when I get a bit more time!

PR #202, opened 2020-12-29
Make context available in transformation callbacks

pitag-ha:
The later versions (V2 or V3) of the Extension and Deriving modules allow to access an expansion context when writing such rewriters. By adding a V2 to Driver, this commit also adds that possibility to whole file transformations registered through the `~impl`, `intf`, `~preprocess_impl`, `~preprocess_intf`, `lint_impl` or `lint-intf` arguments of `Driver.register_transformation`. The expansion context chosen is the Base expansion context in order to be compatible with `Ast_traverse.map_with_expansion_context`.

There are other kinds of transformations that still don't offer that possibility such as other context free rules than extensions (see issue #169).

The PR also adds input name to the expansion context.

### DIFF
--- a/src/driver.ml
+++ b/src/driver.ml
@@ -65,23 +65,29 @@ end
 module Instrument = struct
   type pos = Before | After
 
-  type t = { transformation :  Parsetree.structure -> Parsetree.structure ; position : pos}
+  type t = { transformation : Expansion_context.Base.t -> Parsetree.structure -> Parsetree.structure ; position : pos}
 
-  let make transformation ~position = { transformation ; position }
+  module V2 = struct
+    let make transformation ~position = { transformation ; position }
+  end
+
+  let make transformation ~position =
+    let transformation _ st = transformation st in
+    V2.make transformation ~position
 end
 
 module Transform = struct
   type t =
     { name            : string
     ; aliases         : string list
-    ; impl            : (Parsetree.structure -> Parsetree.structure) option
-    ; intf            : (Parsetree.signature -> Parsetree.signature) option
-    ; lint_impl       : (Parsetree.structure -> Lint_error.t list) option
-    ; lint_intf       : (Parsetree.signature -> Lint_error.t list) option
-    ; preprocess_impl : (Parsetree.structure -> Parsetree.structure) option
-    ; preprocess_intf : (Parsetree.signature -> Parsetree.signature) option
-    ; enclose_impl    : (Location.t option -> Parsetree.structure * Parsetree.structure) option
-    ; enclose_intf    : (Location.t option -> Parsetree.signature * Parsetree.signature) option
+    ; impl            : (Expansion_context.Base.t -> Parsetree.structure -> Parsetree.structure) option
+    ; intf            : (Expansion_context.Base.t -> Parsetree.signature -> Parsetree.signature) option
+    ; lint_impl       : (Expansion_context.Base.t -> Parsetree.structure -> Lint_error.t list) option
+    ; lint_intf       : (Expansion_context.Base.t -> Parsetree.signature -> Lint_error.t list) option
+    ; preprocess_impl : (Expansion_context.Base.t -> Parsetree.structure -> Parsetree.structure) option
+    ; preprocess_intf : (Expansion_context.Base.t -> Parsetree.signature -> Parsetree.signature) option
+    ; enclose_impl    : (Expansion_context.Base.t -> Location.t option -> Parsetree.structure * Parsetree.structure) option
+    ; enclose_intf    : (Expansion_context.Base.t -> Location.t option -> Parsetree.signature * Parsetree.signature) option
     ; instrument      : Instrument.t option
     ; rules           : Context_free.Rule.t list
     ; registered_at   : Caller_id.t
@@ -146,7 +152,7 @@ module Transform = struct
       Some { first with loc_end = last.loc_end }
   ;;
 
-  let merge_into_generic_mappers t ~hook ~expect_mismatch_handler ~tool_name =
+  let merge_into_generic_mappers t ~hook ~expect_mismatch_handler ~tool_name ~input_name =
     let { rules; enclose_impl; enclose_intf; impl; intf; _ } = t in
     let map =
       new Context_free.map_top_down rules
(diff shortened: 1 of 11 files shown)


pitag-ha:
The ocaml-ci fails aren't related to this PR.

src/driver.ml
```diff
@@ -65,21 +65,27 @@ end
...
 
 module Transform = struct
   type t =
     { name            : string
     ; aliases         : string list
-    ; impl            : (Parsetree.structure -> Parsetree.structure) option
-    ; intf            : (Parsetree.signature -> Parsetree.signature) option
+    ; impl            : (Expansion_context.Base.t -> Parsetree.structure -> Parsetree.structure) option
+    ; intf            : (Expansion_context.Base.t -> Parsetree.signature -> Parsetree.signature) option
     ; lint_impl       : (Parsetree.structure -> Lint_error.t list) option
```
NathanReb, line 85: It's probably worth adding it here as well. What do you think?

src/driver.ml
```diff
@@ -65,21 +65,27 @@ end
...
+    V2.make transformation ~position
 end
 
 module Transform = struct
   type t =
     { name            : string
     ; aliases         : string list
-    ; impl            : (Parsetree.structure -> Parsetree.structure) option
-    ; intf            : (Parsetree.signature -> Parsetree.signature) option
+    ; impl            : (Expansion_context.Base.t -> Parsetree.structure -> Parsetree.structure) option
```
NathanReb, line 83: How about making it a labelled argument?

It's not super flexible as I think users will still have to define their function arguments in the right order but I think it's still a bit clearer this way.

NathanReb: The same apply for Instrument I guess!

NathanReb: I just checked and `Deriving` and `Extension` both use a labelled arg so it would also be more consistent!

src/expansion_context.mli
```diff
@@ -5,6 +5,17 @@ module Base : sig
   (** Return the code path for the given context *)
   val code_path : t -> Code_path.t
 
+  (** Return the input name for the given context.
+  For a context [ctxt], the input name given by [input_name ctxt] tends to coincide with
```
NathanReb, line 9: Maybe this should start by explicitly stating how `input_name` is determined, e.g. (if I'm not mistaken) the file name of the driver's argument.

It might actually be worth moving the relevant parts to the `code_path` documentation right above.

How about making it more like a list of properties for both, i.e. in short, something like:
```ocaml
(** ...
  In Driver, Deriving and Extension the context is initialized so that the [file_path] component of
  the [code_path] is determined from the first location found in the input AST. That means that:
  - It's the empty string in empty structure or signature
  - It can be altered by line directives
  etc...
  *)
val code_path : ...

(** ...
  In Driver, Deriving and Extension, the context argument is initialized so that the [input_name]
  matches the input filename passed to the driver on the command line.
  - It is ["_none_"] when rewriting ...
  - It is not affected by line directives
  - It may have a value even for empty files
  etc...
  *)
val input_name : ...
```

NathanReb (commented):
It looks great! Thanks for working on this, I think it's a solid improvement to ppxlib's API.

I have a few remark and suggestions.

Also something I found is that the cram test is a bit overly complex. I wonder if we couldn't simply create a single standalone driver with a single transformation that simply inserts a floating attribute with the values of `tool_name`, `input_name` and the `file_path` component of the `code_path`or even just dump those values on the standard output. It should obviously register `impl` and `intf` transformations. We could then invoke it on the different files you generate in your cram test! I think that would simplify the setup.

What do you think?

NathanReb:
Can you rebase on top of master so we get rid of most of the CI failures?

src/expansion_context.ml
```diff
@@ -2,13 +2,17 @@ module Base = struct
   type t =
     { tool_name : string
     ; code_path : Code_path.t
+    ; input_name : string
```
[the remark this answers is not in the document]
pitag-ha: I definitely agree. Thanks!

src/driver.ml
pitag-ha: Good catch!

src/expansion_context.mli
pitag-ha: That's a very nice idea, thanks! I've structured it that way now.

test/driver/v2/run.t
```diff
@@ -0,0 +1,45 @@
...
+  file_path: 
+
+2. files with directives pointing to other files
+
+  $ cat > directive.ml << EOF
+  > # 1 "file.ml"
+  > let y = 0
+  > EOF
+  $ add_print_fn directive.ml | ocaml -stdin
+  input_name: directive.ml
```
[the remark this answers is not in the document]
pitag-ha: What would you do about the "The three context fields can be accessed in a rewriter" and "V2 also works in the context of interface files" (btw, very bad choice of word "context" here...) tests? Just get rid of them? The main idea of them was to test the new API. So technically it's not only about the cram test there.

src/driver.ml
pitag-ha: The problem is that `Ast_traverse.map_with_expansion_context` takes an unlabelled context argument. So there already is an inconsistency between `Deriving` and `Extension` on one hand and `Ast_traverse` on the other hand. I've chosen to be consistent with `Ast_traverse`, since that's what needs to be used along with the new `Driver.V2` API (see the tests for example). Do you think it might be worth making a V2 for `Ast_traverse.map_with_expansion_context` in order to get rid of that inconsistency and make it labelled eveywhere?

pitag-ha:
> I wonder if we couldn't simply create a single standalone driver with a single transformation that simply inserts a floating attribute with the values of tool_name, input_name and the file_path component of the code_pathor even just dump those values on the standard output.

That would indeed make it simpler. It would have a couple of downsides / difficulties:
- We can't do the "empty file" test anymore, since the file needs to have at least the floating attribute
- Having the tool name around all the time when only testing the input name and the code path might be a bit confusing
- We'd have to think about if and how to test the intf API. Is it possible to dump values on the standard output from within an interface file?
- If I've understood you correctly, you would directly create the standalone executable, no ppx rewriter library. Is that right? How would you do the "using `map_structure` (or `map_signature`)" test then?

But maybe the simplicity makes up for that. In any case, I do think that in that case I would only care about input_name and file_path though and not take the tool_name into account since it wouldn't actually add any test, would it? In fact, I think I would do that anyways. What do you think?

test/driver/v2/run.t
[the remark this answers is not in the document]
pitag-ha: I've renamed it to `expansion_context`.

src/driver.ml
pitag-ha: @ceastlund : I've talked about this with @NathanReb and we were wondering of you could also give us your opinion on this. 

To sum up: The question is if the expansion context should be a labelled or an unlabelled argument.
- in favor of unlabelled: in the class `Ast_traverse.map_with_expansion_context`, it is unlabelled. A common way to use the new API is to create an object `o`, that inherits from that class, and using `o#structure` and/or `o#signature` directly to register the transformation (as done in [this test](https://github.com/ocaml-ppx/ppxlib/pull/202/files#diff-664f09497a98806f341fa6cbe5d2b65091e087fbf4bdf980fb1533f8f17df639R24)). Therefore, we'd need to be consistent with that class.
- in favor of labelled: The callbacks `V2`/`V3` in both `Deriving` and `Extension` use labelled arguments. Furthermore, using a labelled argument seems more intuitive.

I'm leaning towards unlabelled, since I think not being able to use `map_with_expansion_context#structure`/`signature` right away might make the user think that that wasn't meant to be used in combination. But I'd also be happy to make it labelled. What do you think?

pitag-ha:
@NathanReb, in the documentation of `V2.register_transformation` , I've commented on the choice of signature for the V2 callbacks . Do you think I can merge this?

src/driver.mli
```diff
@@ -173,6 +182,39 @@ val register_correction : loc:Location.t -> repl:string -> unit
...
 
+module V2 : sig
+  (** Same as [Driver.register_transformation], but the callbacks [?impl], [?intf],
+      [?preprocess_impl], [?preprocess_intf], [?lint_impl] and [?lint_intf] have access
+      to an expansion context. Their signatures coincide with the signatures of the
+      respective methods in [Ast_traverse.map_with_expansion_context]. *)
+  val register_transformation
+  :  ?extensions       : Extension.t list (* deprecated, use ~rules instead *)
+  -> ?rules            : Context_free.Rule.t list
+  -> ?enclose_impl     : (Location.t option -> structure * structure)
```
NathanReb, line 193: Sorry to think about it just now but do you know if it would make sense to pass the expansion_context here as well?

src/expansion_context.mli
```diff
@@ -2,9 +2,23 @@ module Base : sig
   (** Type for the location independent parts of the expansion context *)
   type t
 
-  (** Return the code path for the given context *)
+  (** Return the code path for the given context
+  In Driver, Deriving and Extension, the context is initialized so that the [file_path] component of
+  the [code_path] is determined from the first location found in the input AST. That means that:
+  - It's the empty string in empty structures or signatures
+  - It can be altered by earlier running PPX's
```
NathanReb, line 9: Is that actually true? Don't we instantiate it only once at the beginning? I'm genuinely asking, I'm not sure exactly how it is initialized.

NathanReb (commented):
It mostly looks good, I just have a couple questions but then it should be good to go!

Sorry for the hassle!

src/driver.mli
pitag-ha: Yes, I think you're right. That way we pass the expansion_context to all direct callbacks here. I've done that now.

src/expansion_context.mli
pitag-ha: Good catch! That's not true. I've removed that.

pitag-ha:
No hassle at all, thanks for your help!

NathanReb (approved):
Awesome, let's merge!

PR #203, opened 2021-01-08
Test OCaml >= 4.04 on OCaml-CI

kit-ty-kate:
just a test for now

### DIFF
--- a/ppxlib.opam
+++ b/ppxlib.opam
@@ -23,7 +23,7 @@ depends: [
   "stdlib-shims"
   "ocamlfind"               {with-test}
   "re"                      {with-test & >= "1.9.0"}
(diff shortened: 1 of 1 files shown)


ppxlib.opam
```diff
@@ -23,7 +23,7 @@ depends: [
   "stdlib-shims"
   "ocamlfind"               {with-test}
   "re"                      {with-test & >= "1.9.0"}
-  "cinaps"                  {with-test & >= "v0.12.1"}
+  "cinaps"                  {with-test & >= "v0.12.1" & ocaml >= "4.07"}
```
kit-ty-kate, line 26: This leads to `false` in all cases (the `ocaml` variable is not defined)

NathanReb: ah interesting, so is there an actual way to express that?

PR #204, opened 2021-01-08
Depend on cinaps only for 4.07+

NathanReb:


ppxlib.opam
```diff
@@ -14,7 +14,13 @@ run-test: [
   ["dune" "runtest" "-p" name "-j" jobs] { ocaml:version >= "4.07" }
 ]
 depends: [
-  "ocaml"                   {>= "4.04.1" & < "4.13"}
+  (
+    ("ocaml" {>= "4.04.1" & < "4.13"}) |
```
dra27, line 18: I think it might need to be:
```suggestion
    ("ocaml" {>= "4.04.1" & < "4.07"}) |
```
or the with-test can choose to ignore the cinaps

NathanReb:
CC @kit-ty-kate @dra

ppxlib.opam
NathanReb: ouh yeah my bad!

NathanReb:
https://github.com/ocaml-ppx/cinaps/pull/4 might render this unnecessary but I'd still love to figure out how this could be expressed in opam constraints if I was to run into a similar situation again in the future!

ppxlib.opam
```diff
@@ -14,7 +14,14 @@ run-test: [
...
 depends: [
-  "ocaml"                   {>= "4.04.1" & < "4.13"}
+  "ocaml" {>= "4.04.1" & < "4.13"}
+  (
+    ("ocaml" {>= "4.04.1" & < "4.07"}) |
+    (
+      "ocaml" {>= "4.07" & < "4.13"}
+      "cinaps" {with-test & >= "v0.12.1"}
+    )
+  )
```
kit-ty-kate, line 24: Note on that: If this is merged, I'd rather keep it in the repository and not put it in the releases as the opam solver struggles quite a bit with disjunctions.

NathanReb: Hopefully we'll rather count cinaps releasing a new 4.04 compatible version. See https://github.com/ocaml-ppx/cinaps/pull/4.

Out of curiosity, how would you handle it in the release? Just dropping the test deps on cinaps? That doesn't feel right.

dra27: It didn’t seem to be too bad when I tested it locally?

NathanReb:
Cinaps v0.15.0 has been released so this isn't necessary anymore, closing it!

PR #205, opened 2021-01-11
Add bucklescript support

pitag-ha:
The two important commits are:

  - Support binary AST's of any supported version
    
    Before, the OCaml version of any binary ast input had to coincide with the version the ppxlib driver got compiled with. That was a problem for bucklescript/rescript.
    
    With this commit, all versions from the Versions module are accepted as input. The input version is preserved by ppxlib when outputting a binary AST.

- Implement `run_as_ppx_rewriter` internally
    
    Besides taking away a burden for Astlib, implementing `run_as_ppx_rewriter` internally has two advantages:
    - it allows making changes with more flexibility
    - `run_as_ppx_rewriter` gets syncronized with `standalone_run_as_ppx_rewriter`
    
    With this change, the three flags `-loc-filename`, `apply`, `dont-apply` are taken into account when given. Before, they were allowed to be passed in but not taken into account.

### DIFF
--- a/ast/versions.ml
+++ b/ast/versions.ml
@@ -487,6 +487,23 @@ end
 let ocaml_412 : OCaml_412.types ocaml_version = (module OCaml_412)
 (*$*)
 
+let all_versions : (module OCaml_version) list = [
+  (*$foreach_version (fun n _ ->
+      printf "(module OCaml_%d : OCaml_version);\n" n)*)
+(module OCaml_402 : OCaml_version);
+(module OCaml_403 : OCaml_version);
+(module OCaml_404 : OCaml_version);
+(module OCaml_405 : OCaml_version);
+(module OCaml_406 : OCaml_version);
+(module OCaml_407 : OCaml_version);
+(module OCaml_408 : OCaml_version);
+(module OCaml_409 : OCaml_version);
+(module OCaml_410 : OCaml_version);
+(module OCaml_411 : OCaml_version);
+(module OCaml_412 : OCaml_version);
+(*$*)
+]
+
 (*$foreach_version_pair (fun a b ->
     printf "include Register_migration(OCaml_%d)(OCaml_%d)\n" a b;
     printf "    (Migrate_parsetree.Migrate_%d_%d)(Migrate_parsetree.Migrate_%d_%d)\n" a b b a
@@ -515,3 +532,20 @@ include Register_migration(OCaml_411)(OCaml_412)
 (*$*)
 
 module OCaml_current = OCaml_OCAML_VERSION
+
+module Find_version = struct
+  type t = Impl of (module OCaml_version) | Intf of (module OCaml_version) | Unknown
+
+  let from_magic magic =
+    let rec loop = function
+      | [] -> Unknown
+      | (module Version : OCaml_version) :: tail ->
+          if Version.Ast.Config.ast_impl_magic_number = magic then
+            Impl (module Version)
+          else if Version.Ast.Config.ast_intf_magic_number = magic then
+            Intf (module Version)
+          else
+            loop tail
+    in
+    loop all_versions
+end
--- a/src/driver.ml
+++ b/src/driver.ml
@@ -591,43 +591,6 @@ let map_signature sg =
    | Entry points                                                    |
    +-----------------------------------------------------------------+ *)
 
-let mapper =
-  let module Js = Ppxlib_ast.Selected_ast in
-  (*$*)
-  let structure _ st =
-    Js.of_ocaml Structure st
-    |> map_structure
-    |> Js.to_ocaml Structure
-  in
-  (*$ str_to_sig _last_text_block *)
-  let signature _ sg =
-    Js.of_ocaml Signature sg
-    |> map_signature
-    |> Js.to_ocaml Signature
-  in
-  (*$*)
-  { Ocaml_common.Ast_mapper.default_mapper with structure; signature }
-;;
-
-let as_ppx_rewriter_main argv =
-  let argv = Caml.Sys.executable_name :: argv in
-  let usage =
-    Printf.sprintf "%s [extra_args] <infile> <outfile>" exe_name
-  in
-  match
-    Arg.parse_argv (Array.of_list argv) (Arg.align (List.rev !args))
-      (fun _ -> raise (Arg.Bad "anonymous arguments not accepted"))
-      usage
-  with
-  | exception Arg.Bad  msg -> Printf.eprintf "%s" msg; Caml.exit 2
-  | exception Arg.Help msg -> Printf.eprintf "%s" msg; Caml.exit 0
-  | () -> mapper
-
-let run_as_ppx_rewriter () =
-  perform_checks := false;
-  Ocaml_common.Ast_mapper.run_main as_ppx_rewriter_main;
-  Caml.exit 0
-
 let string_contains_binary_ast s =
   let test magic_number =
     String.is_prefix s ~prefix:(String.sub magic_number ~pos:0 ~len:9)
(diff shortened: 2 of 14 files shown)


src/driver.ml
```diff
@@ -547,43 +547,6 @@ let map_signature sg =
...
-  { Ocaml_common.Ast_mapper.default_mapper with structure; signature }
-;;
-
-let as_ppx_rewriter_main argv =
-  let argv = Caml.Sys.executable_name :: argv in
-  let usage =
-    Printf.sprintf "%s [extra_args] <infile> <outfile>" exe_name
-  in
-  match
-    Arg.parse_argv (Array.of_list argv) (Arg.align (List.rev !args))
```
pitag-ha, line 574: Why do we reverse the list of arguments here? I've preserved that, but I don't know why that's done.

src/driver.ml
```diff
@@ -919,16 +861,28 @@ let process_file (kind : Kind.t) fn ~input_name ~relocate ~output_mode ~embed_er
...
+          (input_fname, input_version, ast)
+        with exn when embed_errors ->
+          (input_fname, input_version, exn_to_extension exn ~kind) )
+    | Error (error, input_version) when embed_errors ->
+        (input_name, input_version, error_to_extension error ~kind)
+    | Error (error, _) ->
+        Location.raise_errorf
+          ~loc:(Location.in_file input_name)
+          "%s"
+          (Location.Error.message error)
```
pitag-ha, line 884: Is there a better way to do this?

pitag-ha:
The ocaml-ci arm32 failure isn't related to this PR.

There is a minor change in behavior that I haven't added to the changelog: if someone hands in a binary AST with an unknown OCaml version to the standalone, we now raise instead of packing that error into an extension point and returning the corresponding AST; and same if the input file doesn't exist. Should I mention that in the changelog?

ast/versions.ml
```diff
@@ -515,3 +532,20 @@ include Register_migration(OCaml_411)(OCaml_412)
 (*$*)
 
 module OCaml_current = OCaml_OCAML_VERSION
+
+module Find_version = struct
+  type t = Impl of (module OCaml_version) | Intf of (module OCaml_version) | None
```
jeremiedimino, line 537: While naming a constructor `None` is a valid thing to do, the `option` type is so pervasives that it seems best to not reuse the `None` constructor. In this case, `Uknown` seems like a good name (and even better IMO):

```suggestion
  type t = Impl of (module OCaml_version) | Intf of (module OCaml_version) | Unknown
```

src/driver.ml
jeremiedimino: Because the list is in reverse order as the order in which arguments are registered; `add_arg` adds the argument at the beginning of the list.

We could make this more explicit by hidding the `args` reference and only exposing a function that return the list of arguments in the right order. Then the `List.rev` would be closer to the `add_args` function, making the code clearer.

src/driver.ml
jeremiedimino: What about adding a `Location.Error.raise` function?

src/driver.ml
```diff
@@ -1315,6 +1279,33 @@ let standalone () =
...
+      try
+        rewrite_binary_ast_file input_fn output_fn;
+        Caml.exit 0
+      with exn ->
+        Location.report_exception Caml.Format.err_formatter exn;
+        Caml.exit 1 )
+  | [ help; _ ] when String.equal help "-help" || String.equal help "--help" ->
+      parse_input input ~valid_args ~incorrect_input_msg:usage;
+      (* The program should already have terminated with 0 *)
+      Printf.eprintf "Usage: %s\n%!" usage;
```
jeremiedimino, line 1302: We should make it explicit that we expect this to be unreachable by using `assert false`

src/utils.ml
```diff
@@ -105,29 +47,133 @@ module Intf_or_impl = struct
...
     | Intf sg ->
-      let sg = Selected_ast.To_ocaml.copy_signature sg in
-      let sg =
-        if add_ppx_context then
-          Ocaml_common.Ast_mapper.add_ppx_context_sig ~tool_name:"ppx_driver" sg
-        else
-          sg
-      in
-      Ast_io.Intf sg
+        let sg = Input_ast.To_ocaml.copy_signature sg in
```
jeremiedimino, line 136: If `add_ppx_context` is `true`, it seems that:

- we convert from the ppxlib AST version to the OCaml one
- we add the cookies
- then we convert back to the ppxlib version
- then here we convert again to the OCaml one

That seems inefficient. We should skip the last two steps.

src/utils.mli
```diff
@@ -0,0 +1,262 @@
...
+  val equal : t -> t -> bool
+end
+
+module Intf_or_impl  : sig
+  type t =
+  | Intf of signature
+  | Impl of structure
+
+  val map : t ->
+    < arg_label : arg_label -> arg_label;
```
jeremiedimino, line 21: Can we use `Ast_traverse.map` here to make the signature shorter?

jeremiedimino (changes requested):
Looks good overall. The code definitely looks more straightforward now!

jeremiedimino:
> There is a minor change in behavior that I haven't added to the changelog: if someone hands in a binary AST with an unknown OCaml version to the standalone, we now raise instead of packing that error into an extension point and returning the corresponding AST; and same if the input file doesn't exist. Should I mention that in the changelog?

It should go into the changelog, so that users can quickly identify which version introduced this change in behaviour.

But thinking about this again, packing such errors using the compiler AST versions seems like a reasonable behaviour and is more conservative, so I think we should do that instead. Could you do this change?

src/driver.ml
pitag-ha: Yes, good idea! I've done that now. The only downside that has is that if the initial exception was raised without `loc` parameter, here I can't add it anymore. But that only has an impact for `input doesn't exist` errors (see [here](https://github.com/ocaml-ppx/ppxlib/pull/205/files#diff-42ed2efe32ada9a3626b9feb98bca6c1e2643cb2e07c77d8c3f965a3022458e2R54)), where the input file is reported inside the error message anyways.

ast/versions.ml
pitag-ha: Sounds good, thanks!

src/driver.ml
pitag-ha: Yes, totally. Thanks!

src/utils.ml
pitag-ha: As we've discussed, I've squashed the last two conversions now if `add_ppx_context` is `true`.

src/utils.mli
pitag-ha: Yees, so much better! Thanks!

pitag-ha:
For the main standalone, I've also embedded loading errors as extension nodes into an output AST now instead of raising (as was done before and as you've suggested keeping on doing). Therefore, I've had to restructure the loading logic a bit. I've done that in a separate commit to make re-reviewing easier. But the idea would be to squash the last two commits before merging.

src/driver.ml
pitag-ha: Oh right, thanks! I guess that kind of function is already implemented and called `get_args` but not used here, right? Would it make sense to use it here the same way as done in `standalone_run_as_ppx_rewriter`?

src/driver.ml
```diff
@@ -699,69 +637,53 @@ let relocate_mapper = object
...
+    create_versioned_error_f fall_back_version fn ~f:(fun create_f ->
+      create_f "File is a binary ast for an unknown version of OCaml with magic number '%s'" unknown_magic)
+  | Error (System_error (error, fall_back_version)) ->
+    Error (error, fall_back_version)
+  | Error (Not_a_binary_ast (prefix_read_from_file, fallback_version)) ->
+    let parse_input =
+      if String.equal fn "-" then Ast_io.Source_code.Stdin prefix_read_from_file
+      else File fn
+    in
+    match Ast_io.Source_code.parse parse_input ~kind ~input_name with
```
jeremiedimino, line 680: One change compared to before: if the input is a source file rather than a binary AST, then we open it twice:

- once for checking whether it is a binary AST or a source file
- once to parse it

This has two implications:

1. it is slightly less efficient
2. there is a race condition; if the file is modified or deleted in between the parsing might fail

Given that this case is the most commonly used in dune builds (because dune calls the ppx driver on the source file), it seems worth optimising. The second might seem a bit far-fetched, however I have observed it in practice in the past and it was pretty difficult to debug. So it seems worth getting rid of the double file opening for this reason as well.

In conclusion, we should preserve the old behaviour and arrange the code so that the file is opened only once.

src/driver.ml
```diff
@@ -1371,6 +1316,31 @@ let standalone () =
     Caml.exit 1
 ;;
 
+let run_as_ppx_rewriter () =
```
jeremiedimino, line 1323: I'm wondering if we couldn't factorise `standalone_run_as_ppx_rewriter` and `run_as_ppx_rewriter` a bit more. They are starting to look pretty similar. WDYT?

src/driver.ml
```diff
@@ -670,23 +618,13 @@ let with_preprocessed_file fn ~f =
...
-  with_preprocessed_file fn ~f:(fun fn ->
-    if String.equal fn "-" then
-      f stdin
-    else
-      In_channel.with_file fn ~f)
+        match System.run_preprocessor ~pp ~input:fn ~output:tmpfile with
+          | Ok () -> f tmpfile
+          | Error (failed_command, fall_back_version) ->
+            create_versioned_error_f fall_back_version fn ~f:(fun create_f ->
+            create_f "Error while running external preprocessor\n \
```
jeremiedimino:
```suggestion
            create_f "Error while running external preprocessor\n\
```

(I'm guessing this space wasn't intentional)

jeremiedimino:
> The only downside that has is that if the initial exception was raised without loc parameter, here I can't add it anymore. But that only has an impact for input doesn't exist errors (see here), where the input file is reported inside the error message anyways.

About that, looking at the `Location.error` type in the compiler, it should be possible to write a function `update_loc : error -> Location.t -> error` that updates the location of an error.

jeremiedimino (changes requested):
Changes since last review are looking good. Just need to avoid the double file opening, and perhaps factorise a bit more the `*run_as_ppx` functions

jeremiedimino:
I also pushed a slightly more idiomatic version of `create_versioned_error_f`

src/driver.ml
pitag-ha: I guess `get_args` really only makes sense for the standalones. I've left that as it was.

src/driver.ml
pitag-ha: Thanks! I didn't see that performance issue.

src/driver.ml
pitag-ha: I've adapted `run_as_ppx_rewriter` to the workflow of `standalone_run_as_ppx_rewriter` now. That makes it a a little bit harder to read (more lines and less straight-forward), but it makes those two functions even more similar and I don't have to convert between arrays and lists anymore.

src/driver.ml
pitag-ha: Good catch

src/driver.ml
pitag-ha: Update: I've done it the other way around now.

pitag-ha:
Thanks! I think I've implemented all the changes you've suggested.

jeremiedimino (approved):
Looks all good!

PR #207, opened 2021-01-13
Drop travis CI in favor of ocaml-ci

NathanReb:
The only thing triggering build failures on ocaml-ci is the arm32 bug with the cinaps stanza. This will fix itself when dune gets a new release.


jeremiedimino (approved):
LGTM

PR #209, opened 2021-01-15
Fix cli cookies handling

pitag-ha:
Before, cookies handed in through the cli were added to the `Ast_mapper` cookies reference right away when reading the cli arguments. But in the case of a binary AST input, that reference gets reset to only contain the cookies from the ppx context when parsing that context ([here](https://github.com/ocaml/ocaml/blob/trunk/parsing/ast_mapper.ml#L916) the reference gets reset instead of appending to it). Therefore, cli cookies were ignored for binary AST inputs.

This commit fixes that by adding the cli cookies to the `Ast_mapper` cookies reference after parsing the ppx context.

Probably related to https://github.com/janestreet/ppx_expect/issues/26 and to the last comment in https://github.com/janestreet/ppx_inline_test/issues/1

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@ unreleased
 - Add Driver.V2: give access to expansion context in whole file transformation
   callbacks of `register_transformation` (#202, @pitag-ha)
 
+- Driver: take `-cookie` argument into account, also when the input is a
+  binary AST (@pitag-ha, #209)
+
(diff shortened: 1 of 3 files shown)


jeremiedimino (approved):
Looks good!

DIdn't realise that `drop_ppx_context_str` was resetting all the cookies, that's good to know.

PR #211, opened 2021-01-18
Fix ocaml-ci on 4.12

pitag-ha:
To not forget about that, I've recorded in issue #212 that, once 4.12 is released, we should fix the test. cc @NathanReb


jeremiedimino (approved):
Looks good

PR #213, opened 2021-01-21
Fix ppxlib.traverse

NathanReb:
We unintentionally broke traverse in #196 by not declaring it as a ppx anymore. This PR reverts this but also make it a deriver (which it is) instead of a rewriter.

Even though it was working perfectly well being declared as the latter, I think this will improve compat with `ppx_deriving`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 unreleased
 ----------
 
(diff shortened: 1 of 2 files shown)


jeremiedimino (approved):
Looks good

PR #215, opened 2021-01-22
Generate opam file using dune

NathanReb:
I cleaned up and reordered fields of the opam file before switching to the auto-generated one to ease the review so the per commit view should make it easier to spot errors!


jeremiedimino (approved):
LGTM

PR #216, opened 2021-01-27
Run deriving_inline code generation as part of @runtest

NathanReb:
This will have the side effect of verifying deriving_inline generated code is up to date in our CI.


jeremiedimino (approved):
LGTM

PR #223, opened 2021-02-15
Make cram tests private

pitag-ha:


test/driver/error_embedding/test.t/run.t
```diff
@@ -7,13 +7,13 @@ With the `-embed-errors` options, if a PPX raises, the first such exception
 is caught and packed into an output AST
 
   $ echo "let _ = [%raise]" > impl.ml
-  $ raiser -embed-errors impl.ml
+  $ ./../raiser.exe -embed-errors impl.ml
```
NathanReb, line 10: Isn't `../raiser.exe` enough here? Same for the other occurences in this file and some of the others!

NathanReb (commented):
Looks good, just one minor comment!

test/driver/error_embedding/test.t/run.t
pitag-ha: Good catch, thanks!

NathanReb (approved):
Thanks!

Can you add a changelog entry mentioning we fixed a bug where we would install unwanted public binaries?

NathanReb (approved):

PR #225, opened 2021-03-04
Add skip-hash-bang test

pitag-ha:
There are two hash bang syntaxes. Before, only one was tested. This adds a test for the other one.


jeremiedimino (approved):
I never realised we could do that!

PR #226, opened 2021-03-04
Inline skip hash bang

pitag-ha:
This was the only internal use of the Lexer module. So after this change the Lexer module can be removed from the official API.

This is to be merged after #225.

### DIFF
--- a/src/utils.ml
+++ b/src/utils.ml
@@ -98,7 +98,7 @@ module Ast_io = struct
         ; pos_bol   = 0
         ; pos_cnum  = 0
         };
-      Lexer.skip_hash_bang lexbuf;
+      Skip_hash_bang.skip_hash_bang lexbuf;
       let ast : Intf_or_impl.t =
         match kind with
         | Intf -> Intf (Parse.interface      lexbuf)
--- a/src/dune
+++ b/src/dune
@@ -37,3 +37,5 @@
  (deps gen-compiler_specifics)
  (action
   (run %{ocaml} %{deps} %{ocaml_version} %{targets})))
+
+(ocamllex skip_hash_bang)
--- a/src/skip_hash_bang.mli
+++ b/src/skip_hash_bang.mli
@@ -1,0 +1,1 @@
+val skip_hash_bang: Lexing.lexbuf -> unit
--- a/src/skip_hash_bang.mll
+++ b/src/skip_hash_bang.mll
@@ -1,0 +1,17 @@
+{
+open Lexing
+
+let update_loc lexbuf lines_to_skip =
+  let pos = lexbuf.lex_curr_p in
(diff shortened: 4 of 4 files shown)


src/utils.ml
```diff
@@ -1,5 +1,29 @@
...
+  first_lines
+
+let skip_hash_bang ic prefix_read =
+  if
+    String.length prefix_read > 1
+    && String.equal (String.sub prefix_read ~pos:0 ~len:2) "#!"
+  then
+    let first_lines = get_first_lines ic prefix_read in
+    if String.equal first_lines.(2) "!#" then ""
+    else first_lines.(1) ^ first_lines.(2)
```
jeremiedimino, line 24: I think we are missing a `\n` between the two lines.

src/utils.ml
```diff
@@ -88,8 +112,9 @@ module Ast_io = struct
         What we do instead is create a lexing buffer from the input channel and pre-fill
         it with what we read to do the test. *)
       let lexbuf = Lexing.from_channel ic in
-      let len = String.length prefix_read_from_source in
-      Bytes.blit_string ~src:prefix_read_from_source ~src_pos:0 ~dst:lexbuf.lex_buffer ~dst_pos:0
+      let pre_read = skip_hash_bang ic prefix_read_from_source in
+      let len = String.length pre_read in
+      Bytes.blit_string ~src:pre_read ~src_pos:0 ~dst:lexbuf.lex_buffer ~dst_pos:0
```
jeremiedimino, line 117: What about first blitting `prefix_read_from_source` into the lexing buffer, and then calling a lexer that does the same as the `skip_hash_bang` in the compiler? The lexer is a bit easier to read the the OCaml parsing code.

pitag-ha: That's a good idea! I didn't know that was possible. I've named that new lexer `Lexer_internal` for now to not shadow the `Lexer` module we're exposing at the moment. I would rename `Lexer_internal` to `Lexer` in the PR in which I remove the current `Lexer` module.

pitag-ha: I've just noticed that renaming it to `Lexer` won't be possible, since we want to keep the possibility to access the compiler-libs `Lexer` through Ppxlib.

src/utils.ml
pitag-ha: Since I've followed your idea to rewrite the function using ocamllex, I'll close this conversation.

src/utils.ml
jeremiedimino: TBH, i would call this module just `Skip_hash_bang`, since that's its only purpose

jeremiedimino (approved):
Looks good

src/utils.ml
pitag-ha: Oh yes, that makes sense. I've renamed it.

PR #227, opened 2021-03-04
Remove keyword table

pitag-ha:
That's a lighter API. And anyways, all users of `Lexer.keyword_table` only use the keyword table to check if a string is an OCaml keyword.

To be merged after #226.


pitag-ha:
There are currently two users of `Lexer.keyword_table`. I will send a patch to them after this PR gets approved.

src/keyword.ml
```diff
@@ -0,0 +1,65 @@
...
+      "lor";
+      "lxor";
+      "mod";
+      "land";
+      "lsl";
+      "lsr";
+      "asr";
+    ]
+  in
+  List.mem ~set:keywords word
```
jeremiedimino, line 65: Given that we only use the list inside this function, we could write as follow:

```
let is_keyword = function
  | "and" | "as" | ... | "asr" -> true
  | _ -> false
```

which would make it much faster.

jeremiedimino (changes requested):
Looks good

src/keyword.ml
pitag-ha: Right, I didn't think about that. Thanks!

pitag-ha:
I've sent patches to the two libraries that use `Lexer.keyword_table` at the moment: [one](https://github.com/janestreet/ppx_accessor/pull/3) to ppx_accessor and [one](https://github.com/janestreet/ppx_variants_conv/pull/9) to ppx_variants_conv

PR #228, opened 2021-03-04
Remove Lexer

pitag-ha:
After #227, `Lexer` was the same as the compiler-libs `Lexer`, so removing it won't do harm.

To be merged after #227.


jeremiedimino (approved):
Looks good

PR #230, opened 2021-03-08
Reduce the public API

pitag-ha:
This theoretic breaking change is not a real breaking change: none of the users in the [ppx universe](https://github.com/ocaml-ppx/ppx_universe) is using any of the modules that get removed from the API. Unless I'm missing something, those modules weren't meant for external use.

To be merged after #229.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,12 @@ unreleased
 - Remove `Lexer` from the API: it was the same as the compiler-libs
   `Lexer` (#228, @pitag-ha)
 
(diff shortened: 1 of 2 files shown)


pitag-ha:
We could also remove `Selected_ast`, but we need it ourselves currently due to our current [test](https://github.com/ocaml-ppx/ppxlib/blob/master/test/expect/expect_test.ml#L53) structure. Also, even though no-one is using them at the moment, the `Selected_ast` conversions might be useful to some external users.

jeremiedimino (approved):
Looks good

PR #231, opened 2021-03-11
Remove the internal Pprintast module

pitag-ha:
The internal Pprintast module used to augment the compiler-libs Pprintast module by class and module nodes as well as structure item and signature item. Class and module nodes aren't needed. Removing them and therefore the local copy of Pprintast will allow to restructure Ast_helper_light.


pitag-ha:
Is the Pprintast implementation of compilers < 4.08 different to the ones on compilers >= 4.08?

jeremiedimino:
I know that several bug were fixed in the past few years. Is that an issue for the testsuite?

pitag-ha:
I'm closing this due to the bug in `Pprintast` on old compilers (<4.08).

PR #238, opened 2021-03-18
Remove gen-compiler_specifics in ast/

pitag-ha:
`Clflags_helper` and `Misc_helper` aren't used. `Location_helper` isn't used internally. It was exposed through `Location` in the API, but no downstream user in the ppx universe is using it.


jeremiedimino (approved):
Looks good

PR #239, opened 2021-03-25
Allow to use "%a" with Location.Error.createf

mlasson:
# Description 

I had trouble using the function "Location.Error.createf" while migrating to ppxlib and I noticed what seems to be a minor regression introduced in commit 4693339ce7effc8dd5061f05019bf3adc1a2a908. The "restored" version does not allow for using "%a" custom formatters. 

Do not hesitate to close this PR if that was intentional.


jeremiedimino:
That seems reasonable to me, but I'll let @pitag-ha have a look when she's back.

pitag-ha (approved):
Thanks, that's for sure a good behavior for `Location.Error.createf`!

PR #241, opened 2021-04-15
Add a test for the Location.Error exception

pitag-ha:
The exception in the ppxlib `Location` module should be equivalnet to the exception in the compiler `Location` module, but isn't at the moment. The new test shows that.


jeremiedimino (approved):
Looks good!

PR #242, opened 2021-04-15
Location: fix exception equivalence

pitag-ha:
`raise_errorf` raises the compiler-libs exception. This commit ensures that that's equivalent to `Location.Error`. That's important to make sure that exceptions, that are raised via `raise_errorf` and caught later, are caught correctly.

This PR is built upon #241, the PR adding the corresponding test.


jeremiedimino (approved):
Looks good!

PR #244, opened 2021-05-03
Drop Syntaxerr from API

pitag-ha:
We were leaking the compiler-libs module `Syntaxerr` through the public API. I've checked and no downstream user in the ppx universe is using it.


jeremiedimino (approved):
LGTM

PR #246, opened 2021-05-21
Add a test to show wrong parse error reporting

pitag-ha:
The location isn't reported correctly in parse errors.


jeremiedimino (approved):
LGTM

Issue #221, opened 2021-02-10
Syntax errors are not displayed properly


kit-ty-kate:
Using the following test-case
```
$ cat dune
(executable
 (name test)
 (preprocess (pps ppxlib.metaquot)))
$ cat test.ml
let (let+) _ _ = Obj.magic ()
let (and+) _ _ = Obj.magic ()

let () =
  let+ x = ""
  let+ y = ""
  and+ z = ""
  in
  ()
```
and calling dune will result in the following, rather unhelpful, error message:
```
$ dune build --verbose
(cd _build/default && .ppx/0224ad3443a846e54f1637fccb074e7d/ppx.exe -o test.pp.ml --impl test.ml -corrected-suffix .ppx-corrected -diff-cmd - -dump-ast)
File "test.ml", line 1:
Error: Syntax error
```
in contrast, calling the compiler directly results in a more helpful error message:
```
$ ocamlopt test.ml
File "test.ml", line 6, characters 2-6:
6 |   let+ y = ""
      ^^^^
Error: Syntax error
```
----------
Test setup:
* `ppxlib` version: `0.22.0`
* `dune` version: `2.8.2`

jeremiedimino:
BTW, the first weird thing here is that the location is different:

- line 1 in the first case
- line 6, characters 2-6 in the second

maybe the non-printing of the file excerpt is due to the location being wrong. The fact that is has no characters information seems odd.

Armael:
This is most likely the symptom of a `Location.none` leaking out. In that case you still get the right filename because the printing code [will default to !Location.input_name](https://github.com/ocaml/ocaml/blob/trunk/parsing/location.ml#L190), but then you get no information (except `line 1` [for backward compatibility reasons](https://github.com/ocaml/ocaml/blob/trunk/parsing/location.ml#L212)).

aantron:
`git bisect` indicates that the bug was introduced in a50690c5f54ed9b89f6a16906f1b08155db1fe32.

Could you take a look? This is a major usability issue at the moment.

aantron:
A workaround in the meantime is to use ppxlib 0.20.0.

I have at least one project (Bisect_ppx) which requires a higher version, though.

pitag-ha:
This should be fixed on master now. I'll let you all know when making the next release.

aantron:
Thank you!

pitag-ha:
[...]

PR #247, opened 2021-05-21
Fix parse error location

pitag-ha:
When there is a source parse error, the file name responsible for the error is added to the error location. That's done to report the file name even if the error is due to the file not existing. Before, that was done by overwriting the error location by an empty ghost range located in the given file and therefore messing up the start and end positions. This commit fixes that.

This should fix the problem described in #221.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 unreleased
 ----------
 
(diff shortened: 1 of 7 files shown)


pitag-ha:
To be merged after #246.

jeremiedimino (approved):
Makes sense. Looks good!

PR #248, opened 2021-05-21
Improve Astlib documentation / comments

pitag-ha:
This adds documentation to the Astlib modules that didn't have any yet.


astlib/ast_mapper.mli
```diff
@@ -1,3 +1,5 @@
+(** Meta data related interface for a ppx rewriter *)
```
NathanReb, line 1: Maybe this module should be renamed as it doesn't have much in common with the original Ast_mapper anymore!

NathanReb (commented):
I just have a minor comment about some modules name that feel out of date now but otherwise it looks good!

astlib/ast_mapper.mli
pitag-ha: Yes, that's true, even this documentation shows that it's not really related to `Ast_mapper` anymore. I've renamed it to `Ast_metadata`. Thanks!

PR #249, opened 2021-05-31
Reintroduce minimal supported version in migration error messages

kit-ty-kate:
This was removed in https://github.com/ocaml-ppx/ppxlib/commit/8d54618d072f66ba7509e4a493492d8cc95524f5, however I feel like this was very helpful to know which version of OCaml the unsupported feature was introduced, to constrain it in the opam file for instance.

### DIFF
--- a/astlib/migrate_403_402.ml
+++ b/astlib/migrate_403_402.ml
@@ -35,7 +35,9 @@ let inject_predef_option label d =
 let from_loc { Location.txt = _; loc } = loc
 
 let migration_error loc missing_feature =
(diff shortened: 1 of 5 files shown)


pitag-ha:
Thanks, that information is indeed helpful and dropping it wasn't intentional! Could you sign the commit? Since your change brings back the old behavior and the change of behavior I introduced didn't get released, I don't think we need a changelog entry. I've added the 'no changelog' lable for the CI to pass.

pitag-ha (approved):
Thanks! Once you sign the commit, the CI should pass and we can merge.

PR #252, opened 2021-06-02
Make Astlib.Parser.token interchangable with Ocaml_common.Parser.token

kit-ty-kate:
This seems to be necessary to port https://github.com/janestreet/ppx_optcomp to ppxlib master. See in particular https://github.com/janestreet/ppx_optcomp/blob/v0.14/src/cparser.ml which contains the core of the issue.
While trying to port it I tried to change the use of `Parser.token` to `Ocaml_common.Parser` however I got stuck when `assert_no_attributes_in#structure_item` was called with the wrong type.

Feel free to discuss and/or close this as this is pure guess and I am no expert in the new astlib layer.
cc @pitag-ha

### DIFF
--- a/astlib/parser.mli
+++ b/astlib/parser.mli
@@ -1,4 +1,4 @@
-type token
+type token = Ocaml_common.Parser.token
 
 val use_file :
   (Lexing.lexbuf -> token) -> Lexing.lexbuf -> Parsetree.toplevel_phrase list


pitag-ha:
Thanks! I wasn't aware that we were already breaking `ppx_optcomp` and `visitors`! The problem is that they both need the `Parser.token` type to be concrete, so they would still break even with this PR. I've sent patches to them using `Parser.token` from compiler-libs as you've suggested: https://github.com/janestreet/ppx_optcomp/pull/14 and https://gitlab.inria.fr/fpottier/visitors/-/merge_requests/5. Let me know what you think.

PR #255, opened 2021-06-09
Make OMP a conflict

### LINKED ISSUE #251: master (ppxlib 1.0?) lost its ability to work with jbuilder
I can't find explicit mention of that anywhere so I'm assuming this is a bug. Using ppxlib master (https://github.com/ocaml-ppx/ppxlib/commit/3cff08bf007dca5d8cf8a71fb4dfbc593a8fc0ed), packages that previously worked totally fine with ppxlib.0.22.0 now fail with diverse errors related to ppx that were not run. e.g.:
```
#=== ERROR while compiling dockerfile-cmd.4.0.0 ===============================#
# context              2.1.0~beta4 | linux/x86_64 | ocaml-base-compiler.4.12.0 | file:///home/opam/opam-repository
# path                 ~/.opam/4.12.0/.opam-switch/build/dockerfile-cmd.4.0.0
# command              ~/.opam/4.12.0/bin/jbuilder build -p dockerfile-cmd -j 31
# exit-code            1
# env-file             ~/.opam/log/dockerfile-cmd-7-7566b3.env
# output-file          ~/.opam/log/dockerfile-cmd-7-7566b3.out
### output ###
#     ocamlopt src-cmd/.dockerfile_cmd.objs/dockerfile_cmd.{cmx,o} (exit 2)
# (cd _build/default && /home/opam/.opam/4.12.0/bin/ocamlopt.opt -w -40 -g -I src-cmd/.dockerfile_cmd.objs -I /home/opam/.opam/4.12.0/lib/astring -I /home/opam/.opam/4.12.0/lib/base/caml -I /home/opam/.opam/4.12.0/lib/bos -I /home/opam/.opam/4.12.0/lib/bytes -I /home/opam/.opam/4.12.0/lib/cmdliner -I /home/opam/.opam/4.12.0/lib/csv -I /home/opam/.opam/4.12.0/lib/dockerfile -I /home/opam/.opam/4.12.0/lib/fmt -I /home/opam/.opam/4.12.0/lib/fpath -I /home/opam/.opam/4.12.0/lib/logs -I /home/opam/.opam/4.12.0/lib/parsexp -I /home/opam/.opam/4.12.0/lib/ppx_sexp_conv/runtime-lib -I /home/opam/.opam/4.12.0/lib/result -I /home/opam/.opam/4.12.0/lib/rresult -I /home/opam/.opam/4.12.0/lib/seq -I /home/opam/.opam/4.12.0/lib/sexplib -I /home/opam/.opam/4.12.0/lib/sexplib0 -I /home/opam/.opam/4.12.0/lib/stdlib-shims -no-alias-deps -o src-cmd/.dockerfile_cmd.objs/dockerfile_cmd.cmx -c -impl src-cmd/dockerfile_cmd.pp.ml)
# File "src-cmd/dockerfile_cmd.ml", line 55, characters 50-65:
# 55 |   OS.File.write path (Sexplib.Sexp.to_string_hum (sexp_of_cmd_log cmd_log)) >>= fun () ->
#                                                        ^^^^^^^^^^^^^^^
# Error: Unbound value sexp_of_cmd_log
```
```
#=== ERROR while compiling shared-memory-ring.3.0.1 ===========================#
# context              2.1.0~beta4 | linux/x86_64 | ocaml-base-compiler.4.12.0 | file:///home/opam/opam-repository
# path                 ~/.opam/4.12.0/.opam-switch/build/shared-memory-ring.3.0.1
# command              ~/.opam/4.12.0/bin/jbuilder build -p shared-memory-ring -j 31
# exit-code            1
# env-file             ~/.opam/log/shared-memory-ring-7-eb6da2.env
# output-file          ~/.opam/log/shared-memory-ring-7-eb6da2.out
### output ###
#       ocamlc lib/.shared_memory_ring.objs/console_ring.{cmo,cmt} (exit 2)
# (cd _build/default && /home/opam/.opam/4.12.0/bin/ocamlc.opt -w -40 -g -bin-annot -I lib/.shared_memory_ring.objs -I /home/opam/.opam/4.12.0/lib/bigarray-compat -I /home/opam/.opam/4.12.0/lib/cstruct -I /home/opam/.opam/4.12.0/lib/stdlib-shims -no-alias-deps -o lib/.shared_memory_ring.objs/console_ring.cmo -c -impl lib/console_ring.pp.ml)
# File "lib/console_ring.ml", line 23, characters 5-12:
# 23 | 		[%%cstruct
#           ^^^^^^^
# Error: Uninterpreted extension 'cstruct'.
```
I'm guessing this is a consequence of https://github.com/ocaml-ppx/ppxlib/pull/187

### LINKED ISSUE #254: ppxlib master can't mix with other types of ppx derivers anymore
For instance, using ppxlib master to build the latest version of pgocaml_ppx:
```
#=== ERROR while compiling pgocaml_ppx.4.2.2 ==================================#
# context              2.1.0~beta4 | linux/x86_64 | ocaml-base-compiler.4.12.0 | file:///home/opam/opam-repository
# path                 ~/.opam/4.12.0/.opam-switch/build/pgocaml_ppx.4.2.2
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p pgocaml_ppx -j 31
# exit-code            1
# env-file             ~/.opam/log/pgocaml_ppx-15-f6baaa.env
# output-file          ~/.opam/log/pgocaml_ppx-15-f6baaa.out
### output ###
# File "ppx/dune", line 6, characters 4-54:
# 6 |     (pps ppx_tools_versioned.metaquot_408 ppx_optcomp))
#         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
# Error: Too many incompatible ppx drivers were found: ocaml-migrate-parsetree
# and ppxlib.
```
I think this is a similar issue to #251

kit-ty-kate:
Since #187 ppxlib does not depend on OMP anymore. To work around https://github.com/ocaml-ppx/ppxlib/issues/254 it needs to be a conflict instead to prevent OMP1 mixing with ppxlib

Fixes #254
Fixes #251

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,9 @@ unreleased
 
 - `Location`: add `set_filename` and `Error.get_location` (#247, @pitag-ha)
 
-- Drop dependency on OMP (#187, @pitag-ha)
+- Drop dependency on OMP2 (#187, @pitag-ha)
+
+- Make OMP1 a conflict (#255, @kit-ty-kate)
 
 - Drop `Syntaxerr` from the public API. Doesn't affect any user in the
   [ppx universe](https://github.com/ocaml-ppx/ppx_universe) (#244, @pitag-ha)
--- a/dune-project
+++ b/dune-project
@@ -26,6 +26,8 @@
   (cinaps (and :with-test (>= v0.12.1)))
   (base :with-test)
   (stdio :with-test))
+ (conflicts
+  (ocaml-migrate-parsetree (< 2.0.0)))
  (synopsis "Standard library for ppx rewriters")
  (description "Ppxlib is the standard library for ppx rewriters and other programs
 that manipulate the in-memory reprensation of OCaml programs, a.k.a
--- a/ppxlib.opam
+++ b/ppxlib.opam
@@ -33,6 +33,9 @@ depends: [
   "stdio" {with-test}
   "odoc" {with-doc}
 ]
+conflicts: [
+  "ocaml-migrate-parsetree" {< "2.0.0"}
+]
 build: [
   ["dune" "subst"] {dev}
   [


NathanReb:
I don't think that's the right approach. It happens that they are co-installable but they can't be used together, with dune, on the same set of files.

Adding this conflict makes them non co-installable which is much more restrictive. Doing so we prevent people to do smooth migration. For instance without the conflict one can deal with this in to different ways:
1. They really want to use OMP1 and ppxlib based ppx-es on the same set of files: they can add a `ppxlib < 0.23.0` bound in their dependencies. That means they'll get an OMP1 based version of ppxlib and will have to stick to old versions of the ppx-es they use. That's pretty much the current situation but they need to add a bound in their opam file.
2. They can afford to extract different libraries that use the different driver. For instance that can be helpful for users that really rely on old ppx-es that haven't migrated yet. They extract a separate library just to use that specific ppx and can then include it in their main library. That allows them to use recent ppx-es in the rest of their codebase.

Some people have asked why we didn't release omp2 as a separate package so it would be co-installable with omp1. This is exactly what is happening here, they are co-installable but provide separate drivers, which means you can't use them together on the same set  of files. I think it's less of restriction and therefore is preferable.

I agree people might be surprised when getting the "different drivers" error but I believe this is better solved with good documentation here.

@kit-ty-kate @pitag-ha what do you think?

kit-ty-kate:
> I don't think that's the right approach. It happens that they are co-installable but they can't be used together, with dune, on the same set of files.
> 
> Adding this conflict makes them non co-installable which is much more restrictive. Doing so we prevent people to do smooth migration. For instance without the conflict one can deal with this in to different ways:
> 
>     1. They really want to use OMP1 and ppxlib based ppx-es on the same set of files: they can add a `ppxlib < 0.23.0` bound in their dependencies. That means they'll get an OMP1 based version of ppxlib and will have to stick to old versions of the ppx-es they use. That's pretty much the current situation but they need to add a bound in their opam file.
> 
>     2. They can afford to extract different libraries that use the different driver. For instance that can be helpful for users that really rely on old ppx-es that haven't migrated yet. They extract a separate library just to use that specific ppx and can then include it in their main library. That allows them to use recent ppx-es in the rest of their codebase.
> 
> 
> Some people have asked why we didn't release omp2 as a separate package so it would be co-installable with omp1. This is exactly what is happening here, they are co-installable but provide separate drivers, which means you can't use them together on the same set of files. I think it's less of restriction and therefore is preferable.
> 
> I agree people might be surprised when getting the "different drivers" error but I believe this is better solved with good documentation here.

In theory this is a really good analysis but in practice this is quite hard to get the right constraints. Packages that use several ppxs will have their dependencies set to something like:
```
depends: [
  ...
  "ppx_blih" {>= "1.0"}
  "ppx_blah" {>= "0.5"}
]
```
In such a package you don't have explicit dependencies on ppxlib, only individual ppxs. Those ppx might have used OMP1 directly without ppxlib in the past for instance, and adding a dependency or constraint on ppxlib can break people's packages (e.g. lock packages or specific constraint sets).
For instance, in this particular example, if `ppx_blih` only depends on ppxlib since 2.0 and `ppx_blah` only depends on ppxlib since 1.0, to get the right constraint you have to say something like:
```
  ("ppx_blih" {>= "1.0" & < "2.0"} & "ppx_blah" {>= "0.5" & < "1.0") | # both without ppxlib
  ("ppx_blih" {>= "2.0"} & "ppx_blah {>= "0.5" & < "1.0"} & "ppxlib" {< "0.23.0"}) | # only ppx_blih with ppxlib
  ("ppx_blih" {>= "1.0" & < "2.0"} & "ppx_blah {>= "1.0"} & "ppxlib" {< "0.23.0"}) | # only ppx_blah with ppxlib
  ("ppx_blih" {>= "2.0"} & "ppx_blah" {>= "1.0"}) # both with ppxlib
```
this is already quite a mess but now you have to imagine that with much more complex constraints and more ppxs. This is already quite tricky for human to read and write this correctly but also for opam's solver.

NathanReb:
You're right that a dependency isn't the right solution for ppx users here, a conflict would be best.

I definitely understand your point and why we'd want to go down that road but I would much prefer a different solution here such as improving error messages in dune when this error shows up.

I'm happy to help adjust the constraints in opam for packages that suffer from this  as well if that helps but I don't think opam's metadata should prevent people from installing packages that can be installed and used together under certain circumstances, especially considering some users have been asking for this ability in the past. Once the metadata are out there, it's very hard for users to work around them with the current tooling unfortunately.

That being said I'm okay with whatever we choose here as I think we'd much rather put our joint effort into making everyone migrate to the latest ppxlib and solve all of this once and for all!

I'll let @pitag-ha make the final decision on this!

avsm:
I have a preference for the most precise constraints that will result in co-installable packages, but it's a lot of work to get there with the number of packages involved, @NathanReb.  On the bright side, it's a one-off set of changes and all of the historical packages will then have the most accurate constraints.

I'd suggest scoping out how many packages would need to change, @NathanReb, and then if it's just too many to go with the less precise conflicts that @kit-ty-kate suggests. It may also be worth polling on discuss for users that concretely want to use omp1 alongside ppxlib too, to help determine how much effort to put into this.

pitag-ha:
> For instance, in this particular example, if `ppx_blih` only depends on ppxlib since 2.0 and `ppx_blah` only depends on ppxlib since 1.0, to get the right constraint you have to say something like:
>```
 > ("ppx_blih" {>= "1.0" & < "2.0"} & "ppx_blah" {>= "0.5" & < "1.0") | # both without ppxlib
  >("ppx_blih" {>= "2.0"} & "ppx_blah {>= "0.5" & < "1.0"} & "ppxlib" {< "0.23.0"}) | # only ppx_blih with ppxlib
  >("ppx_blih" {>= "1.0" & < "2.0"} & "ppx_blah {>= "1.0"} & "ppxlib" {< "0.23.0"}) | # only ppx_blah with ppxlib
 > ("ppx_blih" {>= "2.0"} & "ppx_blah" {>= "1.0"}) # both with ppxlib
>```

@kit-ty-kate, would you mind explaining why the four lines are necessary? I'm thinking if those four lines could be simplified to something like 
```
("ppx_blih" {>= "2.0"} & "ppx_blah" {>= "1.0"}) |
("ppx_blih" & "ppx_blah" & <conflict with "ppxlib" >= "0.23.0">)
```
I'm not sure if there's an opam way to express the second line in a simple way. If not, maybe we could write a virtual package, say `old-ppx-era`, whose only purpose is to conflict with `ppxlib >= 0.23.0` and replace `<conflict with "ppxlib" >= "0.23.0">` by a dependency on `old-ppx-era`?
The 2-line constraint I've written would work for any number of PPX dependencies and would make a lot of sense given that 0.23.0 will be the ppxlib version marking the end of the OMP era: either all PPXs you use are based on ppxlib rather than OMP or you can only use a ppxlib (and PPXs) from the OMP era.

pitag-ha:
> especially considering some users have been asking for this ability in the past

@NathanReb, do you remember which OMP1 based packages folks wanted to use together with ppxlib based packages?

pitag-ha:
> I'd suggest scoping out how many packages would need to change

That sounds good. And thanks for the rest of the analysis, @avsm!

Let's distinguish between currently published packages on opam (for which we'd need to add the constraint ourselves) and unpublished/future packages (for which we'd need to make sure that adding the constraint would be easy -> documentation). 
About the former: @kit-ty-kate, do you happen to know the opam packages that are affected? Or is [this health check of yours](http://check.ocamllabs.io:8080/?comp=4.12&available=4.12&show-failures-only=true&sort-by-revdeps=true&maintainers=&logsearch=&logsearch_comp=4.12) a good place to find them (by ignoring the packages who's failure is unrelated to ppxlib)?
About the latter: asking on discuss sounds like a good idea! I'll do that once I have a clearer picture about the opam packages.
 
I think, with the exception of one PPX that doesn't have a mergeable port yet and some PPXs that don't have their port released yet, there shouldn't be any PPXs on opam left anymore that depend on OMP1 and have reverse dependencies on opam. So hopefully soon this will all be simpler.

pitag-ha:
Since I'm going to post about this on discuss and want to link in the post to this discussion, I've just rebased and edited the PR title and first comment to make clear this discussion is only about the question if to make OMP a conflict or not (@kit-ty-kate was ok with that).

pitag-ha:
@NathanReb, I would merge this PR now. Is that ok with you? I [asked on discuss if folks still need to use OMP and ppxlib in the same project](https://discuss.ocaml.org/t/do-you-need-ppxlib-and-ocaml-migrate-parsetree-in-the-same-project/8207/4) and almost no one has answered. I think even though from a theoretic point of view it would make more sense to leave the conflict at the level of the build system, practically I don't think the benefit would way out the hassle.

kit-ty-kate:
> > For instance, in this particular example, if `ppx_blih` only depends on ppxlib since 2.0 and `ppx_blah` only depends on ppxlib since 1.0, to get the right constraint you have to say something like:
> > ```
> > ("ppx_blih" {>= "1.0" & < "2.0"} & "ppx_blah" {>= "0.5" & < "1.0") | # both without ppxlib
> > ("ppx_blih" {>= "2.0"} & "ppx_blah {>= "0.5" & < "1.0"} & "ppxlib" {< "0.23.0"}) | # only ppx_blih with ppxlib
> > ("ppx_blih" {>= "1.0" & < "2.0"} & "ppx_blah {>= "1.0"} & "ppxlib" {< "0.23.0"}) | # only ppx_blah with ppxlib
> > ("ppx_blih" {>= "2.0"} & "ppx_blah" {>= "1.0"}) # both with ppxlib
> > ```
> 
> @kit-ty-kate, would you mind explaining why the four lines are necessary? I'm thinking if those four lines could be simplified to something like
> 
> ```
> ("ppx_blih" {>= "2.0"} & "ppx_blah" {>= "1.0"}) |
> ("ppx_blih" & "ppx_blah" & <conflict with "ppxlib" >= "0.23.0">)
> ```

Precisely because opam doesn't know how to say "conflict with ppxlib >= 0.23.0 in this precise case". It can only conflict in every cases or depend on it with a constraint (which wouldn't be correct for the < 2.0 & < 1.0 case in the above example).

(sorry for taking so long to reply)

pitag-ha:
> Precisely because opam doesn't know how to say "conflict with ppxlib >= 0.23.0 in this precise case".

Ok, that's what I thought. That's what I proposed
> maybe we could write a virtual package, say old-ppx-era, whose only purpose is to conflict with ppxlib >= 0.23.0 and replace 
> <conflict with "ppxlib" >= "0.23.0"> by a dependency on old-ppx-era

for. But in my opinion all of that would be too much hassle for something that doesn't seem to have much practical value anymore.

I'll still wait a little bit to see if someone objects and if not I'll merge this afternoon.

NathanReb (approved):
You all spent much more time investigating and thinking about this than I did so I trust your judgement. I think our time is best spent porting the remaining ppx-es anyway!

For the record, the conflict I had in mind for users would be simply something like:
```
conflicts: [
  "ocaml-migrate-parsetree" {< "2.0"}
]
```
or
```
conflicts: [
  "ppxlib" {>= "0.23"}
]
```
I.e. if a user runs into the driver conflict error, we should suggest they explicitly choose one of the two drivers by adding either one of those constraints or deal with it at the library level. They wouldn't have to write complex conflicts, simply to make it impossible for opam to pick one of the conflicting drivers until they sort things out.

pitag-ha:
> I think our time is best spent porting the remaining ppx-es anyway!

Yes!

> For the record, the conflict I had in mind for users would be simply something like:

> ```
> conflicts: [
>   "ocaml-migrate-parsetree" {< "2.0"}
> ]
> ```
>or
> ```
> conflicts: [
>  "ppxlib" {>= "0.23"}
> ]
> ```

Aah ok, I see. I think if someone runs into this problem, then if they define a conflict with `OMP2` inside their project, they will make it impossible for opam to find any solution. It might help to find which PPX based on `OMP1` is (transitively) used inside their project though (by reading the opam error message), but it wouldn't really solve anything.

pitag-ha (approved):

NathanReb:
> Aah ok, I see. I think if someone runs into this problem, then if they define a conflict with OMP2 inside their project, they will make it impossible for opam to find any solution. It might help to find which PPX based on OMP1 is (transitively) used inside their project though (by reading the opam error message), but it wouldn't really solve anything.

Or the other way around, they can prevent opam from selecting recent ppxlib and ppx-es and stick to OMP 1 based versions which solves everything if they're not willing to upgrade.

pitag-ha:
> Or the other way around, they can prevent opam from selecting recent ppxlib and ppx-es and stick to OMP 1 based versions

Yes, I agree that if we didn't add the `OMP1` conflict, advising folks to simply add 
```
conflicts: [
 "ppxlib" {>= "0.23"}
]
```
to their project when they run into this problem, would be the right thing to do. But anyways that shows that we wouldn't really win that very much in comparison with just making `OMP1` a conflict ourselves: if we went down that road, recent ppxlib would only become compatible with `OMP1` PPXs themselves (on the package manager level), but it would stay incompatible with most projects using `OMP1` PPXs (since most of them would add that conflict). So we would only win compatibility with the highest `OMP1` dependency level.

Issue #253, opened 2021-06-02
Release plan


kit-ty-kate:
Given the current state of ppxlib master (many packages are broken downstream when using it for different reasons) I am wondering if it would not be more prudent to backport a handful of fixes (e.g. https://github.com/ocaml-ppx/ppxlib/pull/242 and https://github.com/ocaml-ppx/ppxlib/pull/247 on top of my head) + the OCaml 4.13 support on top of ppxlib.0.22.0 and make it ppxlib.0.23.0.
Then continue working on master and release the next major version later.

This is just my educated guess based on what I'm seeing and past experiences. I'm hoping to avoid having to wait for things to get fixed to release OCaml 4.13 and put more stress on the ppxlib devs than needed.

kit-ty-kate:
cc @pitag-ha

pitag-ha:
Sorry for the delay in answering (I was on vacation). And by the way, thanks for all your help with this!

I'd like to release as soon as possible, particularly because of [the bug messing up parse errors](https://github.com/ocaml-ppx/ppxlib/issues/221) and for the 4.13 support. If we can't/shouldn't release "as normal" (which I think is the case), what you're saying sounds like a good idea to me. To clarify, I think the packages that are broken downstream are
-  4 packages I've sent patches to (the latter 2 just now): [patch](https://github.com/janestreet/ppx_variants_conv/pull/9) to `ppx_variants_conv`, [patch](https://github.com/janestreet/ppx_variants_conv/pull/9) to `ppx_accessor`, [patch](https://github.com/janestreet/ppx_optcomp/pull/14) to `ppx_optcomp`, [patch](https://gitlab.inria.fr/fpottier/visitors/-/merge_requests/5) to `visitors`
- `ocamlformat`: I've discussed that with @gpetiot and he's fixed the problem by [dropping the ppxlib dependency in ocamlformat](https://github.com/ocaml-ppx/ocamlformat/pull/1669)
- any package building with `jbuilder`. That probably only effects old downstream versions. Yet, fixing that might take a while.

Am I missing package(s)?

pitag-ha:
I'm closing this since those releases have already been done now. Thanks for your help, @kit-ty-kate!

PR #256, opened 2021-06-09
Backport location bug fix and prepare release

pitag-ha:
There's currently a [bug due to which no parse error is reported correctly when using ppxlib](https://github.com/ocaml-ppx/ppxlib/issues/221). I'm backporting its fix to 0.21, because making a release over master is a little complicated at the moment (see [here](https://github.com/ocaml-ppx/ppxlib/issues/253#issuecomment-856169765)).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,5 @@
-unreleased
-----------
+0.21.1 (09/06/2021)
+-------------------
 
 - Fix location in parse error reporting (#256, @pitag-ha)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 unreleased
 ----------
 
+- Fix location in parse error reporting (#256, @pitag-ha)
+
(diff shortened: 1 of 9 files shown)


pitag-ha:
Just for the record: this was a mistake. I meant to backport it to the last version, which is 0.22.0, not 0.21.0. But now that this is done, I can as well just release it, too.

CHANGES.md
```diff
@@ -1,5 +1,7 @@
-unreleased
-----------
+0.21.1 (09/06/2021)
```
NathanReb, line 1: I would leave this as `unreleased` and replace it only when you actually go through the release. It's okay to push directly on the 0.21 branch to update the changelog's latest entry before cutting the release.

NathanReb (approved):
Looks good to me!

CHANGES.md
pitag-ha: Ok, thanks, I'll do it like that for 0.22!

PR #261, opened 2021-06-21
Correctly pretty print `Psig_typesubst`.

matthewelse:
ppxlib was previously incorrectly pretty printing the following ast:

```ocaml
type t
type other := t
```

as 

```ocaml
type t
type nonrec other := t
```

The `nonrec` isn't required here, and will just be rejected by the OCaml compiler. This PR updates `pprintast` to correctly omit this flag.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -39,8 +39,8 @@ unreleased
 - Fix in `Location`: make `raise_errorf` exception equivalent to exception
   `Error` (#242, @pitag-ha)
 
-- Fix in `Pprintast`: correctly pretry print local type substitutions, e.g. type
-  t := ... (#261, @matthewelse)
+- Fix in `Pprintast`: correctly pretty print local type substitutions, e.g.
+  type t := ... (#261, @matthewelse)
 
 0.22.0 (04/02/2021)
 -------------------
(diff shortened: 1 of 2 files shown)


pitag-ha (approved):
That's indeed a bug. Thanks for finding and fixing it!

Could you add a changelog entry?

pitag-ha:
I was about to say that it might be good to also upstream the fix to the compiler, but I've just seen you've already done that. Thanks!

matthewelse:
Thanks for reviewing so quickly @pitag-ha - hopefully the change log entry is sufficient

CHANGES.md
```diff
@@ -39,6 +39,9 @@ unreleased
 - Fix in `Location`: make `raise_errorf` exception equivalent to exception
   `Error` (#242, @pitag-ha)
 
+- Fix in `Pprintast`: correctly pretry print local type substitutions, e.g. type
+  t := ... (#261, @matthewelse)
```
pitag-ha, line 43: > hopefully the change log entry is sufficient

Yes, it is, thanks! There's just a typo I'm suggesting to remove. And the line break is to have the code example all in one line, but we can also leave it as is

```suggestion
- Fix in `Pprintast`: correctly pretty print local type substitutions, e.g.
  type t := ... (#261, @matthewelse)
```

PR #262, opened 2021-06-22
Move Keyword to Astlib

pitag-ha:
For Astlib in ppxlib it's ok to hardcode the keywords since the keyword table has been stable since 4.04. When upstreaming Astlib to the compiler, we can see if we can use the compiler [`Parsing.Lexer.is_keyword` function](https://github.com/ocaml/ocaml/blob/trunk/parsing/lexer.mll#L232), which isn't public at the moment.


astlib/keyword.mli
```diff
@@ -0,0 +1,2 @@
+(** Check if a string is an OCaml keyword. *)
+    val is_keyword : string -> bool
```
NathanReb, line 2: This probably shouldn't be indented!

NathanReb (approved):
Looks good to me!

astlib/keyword.mli
pitag-ha: Good catch!

I'm glad that soon we'll be using ocamlformat!

PR #263, opened 2021-06-24
Drop Parser

pitag-ha:
The module `Parser` is too specific for Astlib. For normal parsing, one can use module `Parse`.


pitag-ha:
I've sent a [patch to `embedded_ocaml_tamplates`](https://github.com/EmileTrotignon/embedded_ocaml_templates/pull/3) which was the only user in the ppx_universe left being affected by this change (`ppx_optcomp` and `visitors` were already patched when we made `Parser.token` abstract).

pitag-ha:
The `embedded_ocaml_templates` patch has been merged and released now, so this shouldn't break anyone anymore.

NathanReb (approved):

PR #264, opened 2021-06-24
Add `Ast_pattern.esequence`

matthewelse:
This PR adds support for matching on a list of subsequent expressions that form a sequence, as a natural analogue to `Ast_builder.esequence`.

### DIFF
--- a/src/ast_pattern.ml
+++ b/src/ast_pattern.ml
@@ -224,7 +224,7 @@ let esequence (T f) = T (fun ctx _loc e k ->
   let rec parse_seq expr acc =
     match expr.pexp_desc with
     | Pexp_sequence (expr, next) ->
-      (parse_seq next (expr :: acc))
+      parse_seq next (expr :: acc)
     | _ -> expr :: acc
   in
   k (List.rev_map (parse_seq e []) ~f:(fun expr -> f ctx expr.pexp_loc expr (fun x -> x))))
--- a/src/ast_pattern.ml
+++ b/src/ast_pattern.ml
@@ -227,8 +227,8 @@ let esequence (T f) = T (fun ctx _loc e k ->
       expr :: (parse_seq next)
     | _ -> [ expr ]
   in
-  k (List.map (parse_seq e) ~f:(fun expr -> f ctx expr.pexp_loc expr k))
-)
+  k (List.map (parse_seq e) ~f:(fun expr -> f ctx expr.pexp_loc expr (fun x -> x))))
+;;
 
 let of_func f = (T f)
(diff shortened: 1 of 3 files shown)


src/ast_pattern.ml
```diff
@@ -220,5 +220,15 @@ let elist (T f) = T (fun ctx _loc e k ->
   k (List.map l ~f:(fun x -> f ctx x.Parsetree.pexp_loc x (fun x -> x))))
 ;;
 
+let esequence (T f) = T (fun ctx _loc e k ->
+  let rec parse_seq expr =
+    match expr.pexp_desc with
+    | Pexp_sequence (expr, next) ->
+      expr :: (parse_seq next)
+    | _ -> [ expr ]
```
pitag-ha, line 228: Do you think it might be worth making a tail recursion out if this?

pitag-ha:
Do you have a concrete use case for this in mind?

matthewelse:
A colleague of mine wanted this feature - I don't know exactly what they were trying to do, but it seemed like a simple enough thing to include in ppxlib itself

pitag-ha:
Do you know why your colleague can't use `elist` / prefers `esequence` over `elist`? Do they want to make sure that the first `n-1` expressions return `unit`?

matthewelse:
`elist` matches a list, right?

```
let x = [1; 2; 3]
```

I'm trying to match the following:

```
let () =
  do_first ();
  do_second ();
  do_third ()
```

As far as I can tell, there's no way of getting a list of expressions out of the second example without `esequence`, or some hand written code.

pitag-ha:
Ok, I can see now why to choose that syntax instead of a list. Thanks! To me adding this feature sounds good.

Have you seen [this comment](https://github.com/ocaml-ppx/ppxlib/pull/264#discussion_r658905374)?

src/ast_pattern.ml
pitag-ha: Similarly to [here](https://github.com/ocaml-ppx/ppxlib/blob/master/src/ast_pattern.ml#L209).

matthewelse:
Thanks @pitag-ha. I think this should now resolve the comment.

pitag-ha (approved):
Thanks @matthewelse! LGTM. Since I'm not very familiar with `Ast_pattern`, I'll wait some days to merge this in case @jeremiedimino or @NathanReb want to say something about it before.

matthewelse:
Thanks, sounds great!

jeremiedimino:
`esequence` seems like a good addition to me

PR #266, opened 2021-07-02
add Location.mknoloc

tatchi:
See https://github.com/ocaml-ppx/ppx_import/pull/54#discussion_r653450472

### DIFF
--- a/src/location.mli
+++ b/src/location.mli
@@ -42,6 +42,8 @@ type nonrec 'a loc = 'a loc =
   ; loc : t
   }
 
+val mknoloc : 'a -> 'a loc
+
 val compare_pos : Lexing.position -> Lexing.position -> int
 val min_pos : Lexing.position -> Lexing.position -> Lexing.position
 val max_pos : Lexing.position -> Lexing.position -> Lexing.position
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -45,6 +45,8 @@ unreleased
 - Add `Ast_pattern.esequence`, for matching on any number of sequenced
   expressions e.g. `do_a (); do_b (); ...`. (#264, @matthewelse)
 
+- Add `Location.mknoloc` (#266, @tatchi)
+
 0.22.0 (04/02/2021)
 -------------------
 
--- a/src/location.ml
+++ b/src/location.ml
@@ -62,6 +62,11 @@ type nonrec 'a loc = 'a loc =
   ; loc : t
   }
 
+let mknoloc (txt : 'a) : 'a loc =
+  { txt
+  ; loc = none
+  }
+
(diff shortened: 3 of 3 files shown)


src/location.mli
```diff
@@ -42,6 +42,8 @@ type nonrec 'a loc = 'a loc =
   ; loc : t
   }
 
+val mknoloc : 'a -> 'a loc
```
pitag-ha, line 45: Could you add documentation? I think it would be good also to stress in the documentation that this returns a ghost located value and should be avoided and only be used if one really doesn't need to locate the value.

pitag-ha (commented):
Thanks @tatchi! If we make clear in the documentation that `mknoloc` should only be used causiously, to me adding it sounds good. I've also already had to inline it in at least 2 PPXs I ported (`jsoo` and `bisect_ppx`).

NathanReb:
Thanks for working on this, we always appreciate contributions to ppxlib! Unfortunately I'm not sure we should merge this particular change.

I am rather opposed to it because using `none` locations in ppx generated code have been causing trouble to a lot of ppx users. Authors of ppx-es should always attach a location to the generated fragments of ASTs. If there's no particular location that would make sense to reuse (such as the location of the extension point or the `[@@deriving ...]` attribute), it should at least use a location pointing to the right file.

With that in mind, I don't think ppxlib should make it easy for authors to generate AST nodes with `none` locations. If anything it should make it harder, or impossible.

I'm well aware that writing code sensibly associating the right locations can be tedious but ppxlib does provide helpers 
for this. One can for instance use `Ast_builder.Make` that will provide a similar interface to `Ast_builder.Default` but with the `~loc` argument already set. I seem to recall there's also a function somewhere to recursively set all locations in an ASTfragment.

If I understand correctly, this is to use in the `ppx_import` port. I completely get that you might not want to change the behaviour in the port, in which case I'd suggest inlining such a function in `ppx_import`. If you're willing to set locations instead, I think it'd be a huge improvement to `ppx_import`!

pitag-ha:
Yes, what you're saying convinces me, @NathanReb. I can only think of two cases where `mknoloc` is ok: a port to `ppxlib` that doesn't want to change the behavior; and a PPX that generates a whole file and pprints it. But that doesn't justify making it easy for PPX authors to generate AST nodes with `none` locations in general.

Does it also sound convincing to you, @tatchi, what @NathanReb says?

I'm sorry for the whole confusion! The good thing is that the conversation about this is valuable, at least for me!

tatchi:
> Does it also sound convincing to you, @tatchi, what @NathanReb says?

Yes, that definitely makes sense to me :) I'll move forward and close this PR. Thank you both for your inputs 
🤗

PR #267, opened 2021-07-07
Move `Pprintast` into `Astlib`

pitag-ha:
Currently `Pprintast` prints the AST version that we use internally. Printing is the "reverse operation" of parsing though and it would therefore be more coherent to print the AST version that's parsed, i.e. the compiler version. To achieve that behavior on new compilers, `Pprintast` needs to be moved into the compiler `Astlib`.

If we wanted to do the same on old compilers, we'd have to write a `Pprintast` module for every internal `Astlib` AST version. Instead, for old compilers, we keep the old behavior of having one fixed `Pprintast` for our internal AST version. For coherency with the new behavior, this commit moves that module into `Astlib`. Also, `Pprintast` inside `ppxlib` can be removed when the rest of our `Astlib` copy can be removed, i.e. when we drop support for all compilers without `Astlib`.

### DIFF
--- a/astlib/pprintast.ml
+++ b/astlib/pprintast.ml
@@ -21,13 +21,72 @@
 (* Extensive Rewrite: Hongbo Zhang: University of Pennsylvania *)
 (* TODO more fine-grained precedence pretty-printing *)
 
-open Import
+open Ast_412
 open Asttypes
 open Format
 open Location
 open Longident
 open Parsetree
-open Ast_helper
+
+let varify_type_constructors var_names t =
+  let check_variable vl loc v =
+    if List.mem v vl then
+      Location.raise_errorf ~loc "variable in scope syntax error: %s" v in
+  let var_names = List.map (fun v -> v.txt) var_names in
+  let rec loop t =
+    let desc =
+      match t.ptyp_desc with
+      | Ptyp_any -> Ptyp_any
+      | Ptyp_var x ->
+        check_variable var_names t.ptyp_loc x;
+        Ptyp_var x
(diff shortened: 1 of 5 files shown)


astlib/pprintast.ml
```diff
@@ -21,13 +21,72 @@
...
+  and loop_object_field field =
+    let pof_desc = match field.pof_desc with
+      | Otag(label, t) ->
+        Otag(label, loop t)
+      | Oinherit t ->
+        Oinherit (loop t)
+    in
+    { field with pof_desc; }
+  in
+  loop t
```
pitag-ha, line 88: I've copied this from `Ast_helper`. Another option would be to use the compiler `Ast_helper.varify_type_constructors` by adding it to `Astlib`. But I'm avoiding that,  since we have our own `Ast_helper` inside `ppxlib` and don't need the compiler `Ast_helper` anywhere but here and this will be removed at some point in the future.

astlib/pprintast.ml
```diff
@@ -21,13 +21,72 @@
 (* Extensive Rewrite: Hongbo Zhang: University of Pennsylvania *)
 (* TODO more fine-grained precedence pretty-printing *)
 
-open Import
+open Ast_412
```
pitag-ha, line 24: This will have to be bumped every time we bump the internal AST (and same in the `mli` file). Another option would be to move the information which AST version we're using into the ppxlib `Astlib` and use it both here and in `ast/import`. We would have to move it back, once we drop the ppxlib `Astlib`.

NathanReb (approved):

Issue #268, opened 2021-07-07
Expose Utils.Ast_io.


arozovyk:
Hi, i'm currently working on AST exploration features for VSCode extension: [vscode-ocaml-platform](https://github.com/arozovyk/vscode-ocaml-platform/tree/react-ast-explorer), and i would like to use the `Utils.Ast_io.read` function in order to get the tranformed AST value from `.pp.ml.`  (which i do locally this way: [link](https://github.com/arozovyk/vscode-ocaml-platform/blob/0b5112442ae12cb843d12ad579352e9282a4a579/src/ppx_utils.ml#L26)). 

Would it be possible to expose this module ?

pitag-ha:
Hi @arozovyk, the AST exploration feature for VSCode is super cool, thanks for working on that!

For me it's ok to expose `Utils.Ast_io`. But that module is currently a bit `ppxlib`-internals specific. I'm wondering if we should expose that whole module as is or if it would be enough to expose certain parts that are less specific. From what I understand, you have a binary and need to
1. consume the magic number as in [`read_magic`](https://github.com/ocaml-ppx/ppxlib/blob/master/src/utils.ml#L114-L121). But you already know, unless something is wrong, it will be the impl magic number from the current compiler.
2. consume the `input name` as done [here](https://github.com/ocaml-ppx/ppxlib/blob/master/src/utils.ml#L145). You also don't seem to need the `input name`, at least not currently, but you need to consume what's in the channel before the AST.
3. read the AST and convert it to the `ppxlib`-AST as done [here](https://github.com/ocaml-ppx/ppxlib/blob/master/src/utils.ml#L146-L148)

Is that what you need?

arozovyk:
Hi, @pitag-ha, thanks for replying!  

Your comment made me remember that I wanted to add support for `Inft` and not only `Impl` as I showed [here](https://github.com/arozovyk/vscode-ocaml-platform/blob/0b5112442ae12cb843d12ad579352e9282a4a579/src/ppx_utils.ml#L26).

So basically i will need the whole `ast` field as you can see [here](https://github.com/arozovyk/vscode-ocaml-platform/blob/ppx_tools/src/ppx_utils.ml#L41-L46).  

And i feel like everything that i need (modulo stuff that you mentionned i.e. `input_name`)  is done by [from_channel](https://github.com/ocaml-ppx/ppxlib/blob/master/src/utils.ml#L123-L166)  
So exposing something like : 

```OCaml
[...]

PR #270, opened 2021-07-21
Expose Ast_io read for binary files. 

arozovyk:
This PR is a follow-up on the discussion with @pitag-ha in #268 and is directly linked to [#666](https://github.com/ocamllabs/vscode-ocaml-platform/pull/666). 

It exposes only a specific part of `Utils.Ast_io` that allows reading serialized AST values from binary files.

### DIFF
--- a/src/utils.ml
+++ b/src/utils.ml
@@ -211,7 +211,7 @@ module Ast_io = struct
     | Intf of signature
     | Impl of structure
 
-    let ast_read_error_string (error : read_error) =
+    let read_error_to_string (error : read_error) =
       match error with
       | Not_a_binary_ast ->  "Error: Not a binary ast"
       | Unknown_version (s, _) ->  ("Error: Unknown version " ^ s)
@@ -227,7 +227,7 @@ module Ast_io = struct
       | Ok {ast;_} -> Ok (match ast with
         | Impl structure -> Impl structure
         | Intf signature -> Intf signature )
-      | Error e -> Error (ast_read_error_string e)
+      | Error e -> Error (read_error_to_string e)
 
   end
 end
--- a/src/utils.ml
+++ b/src/utils.ml
@@ -211,16 +211,14 @@ module Ast_io = struct
     | Intf of signature
     | Impl of structure
 
-    type read_error' = Read_error of string
-
     let ast_read_error_string (error : read_error) =
       match error with
-      | Not_a_binary_ast -> Read_error "Error: Not a binary ast"
-      | Unknown_version (s, _) -> Read_error ("Error: Unknown version " ^ s)
+      | Not_a_binary_ast ->  "Error: Not a binary ast"
+      | Unknown_version (s, _) ->  ("Error: Unknown version " ^ s)
       | Source_parse_error (loc, _) ->
-        Read_error ("Source parse error:" ^ Location.Error.message loc)
+         ("Source parse error:" ^ Location.Error.message loc)
       | System_error (loc, _) ->
-        Read_error ("System_error: " ^ Location.Error.message loc)
+         ("System_error: " ^ Location.Error.message loc)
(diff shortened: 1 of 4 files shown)


src/utils.mli
```diff
@@ -53,6 +53,9 @@ module Ast_io : sig
       -> (t, read_error) result
 
     val write : out_channel -> t -> add_ppx_context:bool -> unit
+    module Read_bin : sig 
+      val read : string ->(t, read_error) result
+    end
```
pitag-ha, line 58: Since you only expose `Read_bin` in ppxlib's public API and not its ambient module `Ast_io` (which I think is the right thing to do), the return type of `read` won't form part of the public API. So it won't be possible to handle the return value of `read` (unless folks make use of our private API, which would be good to avoid). So we have to think about what to do with the two types `Ast_io.t` and `Ast_io.read_error` that form part of the return type.

About `Ast_io.t`: What do you think about re-exposing `Ast_io.t` inside `Read_bin` as an abstract type and adding getter functions for `input_name` and for `ast`? For the latter you could choose a good way to express if a signature or a structure is returned, maybe there could be two functions `read_sig` and `read_str` or similar?

About `Ast_io.read_error`: that type is a bit ppxlib internal specific. What do you think about returning a string for an error instead of a `read_error`, in which you describe in words what's gone wrong?

pitag-ha (commented):
Thanks for the PR and sorry for the delay in reviewing it! Apart from the comment I've made, it looks good to me. Could you also sign the commit so that it passes the CI?

src/utils.ml
```diff
@@ -205,6 +205,33 @@ module Ast_io = struct
         output_string oc Input_version.Ast.Config.ast_impl_magic_number;
         output_value oc input_name;
         output_value oc st
+
+    module Read_bin = struct
+    type ast =
+    | Intf of signature
+    | Impl of structure
+
+    type read_error' = Read_error of string
```
pitag-ha: Why do you prefer to define a new type to encapsulate a string instead of using a string directly?

src/utils.ml
```diff
@@ -205,6 +205,33 @@ module Ast_io = struct
...
+
+    let ast_read_error_string (error : read_error) =
```
pitag-ha, line 214: As a side-note: you could call this function `read_error_to_string`. I think `<t1>_to_<t2>` is quite a common pattern to name a type conversion.

By the way, I'm wondering if it might make sense to move that function one module up to `Ast_io`, since that's where `read_error` is defined. What do you think about that?

pitag-ha (commented):
Thanks for the changes! Since the new module is exposed, could you add documentation to it?

I've seen that you've chosen to return the AST directly instead of wrapping it into a record and writing a getter. If we're sure that we don't ever want to extend the function that's good, too. I'm wondering if we might want to extend it at some point, e.g. so that one can also use it to get the input name (which we could already implement now pretty much for free) or information about the cookies. Do you think there's a reason why that won't be of interest?

src/utils.ml
arozovyk: I agree with `read_error_to_string`that does make more sense as a name.

For your second point, I think it is indeed better to keep this function close to the type definition since keeping it in `Read_bin` module doesn't add any value
(and since i dont plan on exposing it because the error message is already passed though ` |  Error e -> Error (ast_read_error_string e)`.

pitag-ha:
By the way, if you could add an entry to CHANGES.md and if you could rebase to get rid of the other commits that ended up in your PR, that would be great!

pitag-ha (approved):
Thanks! To me it looks good.

@NathanReb , could you also have a quick look?

NathanReb:
It looks good to me!

NathanReb (approved):

PR #271, opened 2021-07-23
Add support for registering ppx_import transformation as context-free rule

NathanReb:
This PR adds support for declaring ppx_import special syntax as a context-free (in the ppxlib sense) rule.

It allows registering a `type_declaration -> type_declaration` transformation rule (let's name it `ext`) that is applied whenever we find a type declaration such as:
```ocaml
type t = [%ext: <some core_type>]
```

This transformation is applied during the same pass as other context-free rules.

I had the following things in mind while working on this feature, in order of priority:
1. The feature must work and allow porting ppx_import to ppxlib without changing its behaviour
2. It must be implemented in a way that will allow factoring as much code between this syntax and the new syntax proposed for `ppx_import` that will simply use a regular structure_item or signature_item extension rewriter.
3. It must remain as hidden as possible to other users, this is a temporary solution to allow a transition period towards the new syntax. IMHO, we do not wish to expose this as a regular ppxlib feature as it needlessly complexify how we traverse the AST and apply context-free transformations.

I'm opening this as a draft as there are a few things I wish to improve and test before we can merge. Any feedback is still appreciated at this stage.

### How to implement ppx_import on top of this

To use this feature you'll have to register the ppx_import type_declaration transformation using the undocumented `Extension.__declare_ppx_import`:
```ocaml
val __declare_ppx_import :
  string ->
  (ctxt: Expansion_context.Extension.t -> type_declaration -> type_declaration) ->
  t
```
instead of the regular `Extension.V3.declare`. If you're not familiar with this part of ppxlib's API yet you can read the following [section of the manual](https://ppxlib.readthedocs.io/en/latest/ppx-for-plugin-authors.html#writing-an-extension-rewriter) or take a look at the following [example](https://github.com/ocaml-ppx/ppxlib/tree/master/examples/simple-extension-rewriter).

This will work both in signatures and structures.

The expand function that you have to provide (the `ctxt: Expansion_context.Extension.t -> type_declaration -> type_declaration` part) will receive the individual type declaration surrounding the extension point.

Note that this feature is only required for type declaration and that you can already express the module type transformation as its current syntax is already context-free. You can do so using:
```ocaml
Extension.V3.declare
  "import"
  Extension.Context.module_type
  Ast_pattern.(ptyp (ptyp_package __))
  (fun ~ctxt (module_longident, with_type_list) -> ...)
```

### Preserving attributes

It's important that the expand function preserves the attributes, particularly the `[@@deriving ...]` ones from its argument. It's most likely already the case with the current implementation anyway but it's worth pointing out that ppxlib isn't taking care of this here and certaintly won't reasonably be able to do it for the new syntax.

### Pluging in the new syntax

Using this won't prevent you from also adding support for the new syntax subsequently, using the regular `Extension.V3.declare` with `Context.signature_item` and `Context.structure_item`.
The implementation will be slightly different as you will have to interpret the payload yourself to eventually narrow it down to a `type_declaration`.
Another difference to note is that for the new syntax you will have to declare one extension for `signature_item` and one for `structure_item`.

## TODO

- [x] Provide the correct `Expansion_context`, at the moment it doesn't point to the actual extension point in the source
- [x] Handle module types since ppx_import currently supports them, ~this will go through a separate `declare` function. I'll likely rename the existing one in the process to make the context explicit~ (there's nothing to do here, this part of ppx_import actually is context free!
- [x] Test composition with str/sig_item rewriters
- [x] Investigate mutually recursive type declarations and how they are handled in `ppx_import`. I don't expect it to cause troubles here but it might be a problem for the new syntax.
- [x] Specify and test the behaviour for name collisions with regular `core_type` extensions

### DIFF
--- a/src/extension.ml
+++ b/src/extension.ml
@@ -157,7 +157,23 @@ struct
     with_arg : bool;
   }
 
-  let declare ~with_arg name context pattern k =
+  let declare :
+      type a.
+      with_arg:bool ->
+      string ->
+      a Context.t ->
+      (payload, 'b, 'payload) Ast_pattern.t ->
+      'b Callback.t ->
+      (a, 'payload) t =
+   fun ~with_arg name context pattern k ->
+    (* Check that there is no collisions between ppx_import and core_type
+       extensions *)
+    (match context with
+    | Context.Ppx_import ->
+        Name.Registrar.check_collisions registrar (Context.T Core_type) name
+    | Context.Core_type ->
+        Name.Registrar.check_collisions registrar (Context.T Ppx_import) name
+    | _ -> ());
     Name.Registrar.register ~kind:`Extension registrar (Context.T context) name;
     {
       name = Name.Pattern.make name;
--- a/src/extension.ml
+++ b/src/extension.ml
@@ -54,7 +54,7 @@ module Context = struct
     | Pattern -> "pattern"
     | Signature_item -> "signature item"
     | Structure_item -> "structure item"
-    | Ppx_import -> "ppx_import"
+    | Ppx_import -> "type declaration"
 
   let eq : type a b. a t -> b t -> (a, b) equality =
    fun a b ->
--- a/src/extension.ml
+++ b/src/extension.ml
@@ -77,27 +77,12 @@ module Context = struct
 
   let get_ppx_import_extension type_decl =
     match type_decl with
-    | {
-     ptype_manifest =
-       Some { ptyp_desc = Ptyp_extension (name, PTyp core_type); _ };
-     _;
-    } ->
-        (* If the type_decl follows the syntax expected by ppx_import, we pass the expander
-           the type_declaration with the extension around the type manifest removed.
-           [type declaration] can't be used as payload directly so we wrap it in a dummy str_item
-           from which the Ast_pattern will extract the type_declaration.
-           This str_item will not be seen by the user. *)
-        let expander_type_decl =
-          { type_decl with ptype_manifest = Some core_type }
-        in
+    | { ptype_manifest = Some { ptyp_desc = Ptyp_extension (name, _); _ }; _ }
+      ->
         let virtual_payload =
           Ast_builder.Default.pstr_type ~loc:type_decl.ptype_loc Recursive
-            [ expander_type_decl ]
+            [ type_decl ]
         in
-        (* TODO: Do we need to extract attributes to be later merged here?
-            E.g. in the case [type t = [%import A.t][@attr ...], should attr be re-attached
-            to the generated type manifest, and if so what should we do if said manifest is
-            None, even though ppx_import should never generate such a manifest. *)
         let attr = [] in
         Some ((name, PStr [ virtual_payload ]), attr)
     | _ -> None
@@ -148,7 +133,6 @@ module Context = struct
         assert_no_attributes attrs;
         x
     | Ppx_import ->
-        (* TODO: bis attrs *)
         assert_no_attributes attrs;
         x
 end
@@ -387,8 +371,8 @@ let declare_inline_with_path_arg name context pattern k =
   T (M.declare ~with_arg:true name context pattern k')
 
 let __declare_ppx_import name expand =
-  (* This pattern is used to unwrap the type declaration from the payload assembled by
-     [Context.get_ppx_import_extension] *)
+  (* This pattern is used to unwrap the type declaration from the payload
+     assembled by [Context.get_ppx_import_extension] *)
   let pattern = Ast_pattern.(pstr (pstr_type recursive (__ ^:: nil) ^:: nil)) in
   V3.declare name Context.Ppx_import pattern expand
 
--- a/src/extension.ml
+++ b/src/extension.ml
@@ -16,6 +16,7 @@ module Context = struct
     | Pattern : pattern t
     | Signature_item : signature_item t
     | Structure_item : structure_item t
+    | Ppx_import : type_declaration t
 
   type packed = T : _ t -> packed
 
@@ -53,6 +54,7 @@ module Context = struct
     | Pattern -> "pattern"
     | Signature_item -> "signature item"
     | Structure_item -> "structure item"
+    | Ppx_import -> "ppx_import"
 
   let eq : type a b. a t -> b t -> (a, b) equality =
    fun a b ->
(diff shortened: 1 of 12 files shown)


NathanReb:
I've been looking into this a bit further, specifically working on the support for module type declaration.

The current syntax in ppx_import for module types doesn't easily allow handling it the same way I originally handled the support for type declarations. As a very quick reminder, the current handling of the `type t = [%import: ...]` is that ppxlib only recognize this as a ppx_import-like extension if the payload is a `core_type`, in which case the payload and its context have the same type and the extension point wrapper can be "removed" before passing the type declaration to the expander.
It's harder to do for module type declaration such as `module type M = [%import: (module A)]` because here the payload and the context differ: we need a `module_type` and the payload is `core_type`. We'd need to transform the payload which is doable) but that means we need to deal with the cases where the users wrote a wrong payload that we can't map to a module type. This can cause a few problem which I'm happy to explain in detail if you'd like.

Thinking about all this made me reconsider my original choice. I propose a simpler approach in ppxlib:

Whenever ppxlib encounters one of the following:
```ocaml
type t = [%ext ...]
```
or
```ocaml
module type M = [%ext ...]
```
it will look for a ppx_import-like extension, registered using `__declare_ppx_import_td` or `__declare_ppx_import_mtd`, named `ext` and call its expander on the entire type or module type declaration containing the extension point, without applying any transformation.
This means that the new and old syntax in ppx_import will share slightly less code but it also means that the ppxlib port should keep the diff relatively small.

The problem with the previous approach, which was to pass the expander a payload tailored to look like the payload with the future syntax, is that ppxlib embeds a lot of `ppx_import` logic, expects a very specific form of payload and becomes responsible
for rejecting ill-formed payloads as I explianed above. It is doable in practice but requires slightly more involved changes to be done properly. I'm not sure we want to do this, especially since we agreed this is a temporary solution until we move to the new syntax (even though this is relatively long term).

I also had some time to think about what the new syntax would look like for module type declarations. It's not as straightforward of a transformation but I think it's an overall improvement.
As of today, one would use `ppx_import` like this:
```ocaml
module type M = [%import: (module Hashtbl.HashedType with type t = my_type)]
```
The payload here is a `core_type` and in particular it's what the AST describes as a `package_type`. It is fairly restrictive compared to what on can write for actual module type declaration. The syntax I'm proposing to replace this is the following:
```ocaml
module type%import M = Hashtbl.HashedType with type t = my_type
```
Here the user is writing an actual module type declaration.

In terms of AST, the old syntax looks like this:
```ocaml
{pstr_desc =
   Pstr_modtype
    {pmtd_name = {txt = "M"};
     pmtd_type =
      Some
       {pmty_desc =
         Pmty_extension
          ({txt = "import"},
           PTyp
            {ptyp_desc =
              Ptyp_package
               ({txt = Ldot (Lident "Hashtbl", "HashedType")},
                [({txt = Lident "t"},
                  {ptyp_desc = Ptyp_constr ({txt = Lident "my_type"}, []);
                   ptyp_loc_stack = []})]);
             ptyp_loc_stack = [...]})}}}
```
and the new one like this:
```ocaml
{pstr_desc =
   Pstr_extension
    (({txt = "import"},
      PStr
       [{pstr_desc =
          Pstr_modtype
           {pmtd_name = {txt = "M"};
            pmtd_type =
             Some
              {pmty_desc =
                Pmty_with
                 ({pmty_desc =
                    Pmty_ident {txt = Ldot (Lident "Hashtbl", "HashedType")}},
                 [Pwith_type ({txt = Lident "t"},
                   {ptype_name = {txt = "t"}; ptype_params = [];
                    ptype_cstrs = []; ptype_kind = Ptype_abstract;
                    ptype_private = Public;
                    ptype_manifest =
                     Some
                      {ptyp_desc = Ptyp_constr ({txt = Lident "my_type"}, []);
                       ptyp_loc_stack = []}})])}}}]),
    ...)}
```

One of the possible advantages of the new syntax is that it allows the full range of features supported by the `with_constraint`, that is `with module X = ...`, and eventually the subst versions (`with type t := ...` or `with module := ...`). This could potentially be interpreted by ppx_import.

@pitag-ha @ejgallego @tatchi let me know what you think! If what I'm proposing to add to ppxlib in order to finish the port of the current syntax, I'll update the PR message, especially the part about migrating to ppxlib and then to the new syntax and push the changes shortly!

NathanReb:
Quick update on module types: I just looked at the code of ppx_import and it appears that the module type extension is actually already context free so it can be ported using regular extensions, no need for extra support there. That leaves us with just the type declarations!

NathanReb:
I also looked into how ppx_import handles mutually recursive type definitions and it seems that it doesn't do anything special about them, that is if we have a library `A` that defines two mutually recursive types:
```ocaml
type a =
   | Ai of int
   | Ab of b
and type b =
   | Bs of string
   | Ba of a
```
If you want to import them, you have to import them both or use `[@with ...]` attributes, otherwise you would get errors. To do this with the current syntax you have to write:
```ocaml
type a = [%import: A.a]
and type b = [%import: A.b]
```
With the new syntax you would have to write:
```ocaml
type%import a = A.a
and type b = A.b
```
which is equivalent to:
```ocaml
[%%import
  type a = A.a
  and type b = A.b]
```
The only thing that you can do with the old syntax but won't be able to do with the new one is this:
```ocaml
type a = [%import: A.a]and type = A.b]
```
The only thing that you can do with the old syntax but won't be able to do with the new one is this:
```ocaml
type a = [%import: A.a]
and type b = [%import: A.b]
and type c = <some type definition>
```

The thing is I can't think of a reason why one would want to do that, especially given that they can always write:
```ocaml
type a = [%import: A.a]
and type b = [%import: A.b]

type c = <some type definition>
```

I think we can move forward with the rest then. @pitag-ha pointed out that we have to check that there is no name conflict between `core_type` extensions and ppx_import-like extensions or to at least raise the proper error if we can't determine which one to use.

NathanReb:
I just pushed a commit getting rid of the part that was transforming the type declaration before passing it to the expander. It now receives it untouched, including the `[%ext ...]` from the manifest.

Since the extension point is present in the expander argument, it's probably not worth trying to make a custom Expansion_context pointing to it, the authors of ppx_import will have all the information they need if they want to use extension point loc rather than the type decl one.

NathanReb:
I just updated the main PR message to reflect the recent changes.

All that remains now is more robust testing of corner cases and properly handling name collisions between ppx_import-like and core type extensions.

test/ppx_import_support/test.ml
```diff
@@ -0,0 +1,102 @@
...
+[%%expect{|
+- : unit = ()
+|}]
+
+(* The expander receives the type decl with the extension point removed, it should preserve
+   attibutes *)
+type t = [%id: int]
+[%%expect{|
+type t = int
+|}]
```
pitag-ha, line 29: Do you test here if it preserves attributes? Is there a way to test that with an expect test?

test/ppx_import_support/test.ml
```diff
@@ -0,0 +1,102 @@
...
+val foo : Deriving.t = <abstr>
+|}]
+
+(* It should properly compose with [@@deriving] *)
+type t = [%id: int]
+[@@deriving foo]
+[%%expect{|
+type t = int
+val foo : t = 42
+|}]
```
pitag-ha, line 55: How does this test show that `[%id]` composes well with `[@@deriving]`, i.e. that it's applied before? Does it show that by inferring the type of `foo` as `t`, as opposed to `int`?

pitag-ha (commented):
Nice, thanks!

>I just pushed a commit getting rid of the part that was transforming the type declaration before passing it to the expander. It now receives it untouched, including the [%ext ...] from the manifest.

Why is it better that way? It seemed like your old approach met aim 2. of you PR description quite well.

test/ppx_import_support/test.ml
NathanReb: You're right that the test isn't precise enough with regards to the ordering, I'll rewrite the deriver to check that the manifest isn't an extension point!

test/ppx_import_support/test.ml
NathanReb: The `@@deriving test below should be enough for that!

test/ppx_import_support/test.ml
NathanReb: That was a very good catch! It turns out that it is indeed applied AFTER the deriving. I completely misunderstood the traversal here, I'll work on fixing this!

src/extension.ml
```diff
@@ -57,8 +59,19 @@ module Context = struct
...
+    | Ppx_import       , Ppx_import       -> Eq
     | _ -> assert (Poly.(<>) (T a) (T b)); Ne
 
+  let get_ppx_import_extension type_decl =
+    match type_decl with
+    | {ptype_manifest = Some {ptyp_desc = Ptyp_extension (name, _); _}; _} ->
+      let virtual_payload =
+        Ast_builder.Default.pstr_type ~loc:type_decl.ptype_loc Recursive [type_decl]
+      in
+      let attr = [] in
```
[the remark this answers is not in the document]
NathanReb: They're not captured because they will be part of the payload here, it's ppx_import responsibility to handle and preserve them.

NathanReb:
I rebased on top of #279 to fix the interactions with @@deriving so it needs to be merged after #279!

test/ppx_import_support/test.ml
pitag-ha: Oh, I see. So the comment was a little bit misleading to me.

pitag-ha (approved):
Thanks! LGTM.

PR #277, opened 2021-09-09
Make ppxlib conflict with +effects variants

kit-ty-kate:
Backports https://github.com/ocaml/opam-repository/commit/5c15b4fb9020ea0a8b690d7dca7e6dfe25a86a59


pitag-ha:
Thanks @kit-ty-kate! And sorry for forgetting to upstream this myself.

About the changelog entry: I'm thinking if to add a normal changelog entry or if to add a changelog entry under the 0.23.0 section and to also modify the 0.23.0 github release notes. The former would reflect the 0.23.0 github release and the latter would reflect the 0.23.0 opam release (in which you already added this constraint). Do you as a release specialist have an opinion on which option would be better?

kit-ty-kate:
I'm not sure it's worth adding a changelog for that (at least not a public one) given this just prevents the compilation on an unreleased compiler variant that would've made ppxlib fail building anyway and previous versions also carried the same issue.

kit-ty-kate:
But if you really want to add a changelog entry, I would just add a normal one and not worry about previous releases. It would be more in line with the actual release tarball (opam-repository being something separate, with separate maintenance rules)

pitag-ha:
> I'm not sure it's worth adding a changelog for that (at least not a public one)

Ok, sounds good. Thanks!

PR #278, opened 2021-09-10
Disable expect tests on 4.13 compiler

pitag-ha:
This commit just disables the expect tests on 4.13 for now. Soon, when 4.13 is official, we'll need to decide if to do it the other way around (only enable the expect tests on 4.13 and not on older compilers) or if to add some cross-compiler logic to our local expect-test binary.

### DIFF
--- a/test/base/dune
+++ b/test/base/dune
@@ -1,7 +1,9 @@
 (rule
  (alias runtest)
  (enabled_if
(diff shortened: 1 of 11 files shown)


pitag-ha:
I'm closing this since it's extremely simple to make the tests work on both 4.13 and older compilers.

PR #279, opened 2021-09-10
Expand structure and signature items before applying attribute rules

NathanReb:
This PR changes the order in which context-free rules are applied.

It is a follow up to #276 and "fixes" the behaviour observed there. With this PR a node with attributes attached will be first expanded and then the inline attribute rules applied instead of the other way around.

Imagine the following:
```ocaml
type t = [%some_ext ...]
[@@deriving some_deriver]
```
Before this PR, the `some_deriver` transformation would be applied first, now it is applied to the result of the expansion of the type declaration, meaning after the `some_ext` transformation was applied.

@kit-ty-kate is gracefully testing this on the opam universe to see if it has any serious impact. I don't expect anyone to rely on this behaviour but better be safe than sorry.

Edit: You might want to review this while hiding whitespace diffs as a whole block in `Context_free` has gained one level of identation without any further change.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 unreleased
 ----------
 
+- Expand nodes before applying derivers or other inline attributes based
+  transformation, allowing better interactions between extensions and
+  derivers (#279, @NathanReb)
+
 0.23.0 (31/08/2021)
 -------------------
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -545,3 +545,7 @@ generates new structure or signature items based on the value of an item
 with that attribute attached. This rule is applied before the item has been
 transformed meaning the input of the expander function might contain "wrong"
 information.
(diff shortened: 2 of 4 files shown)


kit-ty-kate (approved):
The check in opam-repository I did, didn't reveal any packages broken by this

NathanReb:
I updated the internal documentation and the tests description to reflect this change. I think this should be ready for review now.

@jeremiedimino @ceastlund, I don't really expect anything to break here but it might be good to test this change inside JS' monorepo. What do you think?

NathanReb:
Rebased on top of (formatted!) main and fixed the bug found by @pitag-ha where I forgot to remove a `self#structure_item` call in `Context_free.map_top_down#structure` resulting in some structure item being traversed twice.

pitag-ha (approved):
Looks good also to me. Thanks!

PR #280, opened 2021-09-10
Make our expect-test version compatible with 4.13 compiler

pitag-ha:
The Warnings.parse_options function was returning a unit on older compilers and is returning an alert option now, but under the hood it still has the same side-effects.


NathanReb (approved):
Looks good! Much better to be able to test on more versions than having to chose between
`< 4.13` or `>= 4.13`!

PR #281, opened 2021-09-10
Use ocamlformat

pitag-ha:
Currently our expect-test lexer is too strict for our expect tests to work well with ocamlformat, so they're excluded from the formatting for now. The lexer could probably be improved easily to include them in the formatting.

Since files that are preprocessed with `pp_rewrite` contain invalid syntax, they must be excluded from ocamlformat for sure.

For files that use cinaps to generate bode blocks from other code blocks to work well with ocamlformat (as e.g. in [code_matcher](https://github.com/ocaml-ppx/ppxlib/blob/main/src/code_matcher.ml#L143-L171)), the closing `(*$*)` has to be separated from the last code block by an empty line. When that's the case it seems to work well.

In files that inject freely formatted code via cinaps, we'd have to rewrite the injection strings to respect the current ocamlformat formatting to make it work. Staying compatible with ocamlformat would then require work, so instead those files are also excluded from ocamlformat.

The second commit ensures that if you run
```
git config blame.ignoreRevsFile .git-blame-ignore-revs
```
once, then this bulk change won't mess up your `git blame`.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# The bulk change commit enabling ocamlformat
+427f96e126d306538eb541ac591f71b2c68e5dd4
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,0 +1,3 @@
+version=0.19.0
+profile=conventional
+parse-docstrings=true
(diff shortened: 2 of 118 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# The bulk change commit enabling ocamlformat
+add3a6ddc9d3e6cf6bda1f8fd753740afe9c807d
```
NathanReb, line 2: That's extremely neat!

NathanReb (approved):
Thanks, this is a long awaited improvement!

PR #283, opened 2021-09-13
Remove Import_for_core

pitag-ha:
Import_for_core was meant to serve ppx_core which was kept for a while after ppxlib was made for backwards compatibility. By now, ppx_core is deprecated and Import_for_core doesn't even form part of ppxlib's public API anymore, so it can simply be fully removed.

### DIFF
--- a/ast/ppxlib_ast.ml
+++ b/ast/ppxlib_ast.ml
@@ -17,5 +17,3 @@ module Parsetree      = Parsetree
 module Pprintast      = Astlib.Pprintast
 module Select_ast     = Select_ast
 module Selected_ast   = Selected_ast
-
-module Import_for_core = Import


NathanReb (approved):
Looks good to me!

For future reference, which package was breaking because of this?

pitag-ha:
It was used in GT. So GT has needed a ppxlib upper bound already for our last release 0.23.0, in which we removed `Import_for_core` from our public API. It seems to me like they used it just to [access the simple conversion module](https://github.com/JetBrains-Research/GT/blob/master/camlp5/core2.ml#L17), which should be done via `Ppxlib_ast.Selected_ast`. So [I opened a PR there](https://github.com/JetBrains-Research/GT/pull/24) to point that out.

PR #284, opened 2021-09-20
add input_name to the Expansion_context.Extension module

tatchi:
That will be needed for https://github.com/ocaml-ppx/ppx_import/pull/54

The new `Extension.__declare_ppx_import` takes an `Expansion_context.Extension.t` : https://github.com/ocaml-ppx/ppxlib/blob/main/src/extension.mli#L197

Whereas the current `ppx_import` port to ppxlib uses a `Ppxlib.Expansion_context.Base` : https://github.com/tatchi/ppx_import/blob/cfd93d7b3034a99e85133a0eba261ca77e0281ea/src/ppx_import.ml#L531

The problem is that the `Extension` module doesn't expose an `input_name` function (like the `Base` module does).

### DIFF
--- a/src/expansion_context.mli
+++ b/src/expansion_context.mli
@@ -51,10 +51,24 @@ module Extension : sig
   (** Return the location of the extension point being expanded *)
 
   val code_path : t -> Code_path.t
-  (** Return the code path for the given context *)
+  (** Return the code path for the given context In Driver, Deriving and
+      Extension, the context is initialized so that the [file_path] component of
+      the [code_path] is determined from the first location found in the input
+      AST. That means that:
+
+      - It's the empty string in empty structures or signatures
+      - It can be altered by line directives *)
 
   val input_name : t -> string
-  (** Return the input name for the given context *)
+  (** Return the input name for the given context. In Driver, Deriving and
+      Extension, the context argument is initialized so that the [input_name]
+      matches the input filename passed to the driver on the command line. That
+      means that:
+
+      - It has a value even for empty files
+      - It is not affected by line directives
+      - It is ["_none_"] when using [Driver.map_structure] or
+        [Driver.map_signature] *)
 
   val tool_name : t -> string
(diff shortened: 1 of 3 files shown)


src/expansion_context.mli
```diff
@@ -53,6 +53,9 @@ module Extension : sig
   val code_path : t -> Code_path.t
   (** Return the code path for the given context *)
 
+  val input_name : t -> string
```
pitag-ha, line 56: What do you think about using the same documentation as in `Base.input_name`? And also for `code_path` using the same as in `Base.code_path`? The long documentation for those two functions in `Base` is to make clear what the difference between `input_name` and `code_path.file_path` is.

pitag-ha (commented):
Thanks for the PR! Yes, adding `input_name` also to other expansion contexts than `Base` is good for sure, both for the `ppx_import` port and in general. What do you think about also adding it to `Expansion_context.Deriver`?

src/expansion_context.mli
tatchi: Done, I literally copied the documentation (I hope that's what you had in mind)

tatchi:
> Thanks for the PR! Yes, adding `input_name` also to other expansion contexts than `Base` is good for sure, both for the `ppx_import` port and in general. What do you think about also adding it to `Expansion_context.Deriver`?

That definitely makes sense to add it also to the `Expansion_context.Deriver`. I should have added it in the first place, I went too fast to open the PR 😅

Thanks for the review!  😄

tatchi:
Let me know if I should squash my commits (I assumed you can do it through Github)

pitag-ha (approved):
Thanks!

> I went too fast to open the PR :sweat_smile:

Same as me last time when I only added it to `Base` :sweat_smile:

> Let me know if I should squash my commits (I assumed you can do it through Github)

Yes, I can do it through Github.

PR #285, opened 2021-09-23
Gensym improvements

ceastlund:
Improved two aspects of `gen_symbol`:
1. passing one `gen_symbol` result as `~prefix` to another no longer results in a double-suffix
2. `name_type_params_in_td` produces more sensible variable names
First added tests for both behaviors, then improved them and updated tests.


pitag-ha (approved):
Thanks, LGTM!

I'm not very familiar with `name_type_params_in_td`. Why is it better to use `a`, `b`, ... as a prefix for the name instead of `v_x` or no prefix at all?

ceastlund:
The `name_type_params_in_td` function assigns names to type parameters declared as underscores, because having names for them is often handy during ppx expansion. Let's say we start with:
```type (_, _) t = int```
Currently, `name_type_params_in_td` produces something like:
```type ('v_x__097_, 'v_x__098_) t = int```
The code is written using `"v" ^ gen_symbol ()`, and `gen_symbol` has a `?prefix` argument that defaults to `"_x"`. I kind of suspect the code meant to use `"v"` as a prefix, and the author didn't think about the `"_x"` default. I suspect the intention was:
```type ('v__097_, 'v__098_) t = int```
There's no real reason why `"_x"` or `"v_x"` are at all meaningful or suggestive names for types. So I wanted to change that code to use `gen_symbol`'s `~prefix` argument.

While I was at it, it occurred to me that `"v"` is also not a great prefix for a type. Usually, `v` is for "value". We usually use the beginning of the alphabet, like:
```type ('a, 'b) t = int```
So I made the code do something similar:
```type ('a__097_, 'b__098_) t = int```

pitag-ha:
Oh nice, thanks for all the explanation! Using `a...`, `b...`, ... for type parameter names does indeed sounds more intuitive than `v_...` or `v_x...`. As a side-note: it's `a...`, `b...`, ..., not `'a...`, `'b...`, ..., right?

ceastlund:
What gen_symbol generates has no apostrophe. But when wrapped in Ptyp_var and rendered back as ocaml syntax, there would be an apostrophe there.

PR #286, opened 2021-09-24
add _opam to .gitignore

tatchi:
Useful when you use a local switch 😄


pitag-ha (approved):
That's true!

I have it in my global gitignore, so I never noticed.

Issue #291, opened 2021-10-11
Improving handling of exceptions in context-free rewriters


panglesd:
A rewriter has several ways to report errors to the user. The best is to embed an error extension point in the generated AST, but the PPX can also either raise a located error exception, or any exception.

Currently, the `ppxlib` driver catches any exception raised by a rewriter, and:
- if the exception is a located error and
  - either it is used as a ppx,
  - or it is used as a standalone with the `-embed-errors` argument,

then it replaces *the whole AST* by a single located error extension node  with a message,
- In any other case an exception is raised again.

Replacing the whole AST is not satisfying as it prevents any other error to be shown, even in the case of located errors. Only the node that generated the exception should be replaced by an error extension node.

Moreover, what do you think about turning any exception raised by a context-free rewriter into located error (?)

PR #292, opened 2021-10-11
Improving error message when rewriters throw exceptions

panglesd:
A rewriter has several ways to report errors to the user. The best is to embed an error extension point in the generated AST, but the PPX can also either raise a located error exception, or any exception.

Currently, the `ppxlib` driver catches any exception raised by a rewriter, and:
- if the exception is a located error and
  - either it is used as a ppx,
  - or it is used as a standalone with the `-embed-errors` argument,
 
then it replaces the whole AST by a single located error extension node  with a message,
- In any other case the caught exception is raised again.

In the case of an unlocated exception, it is hard for the user to know that the ppx is responsible for the error.

This PR mostly does not modify the above handling of exceptions, but enhance the error message as follows:
- if the caught exception is a located error, it adds the name of the ppx which raised the error as a prefix to the error message. For instance, if ppx `example_ppx` raised the located error with message `forbidden payload`, then the new message is `(example_ppx) forbidden payload`
- if the caught exception is not a located error, instead of raising the same exception, it raises a `Failure` with an error message explaining what happened, including: the name of the responsible ppx, the exception itself, and an invitation to report the fact that located exceptions or extensions error nodes should be used instead of regular exceptions to the ppx owners.

The error message looks like this:

```
  Fatal error: exception The following exception was raised during the execution of a context free rewriter:
  (Failure "A raised exception")
  The exception does not contain any information about the location of the error. If you know the responsible ppx, you might want to file an issue to the ppx authors.
```

Note that in the case of derivers or extenders, the exception is caught after the name of the ppx is forgotten. The name of the ppx is thus always `<builtin:context-free>` in the error message. This will be improved by solving issue #291.


test/extensions_and_deriving/test.ml
```diff
@@ -51,7 +51,7 @@ val gen_type_decl : Context_free.Rule.t = <abstr>
 
 [%%expect{|
 Line _, characters 3-19:
-Error: Attributes not allowed here
+Error: (ppx <builtin:context-free>) Attributes not allowed here
```
pitag-ha, line 54: This error is thrown by ppxlib. So in this case, pointing to a PPX (or saying the error was thrown while applying a PPX) might be confusing, since that suggests that the error is somehow related to the PPX, whereas, in fact, it's simply the PPX user who is writing an attribute at a place where the parsetree doesn't expect attributes.

I'm thinking about how we could fix this. It's just a manifestation of the following problem: all located errors raised by ppxlib when traversing the AST to apply all context-free rules (i.e. everything that happens in `Context_free.map_top_down`) are caught just in the same way as the located errors raised by PPXs are caught. I can see two options to approach this:
1. We could simply not improve the error message in the case of located errors inside context-free rules. Once we've approached your issue #291 on handling that kind of errors correctly in ppxlib, that shouldn't be a problem anymore anyways. That might be the most reasonable solution.
2. You could have a look into how complicated it would be to divide the logic of catching exceptions into those coming from whole-file transformations (catch them in `driver.ml` just as you do now) and those coming from rules (catch them in `Context_free.map_top_down`.

What do you think about this? And how do the two options, particularly option 1, sound to you?

pitag-ha (commented):
Nice, thanks for this non-trivial improvement!

I have mainly some minor comments; only one of them is conceptual. Let me know what you think about it.

One detail about one of the cases: we definitely want to discourage writers of whole-file transformations to raise located errors. So maybe, in that case, you could also add some sentence to the error message along the lines "The located error was thrown as an exception instead of embedding it as an extension node into the AST. So merlin won't be able to report any other error in this file anymore. You might want to file an issue to the ppx authors." The way of improving the message in the other cases is great!

And one last thing: could you add a changelog entry (an entry to CHANGES.md), please? Once you do the CI will pass.

src/driver.ml
```diff
@@ -496,6 +496,36 @@ let get_whole_ast_passes ~hook ~expect_mismatch_handler ~tool_name ~input_name =
...
+                  owner_name (Printexc.to_string exn'))
+      | _ -> None)
+
+let process_exn exn owner_name =
+  match Location.Error.of_exn exn with
+  | None -> ExceptionFromPPX (owner_name, exn)
+  | Some error ->
+      Location.Error
+        (Location.Error.set_message error
+           ("(ppx " ^ owner_name ^ ") " ^ Location.Error.message error))
```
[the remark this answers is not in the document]
panglesd: I went for `(context-free ppx pass) ` as an exception can occur without being related to a specific ppx. So the only information we can give is that it happened during the phase of the driver where it applies (in one pass) the context-free ppxes. When #291 is done, this will precisely be exceptions not thrown by a specific ppx but by the driver.

src/driver.ml
```diff
@@ -95,6 +95,14 @@ module Instrument = struct
 end
 
 module Transform = struct
+  type kind =
+    | WholeFileTransform
+    | BuiltinContextFree
+    | BuiltinLinter
+    | BuiltinPreprocess
+    | BuiltinBeforeInstr
+    | BuiltinAfterInstr
```
pitag-ha, line 104: I would say that whole-file transformations are all transformations that transform the whole AST (as opposed to context-free rules, that just extend one node), so that includes instrumentation, linters and preprocess. Since you have some kinds of whole-file-transformations separately, it might be a bit confusing to have one variant called `WholeFileTransform`. Is it possible to separate it into `Impl` (whole-implementation-file-transformations) and `Intf` (whole-interface-file-transformations)?

src/driver.ml
```diff
@@ -169,6 +178,7 @@ module Transform = struct
         lint_intf;
         instrument;
         registered_at = caller_id;
+        kind = WholeFileTransform;
```
pitag-ha, line 181: What do you mean with `WholeFileTransform` here? It is also possible to register just one context-free rule.

test/driver/exception_handling/run.t
```diff
@@ -77,47 +79,52 @@ can be reported about it.
...
+  Fatal error: exception (context-free ppx pass) (Failure "A raised exception")
   [2]
 
  In the case of derivers
 
   $ echo "type a = int" > impl.ml
   $ echo "type b = int [@@deriving deriver_raised_exception]" >> impl.ml
   $ ./deriver.exe -embed-errors impl.ml
-  Fatal error: exception (Failure "A raised exception")
+  Fatal error: exception (context-free ppx pass) (Failure "A raised exception")
```
pitag-ha, line 116: Here it would also be nice to have the suffix "The use of ... is discouraged...", wouldn't it?

test/driver/exception_handling/run.t
```diff
@@ -77,47 +79,52 @@ can be reported about it.
...
 
  In the case of whole file transformations:
 
   $ echo "let x = 1+1. " > impl.ml
   $ ./whole_file_located_error.exe -embed-errors impl.ml
-  [%%ocaml.error "A located error in a whole file transform"]
+  [%%ocaml.error
+    "The following located exception was raised during the whole file transformation phase of the ppx \"raise_exc\":\nA located error in a whole file transform\nThe use of located exception is discouraged in whole file transformation as it prevents other errors to be reported. Instead, errors should be embedded in the AST in extension nodes. You might want to file an issue to the ppx authors."]
 
 3. Raising an exception. The exception is not caught by the driver.
```
pitag-ha, line 99: You could update this test now by saying that it's caught and it's re-raised with an improved message.

src/driver.ml
```diff
@@ -496,6 +496,36 @@ let get_whole_ast_passes ~hook ~expect_mismatch_handler ~tool_name ~input_name =
...
 
+exception ExceptionFromPPX of (string * exn)
+
+let () =
+  Printexc.register_printer (fun exn ->
+      match exn with
+      | ExceptionFromPPX (owner_name, exn') ->
+          Some
+            (match owner_name with
+            | "<builtin:context-free>" ->
```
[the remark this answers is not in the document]
pitag-ha: Sounds good. That way it's also easier to have a better overview over what which transform does. Thanks!

test/driver/exception_handling/run.t
panglesd: How many times will I forget to update the text when the tests change behavior? :sweat_smile:

panglesd:
> > I also wonder maybe more tests should be added? There is currently no test on linters, instrumentations etc...
> 
> That sounds great! Do you also want to do that?

Sure! Also, maybe some tests for extension points in structures, patterns, ... and not only expressions.

It's a bit of tedious work, but I would not be comfident committing code that I have never even tried once :wink:

src/driver.ml
pitag-ha: > Indeed, the current type Transform.t is not ensuring any safety: there can be transforms of kind Linters that contain no linter but an impl...

:heavy_plus_sign: 

> think this leads to errors, and I think it actually does in this particular case: when registering an impl together with an instrumentation, the impl function is never called... (I'll come back with more details on this bug!)

True. Good catch!

test/driver/exception_handling/run.t
pitag-ha: > so is it really discouraged?

That's a good question. So on one hand, you're right that from merlin's point of view throwing any exception inside a context-free rule is not a big deal if we catch it and embed it by locating it around the node that's being rewritten.  On the other hand, it's also good to encourage PPX authors to think about the location carefully and to point to the exact point of the error. For example, if the error is due to an unexpected payload, it's nice to point to the exact part of the payload that's unexpected instead of the whole node.

NathanReb: I think it should be discouraged indeed. If there is a single error, whether it's us or them doing the wrapping won't change anything that's true. Raising in such a way prevents from reporting several errors at once though.

From a user perspective that means: fix one error in the payload, build, repeat, until there's no more errors. I think we can all agree this isn't the best workflow.

Having other exceptions than `Location.Error` escape the ppx expander should definitely be discouraged. Raising `Location.Error` is a slightly different matter as in some cases, doing the warpping yourself can be a bit painful so if you know you can only report a single error, raising can prove convenient, we don't necessarily want to impose this burden on ppx authors.

NathanReb: That being said, I'm not sure this should be part of the error message as it's likely to be read by the ppx users rather than authors. I think pointing to the responsible ppx should be enough. Properly documenting the prefered way to report errors and how and where to report such crashes (i.e. on the ppx issue tracker rather than on ppxlib, we could do that in the github issue pattern btw) would be best.

Also is there a reason why this kind of error are reported as raw sexps rather than nicely worded like the other errors in this cram test?

aalekseyev:
As a ppx maintainer, I'm seeing this diff and I'm not sure who the extra information is for.

```
  Line _, characters _-_:
- Error:
+ Error: The following located exception was raised during the context-free transformation phase:
  ppx_hash: cannot hash open polymorphic variant types
```

The extra message seems very confusing for the end user, so maybe it's intended for me, the ppx maintainer, so that I can improve the code to remove the extra message. But I still don't know what I should do to improve it. Should I be doing something other than raising located exceptions?

The discussion above hints that maybe the ppx authors should be emitting error AST nodes instead, but it would be nice to have an explicit doc explaining how to deal with this warning (if it's supposed to be a warning).

pitag-ha:
Hey @aalekseyev , thanks for your feedback!

> The extra message seems very confusing for the end user ([1])

Yes, we've already noticed the same thing and @panglesd is working on improving it. We're now thinking about an error message along the lines:
```
(PPX " ppx_hash":) cannot hash open polymorphic variant types
(Ppxlib:) Such exceptions are strongly discouraged. Other errors can't be reported anymore.
```

> But I still don't know what I should do to improve it. Should I be doing something other than raising located exceptions?

Yes, you shouldn't be raising located exceptions inside a whole-file transformation, since by doing so you prevent ppxlib from returning a meaningful AST to the compiler. But you're absolutely right: that needs to be documented better. We're adding a section to the ppxlib manual about error reporting inside PPXs. That section will also contain the wording in which we extend exceptions coming from PPXs (something like "Such exceptions are strongly discouraged. Other errors can't be reported anymore.") so that it will be easy to google for more information if someone runs into that.

> The discussion above hints that maybe the ppx authors should be emitting error AST nodes instead, but it would be nice to have an explicit doc explaining how to deal with this warning

Yes, that's exactly right: PPX authors should embed the error into the AST via an extension node. And as I've mentioned already: we're adding a section about that to the manual.

Does that resolve your doubts?

pitag-ha:
By the way, to give some more context about _why_ we're adding that message to your errors: there are two reasons for that: one is to let PPX authors know that the way they're reporting their errors is not the encouraged way; and given your message it seems like we're achieving that :tada: (as a side-note: I thought that would start happening once we release and by then we will already have added a section about how you should report your PPX errors to the manual). The second reason is for end users to understand why merlin is only reporting that one error whereas usually it reports all errors in a file. End users tend to think it's merlin's fault and possibly even file issues at merlin, but in fact that's really just on us, where with "us" I mean the PPX side: ppxlib for not providing better information to the PPX authors and the PPX authors for not returning a meaningful AST.

panglesd:
Hello @aalekseyev,
> The end user doesn't care about the "located" part, the "exception" part, and probably doesn't know what "context-free transformation phase" refers to. They might care about which ppx is responsible, but that information is not there.

The reason the name of the PPX is not mentioned is that the exception, in the case of context-free rewriters, is called too late for that.
After #294 is merged, it is caught earlier, and the error message will look like what @pitag-ha said, except for the part ` Other errors can't be reported anymore`.
Indeed, an error node containing the exception will be embedded in the AST, replacing the extension node. However, directly embedding error extension nodes in the AST allows the ppx for better error reporting, such as multiple errors (which is not possible when an exception is raised).

aalekseyev:
Thanks all for clarifications, I understand.
Here are some of my thoughts on the issue:

- changing the ppx code to avoid exceptions can be pretty disruptive, so it won't happen overnight, so a clear end-user-targeted message in the interim seems important;
- most of the errors I'm seeing are about the "context-free transformation phase", where (I think) the problem is less severe, so it's unclear if it's worth a user-facing warning;
- if the user-facing warning is for merlin, maybe it should only be shown in merlin? (I think ppx knows enough to distinguish this situation)
- maybe it should be possible to mark ppxes as legacy w.r.t this issue. For legacy ppxes we can expect that they raise exceptions, so we create the minimum amount of noise (only warning the merlin users in egregious cases), whereas for non-legacy ppxes we give the gory exception details in all cases

(to be clear, I don't necessarily expect you to change what you're doing based on those thoughts, merely offering my perspective)

pitag-ha:
Thanks for your comments and ideas, that's super helpful!

> most of the errors I'm seeing are about the "context-free transformation phase", where (I think) the problem is less severe, so it's unclear if it's worth a user-facing warning;

Yes, for context-free rules the advantage of embedding the error into the AST is being able to give finer-grained errors, so probably nothing that serious that we should disturb the end-user with that.

> if the user-facing warning is for merlin, maybe it should only be shown in merlin? (I think ppxlib knows enough to distinguish this situation);

Yes, ppxlib knows enough. I didn't think about that idea and it's a great idea to think about! On one hand, that would solve several problems, for example PPX authors wouldn't have to adapt their cram tests to our change and the extra information would only be shown in the context where it matters. On the other hand, users are used to seeing the same error from the compiler and from merlin and breaking with that might be confusing. I'll ask the merlin folks what they think about that.

> maybe it should be possible to mark ppxes as legacy w.r.t this issue. For legacy ppxes we can expect that they raise exceptions, so we create the minimum amount of noise (only warning the merlin users in egregious cases), whereas for non-legacy ppxes we give the gory exception details in all cases.

I'm not sure I agree with that point. There are two reasons why we're doing this (as pointed out in more detail above). One is so that the end-user knows the problem lies in PPX and not in merlin; that's also valid for legacy PPXs. And the other one in the context of non-legacy PPXs is to push the PPX authors to improve the error reporting; in the context of legacy PPXs that's not valid but there I'd say it might push the user to use a non-legacy PPX instead. What do you think about that? Do you have (a) certain legacy PPX(s) in mind?

aalekseyev:
>> maybe [the error] should only be shown in merlin?
> ... On the other hand, users are used to seeing the same error from the compiler and from merlin and breaking with that might be confusing.

One way to present this could be to show two separate errors to merlin: the original one, which will be unmodified, and a separate one that says that merlin is broken because the ppx has raised an exception.

> I'm not sure I agree with [legacy/non-legacy distinction] ...
> One is so that the end-user knows the problem lies in PPX ...
> And the other one in the context of non-legacy PPXs is to push the PPX authors to improve the error reportin

Right. My point here is that the two reasons differ in the reasonable level of sensitivity: only major problems are worthy of end-user attention, while PPX authors may want to know about small oddities and best practices, and they may want to fix some issues, while acknowledging and delaying fixes for other issues.

Anyway, if the warning for context-free rules is gone, then the remaining points have little pragmatic interest to me, since most errors I'm seeing are about context-free rules.

(I'm looking at the tests of Jane Street ppxes, and the vast majority of changes are coming from context-free rules)

> Do you have (a) certain legacy PPX(s) in mind?

Well, until proven otherwise I expect most of the existing PPXes will be in this status for a while.

pitag-ha:
> One way to present this could be to show two separate errors to merlin: the original one, which will be unmodified, and a separate one that says that merlin is broken because the ppx has raised an exception.

That idea of yours of adding information when in the context where that information is relevant is great! And in fact, we're going to do that. @panglesd can give more detail if you'd like.

> Well, until proven otherwise I expect most of the existing PPXes will be in this status for a while.

Ah, I see, you meant to take out all existing PPXs from that burden of having to fix their error reporting. I do agree that context-free rules should be treated differently since there the impact is local. So there "correct" error reporting should be considered a good practice (that should be communicated through the manual and announced on discuss) rather than an imposed norm. But on whole-file transformations, also on already existing ones, the problem of raising errors is irreversible and has such an impact (sometimes even on more levels than merlin) that it should be very clear that that's a problem.

> Anyway, if the warning for context-free rules is gone, then the remaining points have little pragmatic interest to me, since most errors I'm seeing are about context-free rules.

I think for context-free rules we should get rid of the statement you call "warning", but it would still be nice to add the name of the PPX where the error came from to the error message. So you would still have to update your cram tests, but without any implicit in there that your PPXs need fixing. Does that sound ok to you?

aalekseyev:
> (that should be communicated through the manual and announced on discuss)

By the way, another way to communicate this could be a deprecation warning on things like `Location.raise_errorf`, which look like the intended way to report errors, but aren't (anymore?).

> I think for context-free rules we should get rid of the statement you call "warning", but it would still be nice to add the name of the PPX where the error came from to the error message. So you would still have to update your cram tests, but without any implicit in there that your PPXs need fixing. Does that sound ok to you?

I'm not sure what exact format you're thinking of, but many ppxes already add their names to their error messages, so adding it again would result in a duplicate name.

pitag-ha:
> By the way, another way to communicate this could be a deprecation warning on things like `Location.raise_errorf`, which look like the intended way to report errors, but aren't

Adding more info on this in the `Location.raise_errorf` docs is also a great idea! I think it could be a link to the manual section rather than a general deprecation warning: again, context-free rules authors should still be able to decide what they want to focus on: simplicity or finer-grained errors. The problem arises in whole-file transformations.

> (anymore?)

I think this might have been a point of confusion, also in our discussion about legacy PPXs. Nothing has really changed in ppxlib here. The only thing that's changing is that we're trying to communicate better what ways of error reporting there are, which advantages each way has and when it leads to a problem to choose one of them.

> many ppxes already add their names to their error messages

Good point! Unfortunately not all though. Still, a good point.

panglesd:
@aalekseyev following your remarks and discussion with @pitag-ha, I opened a PR to fix the confusing error message!

Your comments have been very valuable. It would be great if you can give your though on the new version!

PR #293, opened 2021-10-19
[Preview] upgrade to ocamlformat.0.20.0

gpetiot:
This is a preview of the not-yet-released `ocamlformat.0.20.0`, please wait until the package is published in opam to merge this PR. The output is still likely to slightly change before the package is released.
The changes are due to:
- `module-item-spacing` is now set to `compact` for the default profile
- `module-item-spacing` is now correctly applied to mutually recursive type definitions


pitag-ha (approved):
Thanks for the preview, @gpetiot! I've only rebased over `main` and added a commit to add your commit to `.git-blame-ignore-revs`.

PR #294, opened 2021-10-19
Embed extenders and derivers exception in AST

panglesd:
This PR is yet another PR about improving the error reporting of PPXs made with ppxlib, more precisely it handles issue #291!

Currently, located exceptions raised by a ppx written using ppxlib are caught by the driver and the whole AST is replaced by a single extension error node. While PR #292 works on improving the error message so that the user knows at least where the exception comes from, the other errors message still disappear with the AST being thrown away.

In the case of extenders and derivers, there is no need to get rid of the whole AST: the error extension node can replace the extension/derivation. This way, other error messages can still be reported by merlin for instance. This is what this PR does.

A few remarks on the modifications:
- I needed to propagate the value of `embed_errors` in order to decide to embed or not the exception in the AST. That added an argument to many functions... 
- In `deriving.ml` I made some code refactoring, replacing most of the duplicates `function_name_sig` and `function_name_str` with `function_name context`, using a GADT type `Context.t`. This is not directly related to issue #291 but I thought it had to be done... I can revert but I think it uses a similar coding style as `extension.ml` (although there is only 2 deriving contexts instead of 12 extension context).
- Not only located exceptions are embedded in the AST, but also regular exception, which are turned into located exception using the default location: either the location of the extension point, or of the deriving attribute. This is debatable but I think it is a good behavior.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@ unreleased
 
 - Improve error messages in ppx raised exceptions (#292, @panglesd)
 
+- Embed error messages in the AST for extenders and derivers (#294, @panglesd)
+
 0.23.0 (31/08/2021)
 -------------------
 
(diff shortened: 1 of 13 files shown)


panglesd:
closing in favor of #315, #316

PR #296, opened 2021-10-26
Generating transformation list on order and with no loss

panglesd:
This PR solves the following problems:
- When an `instrumentation` argument was given, together with an `impl` or `intf` argument, during a `Driver.register` call, both the `impl` and `intf` transformations were discarded.
- The order of application was not the one specified in the doc: rules registered together with a `before instrumentation` were applied before the instrumentation, those registered together with an `after instrumentation` were applied just before the after instrumentation (that is, after the non instrumentation whole file transforms).
- When using the `no-merge` command line option, the order was different from the one specified.

The last point is solved by adding a comment in the option decription that enabling it changes the order of execution.

The first two points are solved in a "hotfix" manner. In the future, I think this part of the code would need to be rewritten to make types ensure sanity and prevent other bugs to appear!

(this PR is based on PR #294. Only the last commit is new. Should I rebase for a clearer PR?)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,9 @@ Unreleased
 
 - Ast patterns: add `drop` and `as` patterns (#313 by @Kakadu, review by @pitag-ha)
 
+- Fixed a bug resulting in disscarded rewriters in the presence of
+  instrumentations, as well as a wrong order of rewriting (#296, @panglesd)
+
 0.24.0 (08/12/2021)
 -------------------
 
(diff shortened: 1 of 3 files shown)


pitag-ha:
Thanks @panglesd and sorry for the super delay on this! There are a lot of new commits on this branch. Do I understand correctly that the only relevant one is the last one?

panglesd:
Yes, it was based on PR #294. and depended on some changes, but I updated it to `main` given that we will most certainly close #294.

pitag-ha:
Thanks! Very good catch and nice quick solution. Could you add a changelog entry mentioning this fix (possibly after rebasing over `main` once more to make sure there's an `unreleased` section)?

pitag-ha (approved):
Nice, thank you!

PR #298, opened 2021-10-27
Use packaged changelog checks

NathanReb:
This replaces our own changelog checks with its packaged version (by @Leonidas-from-XIV) and adds a nice action
that rewrites `#<PR_NUMBER>` into the PR number ref, useful for maintaining our changelog format!


Leonidas-from-XIV:
> Just out of curiosity: what does branding in both actions do? I thought it would maybe define the icon on the left to the action name in the CI, but I've had a look at dune-release's CI and that doesn't seem the case.

Ha, that would be a nice usage indeed! But alas it is only relevant for the [Github marketplace](https://github.com/marketplace) where it displays [a blue hash sign](https://github.com/marketplace/actions/populate-pr-in-changelog) for [both actions](https://github.com/marketplace/actions/check-changelog-for-changes). There is a set of valid colors and symbols to choose from so I just picked whichever seemed the most appropriate at that time.

pitag-ha:
Aah ok, cool. Thanks for explaining!

PR #299, opened 2021-11-04
Fix `type_is_recursive` and `really_recursive`

NathanReb:
These functions used to return `Recursive` if the types appeared inside attributes within the type declarations.

The bug and the fix were discovered by @ceastlund while trying out #297 on Jane Street's code base.

I split this in two parts: first one adds tests, second one fixes the bug and updates the tests accordingly.

Let me know what you think!

### DIFF
--- a/.ocamlformat-ignore
+++ b/.ocamlformat-ignore
@@ -36,3 +36,4 @@ test/location/exception/test.ml
 test/ppx_import_support/test.ml
 test/quoter/test.ml
 test/traverse/test.ml
+test/type_is_recursive/test.ml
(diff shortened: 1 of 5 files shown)


pitag-ha (approved):
Thanks! LGTM.

I'm curious: how did #295 / #297 make you find out about the bug? Since attributes are kept in the AST even after being applied, I wouldn't have expected that #295 / #297 make a difference with regard to this bug.

NathanReb:
It is indeed a broader bug.

@ceastlund will confirm but I think it's discovery is related to the change in application order, i.e. the fact that deriver now receive the expanded node. It seems some ppx used at Janestreet will add attributes that weren't visible by the derivers before. Since those derivers use `really_recursive` (for obvious reasons), they were wrongly assuming the type were recursive because of those generated attributes.

#297 made discovering this possible because before it, the builds would fail at earlier stages, before the ppx-es were all applied.

PR #301, opened 2021-11-05
Add a comment with an example for module_substitution

bcc32:
Signed-off-by: Aaron L. Zeng <azeng@janestreet.com>

### DIFF
--- a/ast/ast.ml
+++ b/ast/ast.ml
@@ -856,6 +856,7 @@ and module_substitution = Parsetree.module_substitution = {
   pms_attributes : attributes;
   pms_loc : location;
 }
+(* S := M *)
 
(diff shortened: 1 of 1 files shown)


pitag-ha (commented):
Thanks for the PR! Adding that comment would indeed be helpful and more coherent. The problem about that is that our `Ast` module is generated from one of the `Astlib.Ast_xx` modules (formerly from OMP's `Ast_xx` module) which in turn is generated from the compilers `Parsing.Patsetree` module. So even if we added this comment here now, it would probably disappear again in the next AST bump. Have you considered opening a PR to the compiler?

bcc32:
Thanks @pitag-ha, of course it is generated :).  Closing in favor of https://github.com/ocaml/ocaml/pull/10784

PR #303, opened 2021-11-15
Document the get_group function in Context_free

NathanReb:
The implementation puzzled me at first so I thought it might make sense to document what it does for future reference.


pitag-ha (approved):
That's interesting. What performance does to code... Thanks for adding the comment!

PR #307, opened 2021-12-02
Compact yet informative error reporting

panglesd:
This PR aims at providing a better reporting of errors, following the discussion from #292  and the new way of embedding errors from #294.

# Aim

After discussion, an error message should:
- Give the information needed for the user, and nothing more, that is:
  - The name of the ppx throwing the error
  - the actual error,
  - the explanation of why some features are missing (such as no other errors reported in merlin, or no location given), only when that makes sense.
  - separated in a clear way
- Give a hint to the ppx author on how to improve the error reporting.
  - Give a warning about the fact that the error is not reported the right way (but only when it is not reported the right way)
  - Use a formulation that is easy to google, hoping that it gets to the ppxlib manual on error reporting (wip)
  
# Implementation

With the aim in mind, here is how the errors are generated:
- If the error is linked with a ppx, the error starts with `(PPX name_of_ppx:) `. If it happens during the rewriting phase but is not linked with a ppx, it starts with: `(Ppxlib context-free phase:) `,
- then there is the error
- if either:
  - Case 1: A located error was raised in a whole file rewriter, or
  - Case 2: A regular exception was raised in a context-free rewriter
  - Case 3: A regular exception was raised in a whole file rewriter
then `(Ppxlib:)  ppx_name has violated the PPX norm on how to report errors.` is added in a next line
- Furthermore:
  - In case 3, a suffix `No location of error can be given.`
  - In case 2, a suffix `Precise location of error cannot be given.`
  - In case 1 or 3, when the tool is merlin, a suffix `Other errors can't be reported anymore.`

# Discussion 

To continue the discussion, I have some questions:
- In case 1, when the tool is not Merlin, the reported error is not justified by a suffix: only  `(Ppxlib:)  ppx_name has violated the PPX norm on how to report errors.`. Would it be better to add `Merlin won't be able to report other errors.`?
- Is case 2 `Precise location of error cannot be given.` Can this be missleading, as the location of the extension node is already given?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@ unreleased
 
 - Improve error messages in ppx raised exceptions (#292, @panglesd)
 
+- Embed error messages in the AST for extenders and derivers (#294, @panglesd)
+
 0.23.0 (31/08/2021)
 -------------------
 
--- a/src/context_free.ml
+++ b/src/context_free.ml
@@ -197,44 +197,38 @@ module Generated_code_hook = struct
     | _ -> t.f context { loc with loc_start = loc.loc_end } x
 end
 
-let rec map_node_rec context ts super_call loc base_ctxt x =
+let rec map_node_rec context ts super_call loc base_ctxt x
+    ?(embed_errors = false) =
   let ctxt =
     Expansion_context.Extension.make ~extension_point_loc:loc ~base:base_ctxt ()
   in
   match EC.get_extension context x with
   | None -> super_call base_ctxt x
   | Some (ext, attrs) -> (
-      match E.For_context.convert ts ~ctxt ext with
+      match E.For_context.convert ts ~ctxt ext embed_errors context x with
       | None -> super_call base_ctxt x
       | Some x ->
           map_node_rec context ts super_call loc base_ctxt
-            (EC.merge_attributes context x attrs))
+            (EC.merge_attributes context x attrs)
+            ~embed_errors)
 
-let map_node (context:'a EC.t) ts super_call loc base_ctxt (x:'a) ~hook ?(embed_errors = false)
-    =
+let map_node (context : 'a EC.t) ts super_call loc base_ctxt (x : 'a) ~hook
+    ?(embed_errors = false) =
   let ctxt =
     Expansion_context.Extension.make ~extension_point_loc:loc ~base:base_ctxt ()
   in
   match EC.get_extension context x with
   | None -> super_call base_ctxt x
-  | Some (((str_loc, _) as ext), attrs) ->
-      (try
-         match E.For_context.convert ts ~ctxt ext with
-         | None -> super_call base_ctxt x
-         | Some x ->
-             map_node_rec context ts super_call loc base_ctxt
-               (EC.merge_attributes context x attrs)
-       with exn when embed_errors ->
-         let extension_node =
-           Location.Error.make ~loc
-             ("(ppx " ^ str_loc.txt ^ ") " ^ Printexc.to_string exn)
-             ~sub:[]
-           |> Location.Error.to_extension
-         in
-         EC.extension_builder context x ~loc:str_loc.loc extension_node)
-      |> fun generated_code ->
-      Generated_code_hook.replace hook context loc (Single generated_code);
-      generated_code
+  | Some (ext, attrs) -> (
+      match E.For_context.convert ts ~ctxt ext embed_errors context x with
+      | None -> super_call base_ctxt x
+      | Some x ->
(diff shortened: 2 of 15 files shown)


aalekseyev:
As requested, I'll give my opinion on what's described.

- `(PPX name_of_ppx:)` seems like a lot of punctuation. `name_of_ppx:` alone seems sufficient, and ideally there'd be a way to avoid duplication of that name in cases where it's already present. I'm not sure why this change is necessarily linked to the rest of the changes, either. Seems like a separate improvement. (or disimprovement in case of duplicated name)
- Maybe a bit more human-friendly and clearer would be to say `No location of error can be given because ppx_bin_prot raised an exception`. This way it's clear what norm was violated, by whom, and how it's related to location. (same for "precise location")
- I think saying `Other errors can't be reported anymore.` is a big understatement: isn't merlin completely broken in that case? Completion doesn't work, type information doesn't work, etc? Maybe I'm wrong about this, but if I'm right the message should probably be more general than simply mentioning the number of errors.

aalekseyev:
By the way, @ceastlund, can you please take my place under this bus?
I can give opinions all day long, but it's you who will be responsible in the end. :-)

pitag-ha:
Thanks @aalekseyev for your input! I agree with @panglesd that it has been very valuable, but also, no need to keep on being involved in this :)

> isn't merlin completely broken in that case? Completion doesn't work, type information doesn't work, etc?

Yes, it's really not a great situation. That's why we're making an effort in communicating well to PPX authors that that should really be avoided.

pitag-ha (commented):
Thanks for the improvement @panglesd! 

I would add two more points to the **Aim** you've formulated above:
- keep it as simple as possible
- avoid making the PPX authors feel like we're blaming them for something they didn't know better due to us

To also keep those two points in mind, I would suggest the following:
For a located exception in whole file transformations:
```
(<name_of_ppx>): <error_msg_from_ppx>
(ppxlib): IDE / source code analysis features aren't available anymore due to PPX error reporting.
```

For a non-located exception anywhere:
```
(<name_of_ppx>): <error_msg_from_ppx>
(ppxlib): the location of the error is undefined and IDE / source code analysis features aren't available anymore, because <name_of_ppx> has raised.

```

Distinguishing more cases depending on different combinations of type of PPX, type of exception and tool that has called us has been important to consider to find a good solution, but it makes things overly complex.

Does that sound good to you?

For the rest, I'll review #294 and #296 first to have a better overview of how things are connected.

Also, apologies for the delay in review and for underestimating the complexity and the amount of work in this context! It's really great that you're going into a lot of detail (and find bugs/totally wrong behavior in ppxlib etc)!

pitag-ha:
Since we're a bit in a hurry to release (for 4.14 support), let's wrap up the whole-file transformation case and finish the context-free rule case (which is far more complex since you've been working on embedding local exceptions into the global AST) after the release.

So maybe you can cherry-pick this commit over to `main` and improve it a little bit to get the behavior we want for now?

Does that sound good to you?

pitag-ha:
To make a bit clearer what I mean: for the release, the only change we want, compared to how it was before improving the error messages is:
For a located exception in whole file transformations:
```
(<name_of_ppx>): <error_msg_from_ppx>
(ppxlib): IDE / source code analysis features aren't available anymore due to PPX error reporting.
```

For a non-located exception anywhere:
```
(<name_of_ppx>): <error_msg_from_ppx>
(ppxlib): the location of the error is undefined and IDE / source code analysis features aren't available anymore, because <name_of_ppx> has raised.
```

So for now, context-free rules would only be affected by the change if they raise non-located exceptions. After the release we will then get to context-free rules and, among the big change of embedding errors, also decide if we want to add their name to their error message always or only if its name isn't already contained in their error message or never.

panglesd:
> So maybe you can cherry-pick this commit over to main and improve it a little bit to get the behavior we want for now?

Sure, I can do this. However, you're asking that in the case of non-located exceptions in context-free rules, the name of the ppx is displayed.
This means that the exception in context-free rules has to be caught earlier, in fact exactly at the same time as needed to embed the exception in the AST. So I will have to add lots of the code from the "embedding in AST" PR #294. This is fine on my side, is it fine on yours?
Also, I will make it so that there is no API change in `extension.mli`.

pitag-ha:
> Sure, I can do this. 

Nice, thanks a lot!

> This means that the exception in context-free rules has to be caught earlier, in fact exactly at the same time as needed to embed the exception in the AST. So I will have to add lots of the code from the "embedding in AST" PR #294.

Very good point! Let's not do that now then. Let's keep it as simple as possible for now.

PR #304, opened 2021-11-16
OCaml 4.14 support

kit-ty-kate:
Mirrors https://github.com/ocaml-ppx/ocaml-migrate-parsetree/pull/119

PR #308, opened 2021-12-02
Update opam file (OCaml 4.14 support added in #304)

kit-ty-kate:


pitag-ha:
Good catch, thanks! Could you sign the commit (sorry for that hussle with signing the commit all the time)? We'll also need to adapt dune-project file, but I can do that if you want.

kit-ty-kate:
Done

pitag-ha (approved):
Cool, thanks!

PR #309, opened 2021-12-08
more informative error reporting for whole file transforms

panglesd:
This PR modifies the error reporting for error reporting, in a more informative way.

Here is the diff for the [test](https://github.com/ocaml-ppx/ppxlib/blob/main/test/driver/exception_handling/run.t) about error reporting, when run in the context of the last release (0.23.0) and after the PR:
```diff
$ diff --context=2 error_reporting_old.txt error_reporting_new.txt
*** error_reporting_old.txt     2021-12-08 11:32:57.606969043 +0100
--- error_reporting_new.txt 2021-12-08 11:29:47.274410449 +0100
***************
*** 65,69 ****
    $ ./whole_file_located_error.exe impl.ml
    File "impl.ml", line 1, characters 0-12:
!   Error: A located error in a whole file transform
    [1]
  
--- 65,70 ----
    $ ./whole_file_located_error.exe impl.ml
    File "impl.ml", line 1, characters 0-12:
!   Error: (raise_exc): A located error in a whole file transform
!   (ppxlib): IDE / source code analysis features aren't available anymore due to PPX error reporting.
    [1]
  
***************
*** 92,96 ****
    $ echo "let x = 1+1. " > impl.ml
    $ ./whole_file_located_error.exe -embed-errors impl.ml
!   [%%ocaml.error "A located error in a whole file transform"]
  
  3. Raising an exception. The exception is caught by the driver, and
--- 93,98 ----
    $ echo "let x = 1+1. " > impl.ml
    $ ./whole_file_located_error.exe -embed-errors impl.ml
!   [%%ocaml.error
!     "(raise_exc): A located error in a whole file transform\n(ppxlib): IDE / source code analysis features aren't available anymore due to PPX error reporting."]
  
  3. Raising an exception. The exception is caught by the driver, and
***************
*** 119,126 ****
  
    $ ./whole_file_exception.exe impl.ml
!   Fatal error: exception (Failure "An exception in a whole file transform")
    [2]
    $ ./whole_file_exception.exe -embed-errors impl.ml
!   Fatal error: exception (Failure "An exception in a whole file transform")
    [2]
  
--- 121,132 ----
  
    $ ./whole_file_exception.exe impl.ml
!   Fatal error: exception (raise_exc): (Failure "An exception in a whole file transform")
!   (ppxlib): the location of the error is undefined and IDE / source code analysis features aren't available anymore, because raise_exc has raised.
!   
    [2]
    $ ./whole_file_exception.exe -embed-errors impl.ml
[...]

PR #311, opened 2021-12-08
Revert "Adding more explanatory error messages in case of exception throwing"

pitag-ha:
This temporarily reverts #292.

The reason for that decision is to have a clean and non-controversial `main` quickly and to be able to cut a non-risky release with 4.14 support easily. After cutting the release, we'll re-integrate that change again, but step-by-step, in an improved way (see #309) and as part of a whole plan (I'll open an issue to discuss that plan) that we'll first discuss with the merlin team as well before merging parts of it again.

By the way: thanks a lot, @panglesd, for your work and for staying with me on this while changing my mind!


pitag-ha:
Also, cc @aalekseyev and @kit-ty-kate, just to inform you about the revert since you've already come across the change that I'm reverting here for now.

Issue #310, opened 2021-12-08
Two questions about first class patterns: dropping and giving a name to subvalues


Kakadu:
While using `Ast_pattern` I got a few issues. 

1) I can't find any function that ignores part of AST (i.e. wildcard semantics). Right now `__` behaves as pattern variable and obliges to receive that value in "action code" (I'm using parser-generator vocabulary here). 
2) I can't find any way to give another name to part of AST, i.e. I want something similar to `as` keyword in OCaml. If it would be available we could express matching a structure item which is a type declaration in more convenient manner, something like 
```
pstr_type nonrecursive  
  (__as (type_declaration ~name:drop ~params:drop ~cstrs:drop ~kind:drop  ~private_:drop ~manifest:drop))
- : (Ppxlib__.Import.structure_item,
     Ppxlib__.Import.structure_item -> '_weak1, '_weak1)
    Ppxlib__.Ast_pattern0.t
= Ppxlib__.Ast_pattern0.T <fun>
(* future hypothetical type *)
```

If I'm not  blind indeed, I probably could make a PRs this functionality.

pitag-ha:
Hey @Kakadu , thanks for the issue and sorry for the delay in answering. Extending `Ast_pattern` sounds great! I have a couple of questions about your example just for better understanding:

What do you mean with `drop`? Is that the name you give to the "function that ignores parts of the AST" as you describe in 1.?

And, what exactly is it that you want to achieve in your example? What I understand is that you'd like to

> express matching a structure item which is a type declaration

Is that correct? If so, is that similar to the following?
```
pstr (pstr_type nonrecursive (__ ^:: nil) ^:: nil)
```

Kakadu:
@pitag-ha We together are masters of delayed answers!

There is a chance that my example from the issue is not the best and you can indeed express what I wanted. But what about "single type declaration with any recursivity flag"?

Also I made a PR where I motivated my desire to have new combinators in context of writing a linter. Could you look into it?

pitag-ha:
Oh, I somehow missed that PR, thanks for pinging me! I'll review tomorrow.

Btw, I've just seen that there's another issue about `drop` with a very similar motivation as what your saying about "single type declaration with any recursivity flag": #89

Kakadu:
Resolved in https://github.com/ocaml-ppx/ppxlib/pull/313

PR #313, opened 2022-01-08
Ast_pattern: add patterns drop and as__ (#310)

Kakadu:


src/ast_pattern.mli
```diff
@@ -109,6 +109,18 @@ val __' : ('a, 'a Loc.t -> 'b, 'b) t
...
+(** Useful when we don't care about some part of OCaml AST. With [__] the actual
+    value will be passed to success continuation, with [drop] it will be ignore.
+    In normal pattern matching it is called wildcard pattern. *)
+
+val as__ : ('a, 'b, 'c) t -> ('a, 'a -> 'b, 'c) t
+(** As-pattern. Passed the current AST to success continuation.
+
+    Pitfall. The part of AST is being passed before checking that it fits the
+    pattern. It means that success continuation may fire earlier than we realize
+    that pattern argument of [as__] doesn't fit. *)
```
pitag-ha, line 122: Good observation! That shouldn't have a strong impact on the user though, should it? I mean, even if the success continuation set some global reference or similar, the program would stop right afterward. (Also, I'm wondering if the same happens in other combinators as well or if this is the only situation where that happens)

test/patterns_as_and_drop/test.ml
```diff
@@ -0,0 +1,69 @@
...
+  in
+  printf "Wrong!\n%!";
+  Ast_pattern.parse (pat ()) loc ast
+    ~on_error:(fun () -> print_endline "error")
+    (fun _zero_expr -> Format.printf "Success and error after that\n%!");
+  printf "Correct!\n%!";
+  Ast_pattern.parse (pat ()) loc ast
+    ~on_error:(fun () () -> print_endline "error")
+    (fun _zero_expr () -> Format.printf "Success and error after that\n%!")
+    ()
```
pitag-ha, line 69: Am I understanding correctly that here the context is a function from `unit` to `string`?

pitag-ha (commented):
Thanks for the PR @Kakadu! Adding `as__` and exposing `drop` seems like a good addition to `Ast_pattern` to me.

About the tests: They're a bit tedious to read since the test code and test result are in two different files. What would you think about making `[%ppx_expect]`-tests out of them? You can see an example of that e.g. [here](https://github.com/ocaml-ppx/ppxlib/blob/main/test/extensions_and_deriving/test.ml). What do you think?

Kakadu:
> What would you think about making `[%ppx_expect]`-tests out of them? You can see an example of that e.g. [here](https://github.com/ocaml-ppx/ppxlib/blob/main/test/extensions_and_deriving/test.ml). What do you think?

I copied your example, and your expect-like executable writes empty file when something is wrong with the test. For example, I changed the output, got empty '.corrected' file, changed back, and got again empty '.corrected' file.  So, it is almost impossible to understand what is wrong. 

Using a clever guess I realized that it should be somehow connected to autoformatting. Indeed, when I added `[@@@ocamlformat "disable"]` to `test/extensions_and_deriving/test.ml` I managed to brake test and revert it back. But I can't do the same trick with my test, because it is already formatted wrong (in a style that your tool `test/extensions_and_deriving/.bin/expect-test` doesn't understand).

> What do you think?

Could we somehow agree on that PR without me touching an abomination called `expect-test`?

test/patterns_as_and_drop/test.ml
Kakadu: I don't understand a question. Ast_pattern0 defines type `context ` which is isomorphic to integer, so it is not a function.

src/ast_pattern.mli
Kakadu: > That shouldn't have a strong impact on the user though, should it? 

I had a bug about it, that took too much time to debug. 

> (Also, I'm wondering if the same happens in other combinators as well or if this is the only situation where that happens)

It looks like it is not being introduced by. I just was very disappointed about this pitfall, that's why this PR looks like it is.

pitag-ha: > I had a bug about it, that took too much time to debug.

That's annoying. Would it take long to explain the bug?

test/patterns_as_and_drop/test.ml
pitag-ha: You're right, I need to explain myself better since in ppxlib there are lots of different concepts that we call context. What I was referring to is the type parameter `'c` in `('a, 'b, 'c) Ast_pattern.t` (i.e. the return type of the continuation `'b`). Usually, that type parameter stands for the context inside the AST in which you want to inject the derived/expanded code.

Kakadu: Context is `unit -> unit` because both `() -> print_endline "error"` and `fun () -> Format.printf "Success and error after that\n%!"` have this type

src/ast_pattern.mli
Kakadu: I forgot to add `()`. I wrote like in `test.ml:62` but it was required to write like on line `test.ml:66`.

pitag-ha:
> For example, I changed the output, got empty '.corrected' file, changed back, and got again empty '.corrected' file. So, it is almost impossible to understand what is wrong.

Yes, I agree, that's very annoying. That happens when you don't have any `%%expect` extension node in your test file (for example, if you have a typo like `%%espect` or `%expect` or so). That's because we're using our own expect test version as opposed to the ppx_expect opam package to avoid circular dependencies.

> it should be somehow connected to autoformatting.

The reason why we disable ocamlformat for expect test files is that our expect test doesn't respect ocamlformat formatting, so promoting `dune runtest` and `dune build @fmt` would yield conflicting results when ocamlformat isn't disabled. 

In case you want to give it another try: I'd expect the following to work (you'd simply have to `dune promote` it):
```
(* The `as__` combinator captures the whole pattern it is given
  in addition to the internally captured values in that pattern.
  The value it captures is passed to the continuation before the internally captured values. *)
let parse_result =
  let loc = Location.none in
  let ast = [%expr List.length xs = 0] in
  let pat =
    let open Ast_pattern in
    let length () =
      as__
        (pexp_apply
           (pexp_ident (ldot (lident (string "List")) (string "length")))
           ((nolabel ** drop) ^:: nil))
    in
    let zero () = as__ (pexp_constant __) in

    pexp_apply
      (pexp_ident (lident (string "=")))
      ((nolabel ** length ()) ^:: (nolabel ** zero ()) ^:: nil)
  in

  Ast_pattern.parse pat loc ast
    (fun first_as_captured_val second_as_captured_val intrenally_captured_val ->
      (first_as_captured_val, second_as_captured_val, intrenally_captured_val))
[%%expect{|
|}]
```
It's pretty much your second test with one modification: it tests the behavior of `as__` itself. Testing the pitfall as well is also great though!

If you still really don't want to use the expect test framework, we can find another solution.

test/patterns_as_and_drop/test.ml
```diff
@@ -0,0 +1,81 @@
...
+let () =
+  let ast = [%expr string_of_int 42] in
+  let pat =
+    let open Ast_pattern in
+    pexp_apply (pexp_ident (lident __)) ((nolabel ** eint (int 42)) ^:: nil)
+  in
+
+  Ast_pattern.parse pat loc ast
+    ~on_error:(fun () -> print_endline "error")
+    (fun s -> printf "success %s, but actually not" s)
```
pitag-ha, line 81: How does this new test reveal the pitfall? At first sight, it seems like the `ast` you define simply fulfils the pattern you define. So why is this "actually no success"? (I'm probably be missing something)

src/ast_pattern.mli
pitag-ha: Out of curiosity: What's your use case of `Ast_pattern`? I'm wondering because you seem to use the `parse` function yourself on the pattern as opposed to passing your pattern to `Extension.declare` or `Deriving.Args.arg` or similar.

test/patterns_as_and_drop/test.ml
Kakadu: Actually, it was a little bit crash and this test doesn't represent what I wanted. Not it should be better.

Kakadu:
> If you still really don't want to use the expect test framework, we can find another solution.

It looks like I managed to rewrite with expect-tests

pitag-ha (commented):
Thanks for the improvements! I only have a couple of minor comments about the documentation that I'd like to be addressed before merging (notice that my suggestions aren't formatted).

src/ast_pattern.mli
pitag-ha: Ok, sounds interesting! And I can get an idea of why you're parsing your ast_pattern yourself.

test/patterns_as_and_drop/test.ml
pitag-ha: Thanks!

Kakadu:
@pitag-ha  ^^

pitag-ha (approved):
Thanks for the change. LGTM!

Issue #314, opened 2022-01-12
Ppxlib erases the AST when embedding an error


panglesd:
# Current state

Currently, in case of a ppx raising an exception, ppxlib does the following:

- If it’s a located exception and either the `embed_errors` or the `as-ppx` flag is set, the whole AST is trashed and replaced by a single error extension node,
- Otherwise, raised exception are not caught at all.

In terms of tools:
- Merlin is calling the ppx with the `as-ppx` flag,
- Dune is calling the ppx without the `embed_errors` or `as-ppx` flag (it uses the `-dump-ast` flag).

# The problem

The problem with this is the use of ppx with Merlin.

When a ppx raise a located exception, the whole AST is removed, which means that *any other analysis or feature from Merlin is not available anymore*. Moreover, the confused user is not given a clue about why this happen.

Finally, the ppxlib API, docs and examples does not incentive the ppx authors to embed errors in the AST but to raise exceptions, which results in many ppxes raising.

# The proposed solution

The behaviour of ppxlib should be modified to replace:
> - When a ppx raise a located exception and the `embed_errors` flag is set, the whole AST is trashed and replaced by a single error extension node

by
- When a ppx raise a located exception and the `embed_errors` flag is set, the last valid AST (the one given to the faulty rewriter) is kept, prefixed by an extension error node containing the error.

Moreover, the docs and API should be changed to reflect that embedding errors in the AST is the preferred way, to explain that raising located errors stops the rewriting process and that unlocated exceptions should never happen.

The tasks are the following:

- Modification of the behaviour:
  - [ ] Add the last valid AST after the error exception node.
- Modification of the doc:
  - [ ] Add an `error_extensionf` function is the public API, and update examples to use it.
  - [ ] Add a section about error reporting in the manual.

PR #316, opened 2022-01-12
Add `error_extensionf` in the API to ease the process of embedding errors

panglesd:
This is part of the #314 tasks.

The examples are updated to embed errors using the new function, instead of raising located errors.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 Unreleased
 ----------
 
+- Added `error_extensionf` function to the `Location` module (#316, @panglesd)
+
 - Ast patterns: add `drop` and `as` patterns (#313 by @Kakadu, review by @pitag-ha)
 
 0.24.0 (08/12/2021)
@@ -23,8 +25,6 @@ Unreleased
 - Fix a bug in `type_is_recursive` and `really_recursive` where they would
   consider a type declaration recursive if the type appeared inside an attribute
   payload (#299, @NathanReb)
-
-- Added `error_extensionf` function to the `Locate` module (#316, @panglesd)
 
 0.23.0 (31/08/2021)
 -------------------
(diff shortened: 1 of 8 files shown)


pitag-ha (commented):
Perfect, LGTM. Thanks, @panglesd! Could you add a changelog entry?

pitag-ha:
Oh, and one more thing: we should also edit the documentation of [`Location.raise_errorf`](https://github.com/ocaml-ppx/ppxlib/blob/main/src/location.mli#L27-L290) to point out that that function should be avoided and `error_extensionf` should be used instead. Also, the sentence "The exception is caught by the driver and handled appropriately" is very confusing, if not wrong. What do you think? Do you want to update that here or rather in a separate PR?

panglesd:
Sure! I added a changelog entry and updated the doc.

When this PR and the #318 PR are merged, there should be a link from the API of those functions to the relevant part of the manual. The maintenance of this link would be eased if the manual was written using `odoc` :wink:

CHANGES.md
```diff
@@ -19,6 +19,8 @@
   consider a type declaration recursive if the type appeared inside an attribute
   payload (#299, @NathanReb)
 
+- Added `error_extensionf` function to the `Locate` module (#316, @panglesd)
```
pitag-ha, line 22:
```suggestion
- Added `error_extensionf` function to the `Location` module (#316, @panglesd)
```

Also, similarly to [here](https://github.com/ocaml-ppx/ppxlib/pull/315#discussion_r797739592): would you mind rebasing over main and moving this entry up to the `unreleaed` section (it's currently in the section of the last release).

pitag-ha (commented):
Perfect, thanks! Let's move the changelog entry to the `unreleased` section and merge this.

pitag-ha (approved):
Perfect, thank you!

PR #319, opened 2022-01-21
Support 5.0

pitag-ha:
This is a temporary solution for ppxlib to support `5.00.0+trunk`. Once [the issue about the AST magic numbers on the compiler](https://github.com/ocaml/ocaml/issues/10936) is solved, this will either become a permanent solution (option 1. on the issue) and I'll undraft, or I'll update the branch to a new solution (option 2. on the issue).

## Available on opam

This preview for `5.00.0+trunk` support is already available on opam as version `0.25.0~5.00preview`. So if you want to try multicore on projects with ppx rewriters, you can simply `opam install ppxlib.0.25.0~5.00preview`.

The preview on opam contains both this commit and a commit removing the dependency on sexplib0.

## About this solution

The AST hasn't changed from 4.14 to 5.0. Furthermore, the magic numbers
from 4.14 haven't been bumped (yet). So the only thing that needs to be
taken care of is the dune `ocaml_version` variable.

IMPORTANT: This might only be a temporary solution: if the magic numbers
still get bumped, ppxlib will need to add some extra logic to support
those magic numbers.


pitag-ha:
The compiler team has decided to bump the magic numbers, so this simple solution doesn't won't work anymore. I've opened #348  instead.

PR #320, opened 2022-01-26
Bump AST to 4.14/5.00

pitag-ha:
This is a preview for bumping the AST to 4.14 (which is the same as the 5.00 AST). I haven't tested it yet. This bump might break quite some PPXs, since `ppat_construct` has changed.

This commit also contains all the changes that have been made to `Pprintast` in the compiler since the last bump since the ppxlib `Pprintast` should be up-to-date with the compiler `Pprintast` anyways and some of those changes were necessary to adapt to the new AST version.


pitag-ha:
This PR is ready for review now. @NathanReb or/and @panglesd, if you could have a look at it, that would be great!

I've created the current [ppx-universe](https://github.com/ocaml-ppx/ppx_universe) with 114 reverse dependencies and, as expected, due to the change to `ppat_construct` _a lot ot_ packages broke at the beginning. That's why I've added the compatibility layer to `Ast_builder` and kept `Ast_helper` stable (btw, we should deprecate `Ast_helper` asap) as you can see in the last two commits. Let me know what you think about that.

I'll be on vacation for the whole of next week. It's likely that 4.14 is going to enter the rc phase next week as well, so I'd like to merge and release this soon once I'm back.

panglesd:
I do not know this part of the code yet, in particular I am not totally sure of which part of the commit is auto-generated from `ocaml` codebase, (eg `printast.ml` I guess) and which part is hand-written (and need review). I'll focus on the last two commits, which are clearly hand-written.

I don't really like the versioning as it is... I think the version name has no clear semantic, resulting in:
- maintenance difficulties (what happens if something else than `Ppat_construct` is changed later, do we add it to `V2` or to `V3`? And what if later, `Ppat_construct` is changed again?)
- User confusion: When reading a doc, it is cumbersome to have `V2`s and `V3`s with subtle changes, it is unclear which to use...

I would rename modules with more explicit name:
```ocaml
module From_413 = struct
  let ppat_construct = ppat_construct

  let pconstruct = pconstruct

  let constructor_declaration = constructor_declaration
end

module Pre_413 = struct
  let ppat_construct ~loc lid p =
    ppat_construct ~loc lid (Option.map p ~f:(fun p -> ([], p)))

  let pconstruct cd arg =
    ppat_construct ~loc:cd.pcd_loc (Located.map_lident cd.pcd_name) arg

  let constructor_declaration = constructor_declaration ~vars:[]
end

let pconstruct = Pre_413.pconstruct
let constructor_declaration = Pre_413.constructor_declaration
```
Another solution that is feasible in this case, is to use default values: even though the signature changes, all valid code remain valid, but the new feature is accessible:
```ocaml
  let ppat_construct ?(vars = []) ~loc lid p =
    ppat_construct ~loc lid (Option.map p ~f:(fun p -> (vars, p)))

  let pconstruct ?(vars = []) cd arg =
    ppat_construct ~vars ~loc:cd.pcd_loc (Located.map_lident cd.pcd_name) arg

  let constructor_declaration ?(vars = []) ~loc ~name ~args ~res =
    constructor_declaration ~loc ~name ~args ~vars ~res
```

I am not sure which one of the two is better.

--------

Is there a reason the new feature is not added on `ast_helper`? Is it because this module will be deprecated? Otherwise, any solution above could be applied to get both compatibility and the new feature.

I hope you will have had a nice vacation!

pitag-ha:
> I think the version name has no clear semantic, resulting in: (...)

Thanks for pointing that out, that's very true!

About the second of your two suggestions, i.e. the one adding an optional argument to allow choosing between the old and the new behavior: I originally ruled that one out since I wanted to provide both a stable and a generated function for the following reason: the stable function would allow users who don't need the new feature to stay stable; the generated function would guarantee that we keep providing ways to add the newest features when creating that node also in the future. To be fair, I'm not sure if the second is really necessary since I'm not even sure if there will be people who'd need to create nodes with those latest features.

I'll have a quick look tomorrow to get an impression if that will be needed. If not, let's go for your second suggestion with one modification: let's make the function with the optional argument stable rather than relying on the generated function. If it seems like we should stick to the idea of the two versions, I'll definitely pick up your first suggestion, i.e. I'll rename the submodule to something like `From_413` or `Post_413` as you've suggested, which is definitely way better than the current meaningless `V2`!

pitag-ha:
Hey @panglesd, I've given this another round of thoughts and my opinion on this now is the following. I still think the idea of optional arguments can be a good idea in general, but in these concrete cases, I don't think it would work: For functions like `constructor_declaration`, optional arguments wouldn't help since all the arguments of the function are labeled. For functions like `ppat_construct`  the suggestion of optional arguments doesn't seem to match the nature of the new feature: what if someone decided to hand in a non-empty list as `vars` while handing `None` to `p`? That wouldn't make sense in the context of the node: `p` being `None` means that the constructor doesn't have arguments, and hence there's no existential, and so no existential could be named.

That's why I'm sticking to the solution of having two versions of the same function: an up-to-date version inside a new module, and a stable version outside that new module. I've taken your advice of giving that new module a more meaningful name than `V2`! It's now called `Unstable`. I prefer `Unstable` to `From_413`, `From_414` since I don't think we should add a new module for every single AST bump. Also, the name `Unstable` makes very clear that it should only be used when needed.

To address your concerns with this solution I'm still sticking to:

> what happens if something else than `Ppat_construct` is changed later, do we add it to `V2` or to `V3`?

I think there should be no such thing as `V3`. There should be only one module containing all concerned functions in its up-to-date version (and that module is called `Unstable` now).

> And what if later, `Ppat_construct` is changed again?

I don't think the functions inside `Unstable` should be kept stable (hence the name for the module). I think they should always be kept up-to-date adapting the newest compiler features. If a user wants to use a function that's likely to stay stable, they should use one from outside `Unstable`. If a user needs to use a function from within `Unstable`, they'll receive a patch PR from us. What do you think about that? I could explain a bit more why I think `Unstable` shouldn't be kept stable.

> User confusion: When reading a doc, it is cumbersome to have V2s and V3s with subtle changes, it is unclear which to use...

Yes, that's a very good point! I've added some documentation to `Unstable` now which hopefully makes clear when to use which of the two versions. Please, let me know if you think that piece of documentation isn't clear enough!

pitag-ha:
Btw, I've added a `unit` to the `Unstable.constructor_declaration`'s arguments to learn from our past mistakes; concretely, to allow the solution of adding an optional argument in the future, even if only inside `Unstable`. What do you think about that? If we think that's a good idea we should make kind of a "policy" out of it (which I could write down somewhere) to keep a pattern in how the functions in and outside `Unstable` look like.

Also, I've moved `Unstable` a bit up to make it easier to understand the structure of `Ast_builder` (as in, what's generated. what isn't).

panglesd:
I agree with the choice of going for modules with clear names instead of optional arguments, as the last option is less adaptable to all situations.

I do think that having`Unstable` module is better than a `V2`, but I am not sure it is the best solution, as `Unstable` is a little bit repulsive for ppx authors/maintainers. I am still a little bit more inclined toward the `From_413` solution.

Suppose the change in `4.13` AST is a killing feature that everybody want (this might not be the case, but it might be the case in future AST change). Then, the ppx author faces a dilemma: either benefiting from the feature but being unstable, or the contrary. (the ppx rewriter might even _need_ the feature and have no choice to go for the scary `Unstable`).
With the `Stable`/`Unstable` solution, the `Stable` module would be stuck with the features of the `4.12` AST, a future feature would never have the chance to be stable...

On the contrary, `From_4.13` provides a stable module with all features from `4.13`. We only provide stable modules, but still offer all features of new releases without judging whether they are important or not for ppx authors, in a consistent way.
I don't think the maintenance would be too big, as not many parts of the AST seems to change at each release (?) and especially they mostly change in disjoints parts (?). Also, having many modules is not confusing for the user, as their meaning is pretty clear from their name.

pitag-ha:
The main problem I see with the `From_414` etc approach is that folks will need to know in which OCaml version a certain feature was added in order to know the module name they'll need to choose. With the `Unstable` (or how-ever we want to call that one module) approach, folks could use the function they want to use, see its signature with merlin, and if it's missing a feature they need, they can just add an `Unstable.` to the function to get all features.

So, to sum up both what I've just said and what you've pointed out, I'd say the trade-off is the following:
- advantage of `Unstable`: more simplicity, a bit better user experience
- advantage of `From_xxx`: a little bit more stability

There are two reasons why I'm saying "a little bit" more stability instead of "more" stability: one is simply that in `Unstable`, we're adding a unit to functions with named arguments, so in the future, we can add certain features to them via optional arguments. And the more important reason is the following observation: if you need to create a node with the currently latest feature, it's likely that the reason why you need to do that is that you match over _all_ nodes of a certain kind, do something with them and construct the same node again, possibly modified. In that case, your rewriter would break anyways, even if `Ast_builder` was 100% stable. For example, in the current AST bump, I'd say that probably the only project that will end up using our `From_xxx`/`Unstable` module is `bisect_ppx` and `bisect_ppx` will break anyways if the same node will be modified in the future again, no matter how stable `Ast_builder` is.

I don't have a strong preference on this though, so if you still prefer the `From_xxx` approach, let's do that! Do you?

panglesd:
That's a nice summary, you convinced me that both solution are sensible!
I think it comes down to: will ppx authors need both to use the new AST features and have stability, in a situation where they don't break due to other reasons (such as matching on the AST)?
(as you pointed out, adding optional arguments to `ppat_construct` doesn't match the nature of the new feature, so we already have an example where `Unstable` would be, well, unstable, even with the last unit argument)

I cannot answer this question and I think it won't matter for a long time, so I'm happy with both solutions!

(As a remark, maybe it could be `Latest` with a mention that it is unstable, instead of `Unstable` with a mention that it corresponds to the latest AST? I am not sure what is more important...)

After all, I emphasize that I'm still happy with the changes as they are right now!

pitag-ha:
> will ppx authors need both to use the new AST features and have stability, in a situation where they don't break due to other reasons (such as matching on the AST)?

I think the answer to that question is that for ppx rewriters that use the newest feature and that wouldn't break due to other reasons than `Ast_builder`, it would of course be nice to have full stability in `Ast_builder`. As I've pointed out above, I don't think that happens commonly. The only situation where I see that would happen easily is the situation you mentioned: some "killer feature" gets added some time in the future.

> As a remark, maybe it could be `Latest` with a mention that it is unstable, instead of `Unstable` with a mention that it corresponds to the latest AST?

Sounds good! I've done that.

P.S.: About your sentence:
> (as you pointed out, adding optional arguments to ppat_construct doesn't match the nature of the new feature, so we already have an example where Unstable would be, well, unstable, even with the last unit argument)

Yes, and in fact, I think that's not an uncommon situation. In the last AST bump the same happened with the newly introduced `injectivity`.

panglesd (approved):

Issue #324, opened 2022-02-11
Improve documentation


pitag-ha:
## Current situation

As usual, we have different kinds of docs: a [manual](https://ppxlib.readthedocs.io/en/latest/), the [API documentation](https://v3.ocaml.org/p/ppxlib/0.24.0/doc/index.html) and a couple of very simple [examples](https://github.com/ocaml-ppx/ppxlib/tree/main/examples). Apart from that, very important resources to learn about how to write a ppx rewriter are a [blog post](https://tarides.com/blog/2019-05-09-an-introduction-to-ocaml-ppx-ecosystem) and the compiler's [parsing/parsetree.mli](https://github.com/ocaml/ocaml/blob/trunk/parsing/parsetree.mli) file.

The only resource that's kind of complete to learn how to write an extension node rewriter or a deriver is the blog post. The "manual" is more of a very high-level introduction to PPX (which is also very valuable but serves a different purpose) than a detailed manual. The API docs miss docs for some important values and aren't rendered perfectly. Apart from the docs for values/types, the API docs also include some important information about some of the modules (e.g. `Ast_builder`, `Ast_pattern`, `Ast_traverse`), which isn't easy to find and difficult to put into context.

## How to improve the situation

For the resources that make sense to co-exist separately, we should make them all easy to find:
- [ ] Port the manual to odoc so that it appears on ppxlib's ocaml.org documentation next to its API docs.
- [ ] Link to the simple examples and -once finished- to the standard derivers from within the beginning of the manual/documentation.
- [ ] Link to parsing.mli and to ast-explorer from within the manual section about the OCaml parsetree (see below).
- [ ] Link to the info about certain modules in the API docs from within the manual while putting that info into wider context.

Apart from those resources, we're currently also strongly relying on the blog post. Instead, we should integrate everything relevant in there (that is, most of it) into our manual and also write up some new bits for our manual:

- [ ]  Make section about the OCaml AST more detailed. Re-use blog post and point to the compilers' parsing.mli and to ast-explorer.
- [ ]  New section about how to construct an OCaml AST node, i.e. a about `Ast_builder` and `metaquot`. For `metaquot` we already have something in the manual.
[...]

PR #325, opened 2022-02-15
Migration of the manual into odoc

panglesd:
In #324, @pitag-ha explained the problems of the documentation of `ppxlib`, and gave a detailed plan on how to improve it.

This PR is one of the first step on this plan, and consists of an almost one-to-one translation of the manual from `restructuredText` `odoc`.

This improves a discoverability, because the API and the manual are now on the same system, with an `index.mld` presenting the library and linking all resources in one place. This homepage is also displayed in v3.ocaml.org. The interconnection between the API and manual is also improved, with many added links from the manual into the API.

There is one thing that I think should be discussed before merging this PR, to avoid too many commits with whole file rewriting: odoc currently does not allow to specify a multipage navigation bar on the left (see ocaml/odoc#479)). I see several workarounds:
   * Make the documentation a single-page document
   * Add `Previous -- Next` links at top and bottom of each page.

I would rather have a one page documentation. What do you think?


panglesd:
After discussing with @pitag-ha, we opted for a one-page manual. (exemple [here](https://choum.net/panglesd/ppxlib-odoc-monoblock/ppxlib/index.html)).

pitag-ha (approved):
Nice, thanks for this @panglesd, and sorry for the super delay in reviewing! I think it's a good improvement! A couple of minor comments:

We should also update the link in the README pointing to the manual. Can you do that? If you're busy, I can also do it myself quickly. (And we also need to rebase over `main` for the CI to pass.)

And the other thing: Given that there might be more places than our own README with a link to the old manual, it might be nice to keep the old manual with just one line: a link to the new manual. What do you think?

panglesd:
The problem with changing the link in the README, is that we will need to update the html of the API in the github pages, but there has been changes in the API since the last release (eg #333).

So we don't want to incorporate those API changes in the current API until the next release.

Either we wait for a release to update the README and the API, or I rebase over `0.26`, so that you can generate the API from since then but with the odoc manual. (is it clear? I feel unclear...)

What would you prefer?

Otherwise, I agree that I there should be a link (or better, an automatic redirection) from the old manual to the new manual.
I'll investigate if an automatic redirection is possible, or just add a link, when rebasing tomorrow!

pitag-ha:
> The problem with changing the link in the README, is that we will need to update the html of the API in the github pages, but there has been changes in the API since the last release (eg https://github.com/ocaml-ppx/ppxlib/pull/333).

Ooh, very good point! Ok, let's just wait until the next release to update the README.

> I'll investigate if an automatic redirection is possible, or just add a link, when rebasing tomorrow!

Perfect. Thanks!

panglesd:
> I'll investigate if an automatic redirection is possible

It seems https://stackoverflow.com/a/69928404/14792825 is exactly what we need.
You need to log in to the rtd interface, though, so I can't do it myself. (I'll do the "replace `index.rst` with a redirection" part)

panglesd:
I added a commit at the root of this PR, corresponding to the last version of the readthedocs manual: a version where the index is only redirecting to the homepage of the new manual.

So to merge this commit, you need to:
- add the redirect on the rtd interface
- add the first commit of this PR and make rtd fetch it and update the docs
- add the rest of the commits
- build the new doc and deploy it on the github page

If for some reason we prefer not to set up the redirect in the rtd interface, I can also turn each page (there are only 3) into a redirect in the first commit.

pitag-ha:
> You need to log in to the rtd interface, though, so I can't do it myself.

The funny thing is that I don't have access to our rtd either. That's why I was suggesting back then to simply replace our rtd docs with a link to our new odoc docs site. Let me find out if there's still anyone around with access to our rtd, since the clean-up with redirection would be far nicer.

Would it be much work to just keep both for now to be able to merge this PR? It would be very nice to have this merged...

panglesd:
Since readthedocs is reading the docs from `master` and the main branch is now called `main`, we can safely merge this. :tada:

Issue #317, opened 2022-01-18
Ppx_deriving utility functions


sim642:
Even though ppxlib is recommended for implementing new ppx derivers, ppxlib itself doesn't have a bunch of utility functions that are provided by ppx-deriving.api: https://github.com/ocaml-ppx/ppx_deriving/blob/master/src/api/ppx_deriving.cppo.mli.
In particular:
- [ ] Quoting.
- [ ] Mangling.
- [ ] Polymorphism handling.

I would say that most nontrivial ppx derivers need that functionality anyway. Currently there's two options:
1. Just depend on ppx-deriving.api in addition to ppxlib. This is done by, e.g., ppx_deriving_yojson. This is not ideal, since the API package doesn't just contain the utilities, but also ppx-deriving's own registration system, which is replaced by ppxlib. Moreover, it prevents deprecation of the API in favor of ppxlib: https://github.com/ocaml-ppx/ppx_deriving/pull/250.
2. Copy the utility functions' implementations to own ppx deriver project. This is done by, e.g., ppx_show. This avoids the dependency but involves copying code, which isn't ideal either. Moreover, fixing or optimizing those utilities is much easier in a central place (https://github.com/ocaml-ppx/ppx_deriving/pull/252).

pitag-ha:
Thanks @sim642 for opening and working on this!

Are you also working or would like to work on the missing one, i.e. polymorphism handling? I was just ask by someone what could be a good issue to work on for someone new working on `ppxlib` and this could be a fit - unless you wanted to implement it yourself of course.

sim642:
Feel free to let them do it! I won't rush to it then.

Once we fix the submodule name for all these utilities in #370, then it should be very straightforward to bring any remaining ones over from ppx_deriving. If I recall correctly, `Ppx_deriving` contains a bunch of utility functions that don't belong to any of the categories I initially listed, but might be worth bringing over nevertheless.

PR #327, opened 2022-03-20
Fix and optimize quoter

sim642:
Thinking about #317 again, I realized that quoting had already been ported to ppxlib, but I was completely unaware of it before (it was very easy to miss in the middle of all the autogenerated documentation).

This PR ports my optimization to the quoter from ppx_deriving to ppxlib: https://github.com/ocaml-ppx/ppx_deriving/pull/252.

Before doing that, I realized that the existing implementation was actually broken: it was missing application of `()` to the quoted name, so I fixed that first.
For identifier-only quoted expressions the two things somewhat cancel each other out, so the changes might be easier understood looking at the commits separately.

I guess nobody noticed it being broken for all these years because nobody is using `Ppxlib.Quoter` (also according to [Sherlocode](https://sherlocode.com/?q=Quoter)).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@ unreleased
 
 - Fix Ast_pattern.many (#333, @nojb)
 
+- Fix quoter and optimize identifier quoting (#327, @sim642)
+
 0.26.0 (21/03/2022)
 -------------------
 
--- a/src/quoter.ml
+++ b/src/quoter.ml
@@ -30,4 +30,4 @@ let quote t (e : expression) =
   in
   t.bindings <- binding :: t.bindings;
   t.next_id <- t.next_id + 1;
-  Ast.evar name
+  Ast.eapply (Ast.evar name) [ Ast.eunit ]
--- a/src/quoter.ml
+++ b/src/quoter.ml
@@ -18,16 +18,25 @@ let quote t (e : expression) =
   let loc = e.pexp_loc in
   let (module Ast) = Ast_builder.make loc in
   let name = "__" ^ Int.to_string t.next_id in
+  let binding_expr, quoted_expr =
+    match e with
+    (* Optimize identifier quoting by avoiding closure.
+       See https://github.com/ocaml-ppx/ppx_deriving/pull/252. *)
+    | { pexp_desc = Pexp_ident _; _ } -> (e, Ast.evar name)
+    | _ ->
+        let binding_expr =
+          Ast.pexp_fun Nolabel None
+            (let unit = Ast_builder.Default.Located.lident ~loc "()" in
+             Ast.ppat_construct unit None)
+            e
+        in
+        let quoted_expr = Ast.eapply (Ast.evar name) [ Ast.eunit ] in
+        (binding_expr, quoted_expr)
+  in
   let binding =
     let pat = Ast.pvar name in
-    let expr =
-      Ast.pexp_fun Nolabel None
-        (let unit = Ast_builder.Default.Located.lident ~loc "()" in
-         Ast.ppat_construct unit None)
-        e
-    in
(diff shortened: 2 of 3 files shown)


sim642:
I'm not sure what to make of this failure diff with OCaml 4.14:
```diff
-            {Ppxlib__.Import.pos_fname =
-              "_no"... (* string length 6; truncated *);
-             pos_lnum = 1; pos_bol = 0; pos_cnum = -1};
+            {Ppxlib__.Import.pos_fname = "_none_"; pos_lnum = 1; pos_bol = 0;
+             pos_cnum = -1};
```
Structurally it's the same thing, but the testing is fragile w.r.t. unrelated changes somewhere.

pitag-ha:
Thanks for the PR, @sim642! I'll have time to review it Thursday or Friday.

pitag-ha:
Thanks @sim642, both for the bug fix and for the optimization! Apart from the failing CI and a missing changelog entry, this looks perfect. Would you mind rebasing over `main` to see if that fixes the CI, and adding a changelog entry?

Also, I've read your conversation with gasche on the `ppx_deriving` PR, which is very interesting as well.

Btw, we should really add documentation to the quoter functions!

sim642:
> Would you mind rebasing over `main` to see if that fixes the CI, and adding a changelog entry?

Both done, but the CI still breaks for OCaml 4.14. I think the issue is that it's using toploop's own printing from the compiler instead of ppxlib's copy of `Pprintast`.

pitag-ha:
Hey @sim642! This has taken a while, but the CI should be fixed now. Your branch needs to be rebased over `main` one more time, and the test needs to be promoted. Would you mind doing that?

sim642:
Done and the CI looks happy now.

Weirdly, though, running locally in a OCaml 4.14 switch `dune runtest` gives a handful of other diffs like:
```diff
diff --git a/_build/default/test/base/test.ml b/_build/default/test/base/test.ml.corrected
index 7cf282a..77c27e5 100644
--- a/_build/default/test/base/test.ml
+++ b/_build/default/test/base/test.ml.corrected
@@ -106,17 +106,17 @@ let _ = convert_longident "Base.( land )"
 
 let _ = convert_longident "A(B)"
 [%%expect{|
-Exception: (Invalid_argument "Ppxlib.Longident.parse: \"A(B)\"")
+Exception: Invalid_argument "Ppxlib.Longident.parse: \"A(B)\"".
 |}]
 
 let _ = convert_longident "A.B(C)"
 [%%expect{|
-Exception: (Invalid_argument "Ppxlib.Longident.parse: \"A.B(C)\"")
+Exception: Invalid_argument "Ppxlib.Longident.parse: \"A.B(C)\"".
 |}]
 
 let _ = convert_longident ")"
 [%%expect{|
-Exception: (Invalid_argument "Ppxlib.Longident.parse: \")\"")
+Exception: Invalid_argument "Ppxlib.Longident.parse: \")\"".
 |}]
 
 let _ = Ppxlib.Code_path.(file_path @@ top_level ~file_path:"dir/main.ml")
```
Doesn't seem to have bothered the CI though, so I'm not sure what that's about.

pitag-ha (approved):
Perfect, thanks for this improvement, @sim642!

pitag-ha:
> Weirdly, though, running locally in a OCaml 4.14 switch dune runtest gives a handful of other diffs like:

You seem to have version `v.0.15.0` of `sexplib0` on your 4.14 switch. The problem is that `sexplib0` has re-formatted one of its exception printers in `v.0.15.0`. That's why the `ppxlib` opam file currently requires `sexplib0` to be of lower version than `v0.15.0` for `with-test` (which is why the CI installs a lower version and everything is fine on there).

PR #333, opened 2022-04-05
Fix Ast_pattern.many

### LINKED ISSUE #331: `Ast_pattern.many` buggy?
The combinator `Ast_pattern.many` has the following signature:

https://github.com/ocaml-ppx/ppxlib/blob/770080baac7184326203974599d48cc2035b525e/src/ast_pattern.mli#L160

which seems a bit fishy. Shouldn't the signature be something like

```ocaml
val many : ('a, 'c -> 'd, 'd) t -> ('a list, 'c list -> 'd, 'd) t 
```

instead?

Looking at the implementation:

https://github.com/ocaml-ppx/ppxlib/blob/770080baac7184326203974599d48cc2035b525e/src/ast_pattern.ml#L175-L176

the use of the identity `(fun x -> x)` looks out of place. Shouldn't it be something like

```ocaml
let many (T f) =
  T (fun ctx loc l k ->
    let rec aux accu = function
      | [] -> k (List.rev accu)
      | x :: xs ->
        f ctx loc x (fun x -> aux (x :: accu) xs)
    in
    aux [] l)
```

instead?

nojb:
Fixes #331


src/ast_pattern.mli
```diff
@@ -157,7 +157,7 @@ val map2' :
 
 val nil : (_ list, 'a, 'a) t
 val ( ^:: ) : ('a, 'b, 'c) t -> ('a list, 'c, 'd) t -> ('a list, 'b, 'd) t
-val many : ('a, 'b -> 'b, 'c) t -> ('a list, 'c list -> 'd, 'd) t
+val many : ('a, 'c -> 'd, 'd) t -> ('a list, 'c list -> 'd, 'd) t
```
panglesd, line 160: I'm not sure why you skipped the letter `b`, but that's not important at all

panglesd (approved):
Once you format to fix the CI, that's good to me!

nojb:
> Once you format to fix the CI, that's good to me!

Should be OK now, thanks.

panglesd:
Thanks!

PR #332, opened 2022-04-05
add some common cases to the doc of Ast_pattern

c-cube:

PR #335, opened 2022-04-08
add some common cases to the doc of Ast_pattern

pitag-ha:
I'm just re-opening #332, since I think that was a good idea. @c-cube, please let me know if that's ok with you (a simple :+1: would do). If so, thanks for improving the docs!


c-cube:
That's fine with me, yes :+1:

PR #340, opened 2022-05-30
Allow toplevel_printer attributes

pitag-ha:
The `toplevel_printer` attributes are interpreted by the toplevel/utop (see https://github.com/ocaml-community/utop/pull/269). This commit makes sure that the ppxlib driver doesn't complain when encountering them while running with attribute checks enabled.

Btw, we should rename the `Whitelisted` module to a less discriminatory name. If anyone has an idea for a better name, let me know (or, simply change it!).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@ unreleased
 
 - Fix quoter and optimize identifier quoting (#327, @sim642)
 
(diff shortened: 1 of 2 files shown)


panglesd (approved):
Looks good to me!

PR #341, opened 2022-06-01
Update language to be more inclusive

metanivek:
Replace whitelist with allowlist (following [Linux coding style](https://github.com/torvalds/linux/blob/master/Documentation/process/coding-style.rst)).

Thanks @pitag-ha for pointing out in #340 🙌

### DIFF
--- a/src/attribute.ml
+++ b/src/attribute.ml
@@ -416,12 +416,12 @@ end
 let check_attribute registrar context name =
   if
     (not
-       (Name.Whitelisted.is_whitelisted ~kind:`Attribute name.txt
+       (Name.Allowlisted.is_allowlisted ~kind:`Attribute name.txt
        || Name.ignore_checks name.txt))
     && Attribute_table.mem not_seen name
   then
-    let white_list = Name.Whitelisted.get_attribute_list () in
-    Name.Registrar.raise_errorf registrar context ~white_list
+    let allowlist = Name.Allowlisted.get_attribute_list () in
+    Name.Registrar.raise_errorf registrar context ~allowlist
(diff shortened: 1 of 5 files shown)


panglesd:
Thanks for your addition!
Ah, unfortunately when re-reviewing I've just seen that you modify the signatures in `name.mli`... I'll try to see if other codebase use parts of this...

panglesd:
From what I've seen in sherlocode, none of the package in the `opam` repository use one of the modified public function, so that's a good start. I don't know if we should check on other repos, any idea @pitag-ha?

As the changes are not only internal, you could also add a changelog entry?

metanivek:
Happy to contribute!

I just took a closer look, and (thankfully!) it appears that `Name` is considered private (https://github.com/ocaml-ppx/ppxlib/blob/main/src/ppxlib.ml#L71). I added an entry to `CHANGES.md` but happy to remove if unnecessary.

pitag-ha (approved):
Thanks for the contribution, @metanivek! It's nice to decrease discrimination in the code base step by step!

To double-check, I've just had a look at the ppx-universe, and indeed no one is using `Whitelisted` or the arguments whose name you've changed (I mean, no one is supposed to do so, but you never know...).

Since it's meant to be private, I'd remove the changelog entry, if that sounds good to both of you.

metanivek:
Thanks for double-checking -- I learned about two new things (sherlocode and ppx_universe) through this PR. 

And definitely agreed! The language is more inclusive and, in my opinion, more clear. Double win. 😄 

I've removed the changelog entry. 👍

pitag-ha:
Perfect, thanks!

PR #344, opened 2022-06-06
Annotate types of metaquot generated code.

ceastlund:
Adds type annotations to generated code from metaquot. This cleans up some existing problems there can be with metaquot:

1. Uses of metaquot often won't compile without type annotations, because nothing indicates to the compiler what type the constructors come from.
2. Trivial uses of metaquot that put an escape directly inside a quotation essentially compile away to nothing, and can thus return arbitrary types from the escape that make no sense in context.
3. Metaquot can be accidentally used for the wrong AST version. This will succeed in many cases, until a conflicting AST feature is used, at which point the type error is very confusing.

This feature adds expect tests to show what fails, then updates metaquot to fix the tests.


metaquot/ppxlib_metaquot.ml
```diff
@@ -39,41 +47,46 @@ struct
 
       method! expression e =
         match e.pexp_desc with
-        | Pexp_extension (({ txt = "e"; _ }, _) as ext) -> M.cast ext
+        | Pexp_extension (({ txt = "e"; _ }, _) as ext) ->
+            self#typed (M.cast ext) "expression"
```
pitag-ha, line 51: Just out of curiosity: these changes here aren't strictly needed for the problems you've mentioned, but add another level of general hygene, right?

pitag-ha (approved):
Thanks for improving metaquot!

Does this PR also fix the third problem you've mentioned? ("Metaquot can be accidentally used for the wrong AST version.")

pitag-ha:
Oh, I've just seen that the CI is failing. Do you know why?

metaquot/ppxlib_metaquot.ml
ceastlund: True, it's possible they're unnecessary, now that I think of it. The position the escape is in probably makes the type obvious in every case, now that the outermost form is annotated. But I don't think these hurt much, either. It's one type annotation per escape.

ceastlund:
Older versions phrase type errors differently, so the expect tests fail.

ceastlund:
And yes, this feature should fix the third problem above. Because these expressions are now annotated with types, they can't accidentally get confused with other versions of the AST type.

PR #345, opened 2022-06-06
Expand header and footer from enclose_intf and enclose_impl

ceastlund:
This change allows `enclose_intf` and `enclose_impl` to produce code using attributes and/or extension points.


pitag-ha (approved):
Thanks!

It has called my attention that it seems like headers and footers were deliberately left out for rewriting (and now they aren't anymore). But still, I can't find a reason _why_ one would want to leave them out. So the change looks good to me!

ceastlund:
Originally, the body was expanded and then header/footer were slapped on. It was probably assumed they were fully expanded since they came from an expander, or perhaps nobody thought about it.

Later, I split out attributes from the body. To preserve existing behavior, I continued expanding those and continued _not_ expanding the header/footer.

So I think it was just a carefully preserved oversight.

PR #346, opened 2022-06-10
implementing Code_path.enclosing_module

ceastlund:
We have a ppx extension we use internally where knowing the nearest enclosing module name, including first class modules, would be handy. I've extended Code_path and its tests accordingly.

### DIFF
--- a/src/ast_traverse.ml
+++ b/src/ast_traverse.ml
@@ -103,7 +103,7 @@ class map_with_expansion_context =
             let name = self#loc (self#option self#string) ctxt name in
             let module_expr =
               self#module_expr
-                (ec_enter_module_opt ~loc:name.loc name.txt ctxt)
+                (ec_enter_module_opt ~loc:module_expr.pmod_loc name.txt ctxt)
                 module_expr
             in
             let body = self#expression ctxt body in
--- a/src/ast_traverse.ml
+++ b/src/ast_traverse.ml
@@ -94,8 +94,23 @@ class map_with_expansion_context =
   object (self)
     inherit [Expansion_context.Base.t] map_with_context as super
 
-    method! expression ctxt expr =
-      super#expression (Expansion_context.Base.enter_expr ctxt) expr
+    method! expression ctxt { pexp_desc; pexp_loc; pexp_loc_stack; pexp_attributes } =
+      let ctxt = Expansion_context.Base.enter_expr ctxt in
+      let pexp_desc =
+        match pexp_desc with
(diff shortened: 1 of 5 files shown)


src/ast_traverse.ml
```diff
@@ -94,8 +94,26 @@ class map_with_expansion_context =
...
+    method! expression ctxt
+        { pexp_desc; pexp_loc; pexp_loc_stack; pexp_attributes } =
+      let ctxt = Expansion_context.Base.enter_expr ctxt in
+      let pexp_desc =
+        match pexp_desc with
+        | Pexp_letmodule (name, module_expr, body) ->
+            let name = self#loc (self#option self#string) ctxt name in
+            let module_expr =
+              self#module_expr
+                (ec_enter_module_opt ~loc:name.loc name.txt ctxt)
```
pitag-ha, line 106: Why are you using `name.loc` here instead of `module_expr.pmod_loc`?

```suggestion
                (ec_enter_module_opt ~loc:module_expr.pmod_loc name.txt ctxt)
```

test/code_path/test.ml
```diff
@@ -47,5 +60,39 @@ let module M = struct
...
+    ;;
+
+    ignore x
+  end
+end
+let _ = let module M = Functor() in !M.code_path
+[%%expect{|
+module Functor : functor () -> sig val code_path : string ref end
+- : string =
+"(code_path(main_module_name Test)(submodule_path(Functor _))(enclosing_module First_class)(value(x))(fully_qualified_path Test.Functor._.x))"
```
pitag-ha, line 97: This is interesting, right? So, if you store a code_path in a reference that is updated inside submodules/first class modules, the value of that reference will depend on the order in which the submodules/first class modules are compiled.

pitag-ha: Btw, on a different note: is it expected that `fully_qualified_path` is `Test.Functor._.x` here?

pitag-ha (approved):
Thanks for adding this! Looks good to me.

test/code_path/test.ml
ceastlund: > This is interesting, right? So, if you store a code_path in a reference that is updated inside submodules/first class modules, the value of that reference will depend on the order in which the submodules/first class modules are compiled.

I don't think that's right. The reference is only updated at runtime.

ceastlund: > Btw, on a different note: is it expected that `fully_qualified_path` is `Test.Functor._.x` here?

Yes, that's what I expected it to be based on existing tests and documentation.

PR #348, opened 2022-06-10
Add 5.0 support

pitag-ha:
This should make ppxlib compatible with OCaml 5.0.

The AST hasn't changed since 4.14. So I've just added direct identities between 4.14 and 5.0 instead of doing the deep migrations we usually do. I've also considered not adding a 5.0 module to Astlib at all (similarly to what I did in #319), but that would have gotten quite hacky since the compiler has bumped the magic numbers in the end.

This PR also directly bumps our ppxlib AST to 5.0, since that doesn't really change anything.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 unreleased
 ------------
 
+- Add support for OCaml 5.0 (#348, @pitag-ha)
+
 - Add `Code_path.enclosing_value` (#349, @ceastlund)
 
 - Add `Code_path.enclosing_module` (#346, @ceastlund)
--- a/ast/ast_helper_lite.ml
+++ b/ast/ast_helper_lite.ml
@@ -17,7 +17,7 @@
 open Stdlib0
 module Location = Astlib.Location
 module Longident = Astlib.Longident
-open Astlib.Ast_414
+open Astlib.Ast_500
 
 [@@@warning "-9"]
 
--- a/ast/ast_helper_lite.mli
+++ b/ast/ast_helper_lite.mli
@@ -15,7 +15,7 @@
 
 (** Copy of Ast_helper from OCaml 4.14 with docstring related stuff removed *)
 
-open Astlib.Ast_414
+open Astlib.Ast_500
 open Asttypes
 open Parsetree
 
--- a/ast/import.ml
+++ b/ast/import.ml
@@ -4,7 +4,7 @@
    It must be opened in all modules, especially the ones coming from the compiler.
 *)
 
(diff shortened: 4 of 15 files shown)


ceastlund (approved):

astlib/dune
```diff
@@ -2,7 +2,7 @@
  (name astlib)
  (public_name ppxlib.astlib)
  (libraries ocaml-compiler-libs.common compiler-libs.common)
- (flags -w -9)
+ (flags -w -9 -I +dynlink)
```
dra27, line 5: This doesn't look right for a Dune-based system - the fix ought to be adding `dynlink` to a `libraries` field somewhere, but I also didn't see this warning triggering locally?

dra27 (commented):
The dynlink fix looks odd to me (it can't do any harm, but it shouldn't need to being done this way).

Entirely drive-by comment, but is it intention to keep with the `500` format even though the release versions have dropped the double-zero (`Ast_50`, `Ast_51`, `Ast_52`, etc.)?

astlib/dune
pitag-ha: Good catch, thanks! Fun side-note: I hadn't even added the `dunlink` thing to the right dune file.

dra27: Ah, I was wondering how it had ended up there 🙂

test/expect/dune
```diff
@@ -17,6 +17,7 @@
   ;; special library that causes to record the various
   ;; libraries statically linked in with findlib so that
   ;; they are not loaded dynamically at runtime
-  findlib.dynload))
+  findlib.dynload
+  dynlink))
```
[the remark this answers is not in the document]
dra27: Oh no, it's not a bug in dune - I think it's because you're using ocamlfind 1.9.3?

pitag-ha:
> Is the printing of the version number entirely internal?

Good point! It isn't: it could theoretically be accessed via `Ppxlib.Selected_ast.string_version`. So I'll change that to single-zero.

Btw, also `Ppxlib.Selected_ast.version` could be accessed and would currently yield  `500`. However, tbf, I think it's pretty clear that `500` is just a ppxlib representation. So I'd change `string_version` to single-zero and leave `version` with its double-zero representation to keep a well-ordered file structure.

test/expect/dune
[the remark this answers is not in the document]
pitag-ha: Thanks for fixing this in dune, @dra27! I've gotten rid of my commit adding `dynlink` to the dune file here.

PR #350, opened 2022-06-13
Fix the CI

pitag-ha:
I'm just disabling some tests, that were recently introduced in #344 and #346, on old compilers in the attempt to make the CI work again.

### DIFF
--- a/test/code_path/dune
+++ b/test/code_path/dune
@@ -1,7 +1,7 @@
 (rule
  (alias runtest)
  (enabled_if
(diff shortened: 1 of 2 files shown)


pitag-ha:
hey @ceastlund, let me know if it's ok for you that I'm just disabling those tests on old compilers instead of trying to make them cross-compiler compatible.

ceastlund:
Yeah, I think this is fine.

PR #352, opened 2022-06-14
Code path attributes

ceastlund:
Added attributes for code path manipulation to `Ast_traverse`. This is especially useful for ppxes that may want the expanded code to nevertheless have human-readable code paths, since code paths may show up in error messages, etc.


src/context_free.ml
```diff
@@ -478,6 +478,15 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
       map_node EC.pattern pattern super#pattern x.ppat_loc base_ctxt x
 
     method! expression base_ctxt e =
+      let base_ctxt, e =
+        (* Make sure code-path attribute is applied before expanding. *)
+        match Attribute.get Ast_traverse.enter_value e with
+        | None -> (base_ctxt, e)
+        | Some { loc; txt } ->
+            ( Expansion_context.Base.enter_value ~loc txt base_ctxt,
+              Attribute.remove_seen Expression [ T Ast_traverse.enter_value ] e
```
pitag-ha, line 487: Have you seen the new `Attribute.remove_seen_res`? @panglesd has done the very nice job of making sure the ppxlib driver embeds errors instead of raising. What do you think about doing that here as well?

src/context_free.ml
```diff
@@ -562,6 +571,16 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
         x
 
     method! module_expr base_ctxt x =
+      let base_ctxt, x =
+        (* Make sure code-path attribute is applied before expanding. *)
+        match Attribute.get Ast_traverse.enter_module x with
+        | None -> (base_ctxt, x)
+        | Some { loc; txt } ->
+            ( Expansion_context.Base.enter_module ~loc txt base_ctxt,
+              Attribute.remove_seen Module_expr
```
pitag-ha, line 580: Similar as above

src/context_free.ml
```diff
@@ -478,6 +478,15 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
...
     method! expression base_ctxt e =
+      let base_ctxt, e =
+        (* Make sure code-path attribute is applied before expanding. *)
+        match Attribute.get Ast_traverse.enter_value e with
+        | None -> (base_ctxt, e)
+        | Some { loc; txt } ->
+            ( Expansion_context.Base.enter_value ~loc txt base_ctxt,
+              Attribute.remove_seen Expression [ T Ast_traverse.enter_value ] e
+            )
+      in
```
pitag-ha, line 489: What about the `do_not_enter_let_module` attribute? Is it not important for that one to be handled before expanding as well? (see my comment above: I'm not really sure why it's important to apply some of the attributes before expanding)

pitag-ha (commented):
Sounds good to me, thanks! (I'm curious now about what you're working on that you need such an elaborate code path manipulation)

I think it would be nice to add a `ast_traverse0.mli`, just for general hygiene. What do you think?

Also: Why is the last commit needed (applying the new attributes before expanding extension nodes)? Is that easy to explain? Concretely, I'm surprised that the tests you added before don't seem to be affected at all by that.

ceastlund:
The thing I'm working on is improving a ppx we have internally where `[%var_name]` resolves to a string constant of the nearest enclosing variable name, and similarly for `[%module_name]`. The attributes are to make it easier for things like `let%bind` to expand code however they want and still produce the code path that `[%var_name]` should see.

ceastlund:
And as for adding attributes before expansion, it does affect the tests. Note that I changed one of the tests in the same commit. Before that change, putting the attribute directly on an expansion node would not have worked, so I added a type annotation as an extra wrapper ast node.

pitag-ha:
> putting the attribute directly on an expansion node would not have worked, so I added a type annotation as an extra wrapper ast node.

Hahaaa, I see, nice trick! Missed that.

src/context_free.ml
ceastlund: The only time it really matters to apply the attributes before expanding is when we're actually going to expand an extension node. Only let_module expressions can have do_not_enter_let_module attributes; extension nodes can't. So there's nothing to worry about with that attribute on extension nodes about to be expanded.

ceastlund:
Does the error handling look sensible now?

panglesd:
Unfortunately, it is not yet perfect in terms of error handling...

Indeed, you are introducing raising functions inside `map_with_expansion_context` (through the use of `Attributes.get`).

This means that if at some point the AST has a duplicated attributes, the context-free pass will raise, and `merlin` will stop functioning. This is annoying if the duplicated attribute has been written by the user, but even more annoying if the duplicated attributes are generated by a ppx!

I am not sure what is the best way to solve this problem. The one I would think of (in the line of what I have done previously) is to add a "`fold_map`" counterpart to  `map_with_extension_context`, where the fold is the list of errors, and embed those errors when possible (probably in between structure/signature items). Then, `context_free.ml` would use this instead of `map_with_extension_context`.

`map_with_expansion_context` would be kept for compatibility, and implemented from the `fold_map` version, but with a mention in the doc that the `fold_map` version should be used instead, as it takes the errors into account.

It involves a little bit of work though...

ceastlund:
I've produced a version of `map_with_expansion_context` that injects `ocaml.error` extensions into its results when errors crop up. It mostly tries to produce equivalent syntax with added attributes, but sometimes there's no perfect way so it overwrites something with an extension. I made it elide the errors when `tool_name = "merlin"`. How's that look?

panglesd:
This is yet not ideal, as you have to embed the errors immediately in the node you are traversing. Your way of embedding errors generates other errors. For instance, in the case of module expression, adding a signature hides all elements of the module, generating unrelated errors. For instance, in
``` ocaml
let _ =
  (let module Ignore_me = struct
     let code_path = 1
   end in
  Ignore_me.code_path)
  [@do_not_enter_module] [@do_not_enter_module]
```
There is an error saying that `Ignore_me.code_path` is unbound.

I am working on a PR which will makes it easier to generate errors to be embedded later in the AST traversal. If you can wait a little bit, and rebase on it, that would be nice.

Another possibility is to merge your PR as is, and I'll fix the error reporting in my PR.

ceastlund:
Rebased on top of the error handling changes, how's it look now?

PR #354, opened 2022-06-15
update README following new documentation location

panglesd:
Everything is in the title!
Now that there is an intro page in the documentation aggregating all documentation resources (API and manual, maybe later tutorial/examples?), I don't think it is needed to link specifically to the different parts. So all doc links in the README link to the intro page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,8 +7,7 @@
 [appveyor]:       https://ci.appveyor.com/project/diml/ppxlib/branch/main
 [appveyor-img]:   https://ci.appveyor.com/api/projects/status/bogbsm33uvh083jx?svg=true
 
-[User manual][man]
-[API documentation][api-doc]
+[Ppxlib documentation][doc]
 
 # Overview
 
@@ -24,7 +23,7 @@ ways of automatically traversing values of a given type, in particular
 allowing to inject a complex structured value into generated code.
 
 For more information about ppxlib and how to use it, please consult the
-[user manual][man].
+[documentation][doc].
 
 # What is the relation between ppxlib and other ppx libraries?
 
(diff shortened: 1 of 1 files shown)


pitag-ha (approved):
Perfect, thanks!

Just one thought: what do you think about keeping a reference for the manual (i.e. a link [manual](https://ocaml-ppx.github.io/ppxlib/ppxlib/manual.html)) somewhere in the README? Either at the beginning or at "For more information about ppxlib and how to use it..."? (that might make it clearer for people that the manual has not disappeared, it has just moved)

Btw, just for future-reference: I think it would also be quite helpful to have entry points to our API in the README. Something along the lines: for more information on how to write an extension node rewriter or a deriver (with reference to the manual what that is), have a look at the [context-free rules API](https://ocaml-ppx.github.io/ppxlib/ppxlib/Ppxlib/Context_free/Rule/index.html); to build AST nodes, there are three options: `Ast`, `Ast_builder` and `metaquot` (with links); etc. What do you think? I could add that to our issue about improving the documentation, if you agree. (the other option would be to have that inside the manual btw, but I think it would be nice to have it directly in the README).

panglesd:
I think the sentences such as:
> for more information on how to write an extension node rewriter or a deriver, have a look at the [context-free rules API](https://ocaml-ppx.github.io/ppxlib/ppxlib/Ppxlib/Context_free/Rule/index.html); to build AST nodes, there are three options: Ast, Ast_builder and metaquot

should be in the doc entry page rather than in the README. It is not clear if someone who want information on how to write an extension node rewriter should look at the API or the doc. I think the entry page would solve this by describing each section (what each module of the API do, what each section of the manual describe, what each tutorial or example do...) and then the user picks what they prefer.

I have change from (API, Manual) to (ppxlib documentation) to make it clear that it is not that API or Manual that have been removed, but included as part of a bigger thing: documentation.

PR #355, opened 2022-06-17
Add suport for OCaml 5.0 on ppxlib.0.25

pitag-ha:
The AST hasn't changed from 4.14 to 5.0.

Signed-off-by: Sonja Heinze <sonjaleaheinze@gmail.com>

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,9 @@
+unreleased
+------------
+
+- Add support for OCaml 5.0 (#355, @pitag-ha)
+
+
 0.25.0 (03/03/2022)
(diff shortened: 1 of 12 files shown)


pitag-ha:
I'm just backporting this to 0.25 (skipping 0.26 since that was the release in which we bumped the AST, so some PPXs might still be behind on that). Actually, I'd rather work on the assumption that the PPX ecosystem is up-to-date with ppxlib (unless there's some blocker for that) instead of backporting important commits. However, I think in this case it's ok to make an exception to make sure everyone can try 5.0!

PR #357, opened 2022-07-05
Folding errors when AST traversing, to embed them at the toplovel structure/signature.

panglesd:
Currently, when an error occurs during an AST traversal, it has to be embedded directly.

This is currently not a very big problem, since an error can happen only when rewriting an extension point, which is a very indicated place to embed the error directly.

However, there are few problems with this approach:
- First, only one error can be embedded, while several errors could happen: This happens [here](https://github.com/ocaml-ppx/ppxlib/blob/main/src/context_free.ml#L235), where only the first error is embeded, the rest are discarded.
- If we modify the ast traversal in a way that errors can happen in other nodes, where embedding even one error is not trivial, there is no way around. This happened in #352.

This PR implements another approach, where the AST traversal is a `fold_map`, where the fold is the list of errors.

The errors are then embedded by the driver at the beginning of the structure/signature.

Another advantages of this approach is that it opens the possibility for rewriters to output both a rewritten node and a list of "non fatal errors" that the driver will embed at the toplevel structure/signature, something currently impossible with context-free rewriters.

The code is ready to review, I still need to add a few tests, but I wanted to open the PR now to have feedback on the idea.

### DIFF
--- a/src/context_free.ml
+++ b/src/context_free.ml
@@ -196,60 +196,66 @@ module Generated_code_hook = struct
     | _ -> t.f context { loc with loc_start = loc.loc_end } x
 end
 
-let rec map_node_rec context ts super_call loc base_ctxt x =
-  let open Result in
+let rec map_node_rec context ts super_call loc base_ctxt x errors =
   let ctxt =
     Expansion_context.Extension.make ~extension_point_loc:loc ~base:base_ctxt ()
   in
   match EC.get_extension context x with
-  | None -> Ok (super_call base_ctxt x)
+  | None -> super_call base_ctxt x errors
   | Some (ext, attrs) -> (
-      E.For_context.convert_res ts ~ctxt ext >>= fun converted ->
-      match converted with
-      | None -> Ok (super_call base_ctxt x)
-      | Some x ->
-          EC.merge_attributes_res context x attrs >>= fun x ->
-          map_node_rec context ts super_call loc base_ctxt x)
+      match E.For_context.convert_res ts ~ctxt ext with
+      | Error e -> (x, NonEmptyList.to_list e @ errors)
+      | Ok converted -> (
+          match converted with
+          | None -> super_call base_ctxt x errors
+          | Some x -> (
+              match EC.merge_attributes_res context x attrs with
+              | Ok x ->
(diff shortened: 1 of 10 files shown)


src/context_free.mli
```diff
@@ -150,3 +150,11 @@ class map_top_down :
        Generated_code_hook.t (* default: Generated_code_hook.nop *)
   -> Rule.t list
   -> Ast_traverse.map_with_expansion_context
+
+class map_top_down_with_errors :
```
ceastlund, line 154: Why do we want to export a new `map_top_down_with_errors` instead of just changing the type of the existing class?

src/context_free.ml
```diff
@@ -196,60 +196,66 @@ module Generated_code_hook = struct
     | _ -> t.f context { loc with loc_start = loc.loc_end } x
 end
 
-let rec map_node_rec context ts super_call loc base_ctxt x =
-  let open Result in
```
ceastlund, line 200: Might we simplify some of this code, and make it harder to accidentally drop some errors on the floor, if we made a `With_errors` module and use it in place of `Result`? Then we could use `return` and `>>=` and make most of the error handling implicit, rather than passing around errors explicitly.

src/context_free.mli
panglesd: I think the idea is to not introduce any breaking changes (hence the presence of a "raising" version and a "result" version for many functions modified in #329).

I think however that even though `map_top_down` is public, it is not meant to be used by anyone else than ppxlib's driver, so it would be ok to change its signature.

ceastlund: The main concern I have is that the code is nontrivial and two copies of it might easily drift out of sync. Also it will be easier to see in review that the behavior is the same if it's just changes to the original.

src/context_free.ml
ceastlund: There are plenty of places where "errors" is used in multiple clauses, one of which has something else appended to it. It would be easy to forget to append "errors" to a new batch of errors, and not have it caught in any way.

ceastlund: Actually, I see what you mean about this not working well for a fold. I think it's fine as-is.

ceastlund:
I wrote #358 as an alternative to this, see what you think of it.

PR #360, opened 2022-07-26
Set loc_ghost=true for errors reported alongside expanded syntax.

ceastlund:
When we report errors alongside original or expanded code, the errors should not appear to be the primary syntax for that location. So we set `loc_ghost=true` for error nodes.


ceastlund:
going to combine this with some other changes in a separate PR

PR #362, opened 2022-07-27
Update a function pattern to avoid warning 68 (when enabled).

ceastlund:
Avoids pattern matching on mutable state in a function parameter.


panglesd (approved):
Seems a safe change, removing a warning.

PR #365, opened 2022-08-09
Improve docs in module Ast_pattern (fix #178)

### LINKED ISSUE #178: Ast_pattern: example in doc doesn't type
[An example in the doc](https://github.com/ocaml-ppx/ppxlib/blob/a3878ab32e493e03286bc9a8d72d27c8c2f3d178/src/ast_pattern.mli#L57) doesn't type:
```ocaml
let payload_pattern name expr =
        let open Ast_pattern in
        pstr (pstr_value nonrecursive (value_binding ~pat:(pstring __) ~expr:__) ^:: nil)
```
Here's the compiler's complaint:
```
This expression has type (value_binding, label -> expression -> 'a, 'a) t
but an expression was expected of type (value_binding list, 'b, 'c) t
Type value_binding is not compatible with type value_binding list
```
Here's what I think is a possible fix:
```ocaml
let payload_pattern name expr =
  let open Ast_pattern in
  pstr (pstr_value nonrecursive ((value_binding ~pat:(pstring __) ~expr:__) ^:: nil) ^:: nil)
```

Kakadu:
Signed-off-by: Kakadu <Kakadu@pm.me>

### DIFF
--- a/src/ast_pattern.mli
+++ b/src/ast_pattern.mli
@@ -47,7 +47,8 @@ open! Import
       let build_payload ~loc name expr =
         let (module B) = Ast_builder.make loc in
         let open B in
-        [ pstr_value Nonrecursive [value_binding ~pat:(pstring name) ~expr] ]
+        Parsetree.PStr
+          [ pstr_value Nonrecursive [value_binding ~pat:(pstring name) ~expr] ]
     ]}
 
     Constructing a first class pattern is almost as simple as replacing
--- a/src/ast_pattern.mli
+++ b/src/ast_pattern.mli
@@ -24,16 +24,28 @@ open! Import
 
     {[
       let match_payload = function
-        | PStr [ { pstr_desc =
-                    Pstr_value (Nonrecursive,
-                      [ { pvb_pat =
-                            { ppat_desc = Ppat_constant (Pconst_string (name, _, None))
-                            ; _
-                            }
-                        ; pvb_expr = e
-                        ; _ } ])
-                 ; _ } ] ->
-          (name, e)
+        | PStr
+            [
+              {
+                pstr_desc =
+                  Pstr_value
+                    ( Nonrecursive,
+                      [
+                        {
+                          pvb_pat =
+                            {
+                              ppat_desc =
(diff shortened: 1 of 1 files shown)


src/ast_pattern.mli
```diff
@@ -42,25 +45,26 @@ open! Import
 
     {[
       let build_payload ~loc name expr =
-        let (module B) = Ast_builder.with_loc loc in
+        let (module B) = Ast_builder.make loc in
         let open B in
-        pstr
-          [ pstr_value Nonrecursive (value_binding ~pat:(pstring name) ~expr) ]
+        [ pstr_value Nonrecursive [value_binding ~pat:(pstring name) ~expr] ]
```
panglesd, line 50: I don't know why the `Ast_builder` does not have a `pstr` function, but as is it is not building a `payload` but a `structure`.
One possibility is:
```suggestion
        Parsetree.PStr
          [ pstr_value Nonrecursive [value_binding ~pat:(pstring name) ~expr] ]
```
the other possibility is adding (back?) the `pstr` function to `Ast_builder` and using it in the example!

panglesd (commented):
Thanks a lot for the PR!

panglesd:
(Also, note that the formatting is failing. Can you `ocamlformat` your additions?)

panglesd (approved):
Thanks again!

pitag-ha (approved):
Thanks @Kakadu! Improvement/fixing of our docs is always highly appreciated!

As a comment that's only related to this PR on a meta-level: for this kind of code example documentation, it would probably be quite convenient to use mdx-style documentation to make sure our examples compile and stay up-to-date. @panglesd, do you happen to know if that would be easy to do?

panglesd:
> As a comment that's only related to this PR on a meta-level: for this kind of code example documentation, it would probably be quite convenient to use mdx-style documentation to make sure our examples compile and stay up-to-date. @panglesd, do you happen to know if that would be easy to do?

Sure, good idea, from my first experiments it seems very doable, I'll give it a go!

PR #368, opened 2022-08-19
Deprecate unused attributes in `Deriving.Generator`.

### LINKED ISSUE #90: Request for docs: The `?attributes` parameter in Deriving.Generator.make
The function make takes this `?attributes` argument but as far as I can tell doesn't do anything except store it. I've dug into the source and I see some commented source that would remove this attribute, but this is no longer done. Can the purpose of this argument be documented perhaps?

sim642:
Closes #90.

Adds documentation to the `make` functions in order to have a place to deprecate the argument (similarly to `Driver.register_transformation` extensions).
Also removes the corresponding unused record field and the commented-out code, which used it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 unreleased
 -------------------
 
+- Deprecate unused attributes in `Deriving.Generator` (#368, @sim642)
+
 - Remove a pattern match on mutable state in a function argument. (#362, @ceastlund)
 
 - Add code-path manipulation attributes. (#352, @ceastlund)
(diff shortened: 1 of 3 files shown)


ceastlund (approved):

panglesd (approved):
Thanks!

pitag-ha:
Thanks, LGTM! I was about to suggest just removing the argument and sending patch PRs to the PPXs we break with that. That would be both better for them and for us. However, I've just seen that quite some PPXs use that argument (even new ones - which shows even more that this PR is important), so just adding deprecating documentation sounds good.

Could you add a changelog entry and format the code?

sim642:
> Could you add a changelog entry and format the code?

Done!

PR #372, opened 2022-09-05
Adding redirection from rtd docs to new documentation

panglesd:
This adds a folder with a doc whose only purpose is to redirect to the new doc. Readthedoc will use this folder as long as it is the only one containing a `conf.py` file (which I copied its last version, before the transition to odoc).

All files that were present at the time are included to make sure that all old links still present somewhere will redirect to the new documentation.


pitag-ha:
Nice, hope this works. Thank you! 

About "hope this works": doesn't it need to be merged into `master` to be picked up by the Read the Docs pipeline? This Read the Docs problem is quite a mess...

panglesd:
Not, it seems that for some reasons rtd picks from `main`. See the build: https://readthedocs.org/projects/ppxlib/builds/18022305/ which does:

```
git checkout --force origin/main
```

pitag-ha (approved):
Thank you, @panglesd!

Read the Docs fetching from `main` explains a lot of things...

PR #376, opened 2022-09-29
Add benchmark for ppxlib driver

gridbugs:
This adds a target `make bench` which measures the time taken to run the ppxlib driver on some input files. I've added two drivers to be benchmarked:
 - one which applies ppx_sexp_conv (which is added as a vendored package)
 - one which does nothing

Each driver specifies some input files copied from MIT licensed projects. The benchmark runner invokes each driver on each of its specified inputs and measures the mean and stddev of the time it took to run each driver on each of its inputs.


tmattio:
This is awesome, thanks @gridbugs!! 🎉 

I'm curious to know if we could integrate it with [current-bench](https://github.com/ocurrent/current-bench)? (that would allow us to have benchmark reports in GitHub PRs automatically IIUC?)

Do you know how easy that would be @shakthimaan?

gridbugs:
Thanks @tmattio! I forgot to mention in the description, but integrating with current-bench is the goal for this.

gridbugs:
This fails to build on 4.04 and 4.05 because it uses standard library functions that didn't exist in those versions of the language. Also, the pinned version of ppx_sexp_conv is too new to be built by those versions of the compiler. Is it making sure that the benchmarks build on old versions of the compiler?

panglesd:
There is actually another problem with not separating the benchmark in another package: the dependencies of the benchmark and the core library are mixed. This means that:
- `yojson` is added as a dependency while it is not useful to `ppxlib` itself
- Any time we want to add a new benchmark, we might add extra dependencies.

Since `ppxlib` is meant to be the "goto" library for all ppx rewriters, I think we should avoid as much as possible adding new dependencies. So I think that we should really go for solution 3.

If you do not have that much time to give to ppxlib, I could try to implement this solution myself. In any case, thanks a lot for the work! As @pitag-ha said, it is really great result!

pitag-ha:
> the dependencies of the benchmark and the core library are mixed

To avoid that, @gridbugs has added it as a `with-test` dependency, not as a hard dependency. I myself don't have a preference between the two things: `with-test` dependency or separate opam package. I've also asked @kit-ty-kate if one of them would be better from an `opam` point of view and she's said that both are equally fine. So I'd say it's up to you, @gridbugs and @panglesd.

panglesd:
> To avoid that, @gridbugs has added it as a with-test dependency

Yes, that's right, sorry I forgot about that!

So in fact, the [reasoning](https://github.com/ocaml-community/yojson/pull/117#issuecomment-983447660) for separating benchmark from the tests in yojson was that it can be time costly (and also, depending on your definition of test, it is not exactly a test as there are no exact failing/succeeding conclusion).

In this case, if the tests run sufficiently fast and are not computation-intensive, I'm fine with any solution!

shakthimaan:
> I'm curious to know if we could integrate it with [current-bench](https://github.com/ocurrent/current-bench)?

Yes. Please follow the steps in `Enrolling a repository` section provided in the following blog post to add your project to current-bench:
https://tarides.com/blog/2021-08-26-benchmarking-ocaml-projects-with-current-bench

pitag-ha:
> Yes. Please follow the steps in Enrolling a repository section provided in the following blog post

Yes, I'll do that as soon as this PR is merged.

Btw, im currently at the MirageOs retreat and won't have much time to review this and next week. If you have time, @panglesd , that'd be very cool. If not, I'll review as soon as I'm back.

gridbugs:
I added a `ppxlib-bench` package to this repo with the extra deps needed for benchmarks. This package relies on a newer version of ocaml (4.10 rather than 4.04 due to its dependency on ppx_sexp_conv). This has resulted in ci only being run on versions of the compiler >= 4.10.

An alternative would be to move the benchmark to a separate repo. How important is it that ppxlib continue to support versions of ocaml older than 4.10?

pitag-ha:
> How important is it that ppxlib continue to support versions of ocaml older than 4.10?

Short and concrete answer to this: I'd prefer to avoid dropping compiler support up to 4.10 to add benchmarks.

Longer and more general answer to this: For the same reason @panglesd mentioned above why we never add new hard dependencies to `ppxlib`, we also tend to try to avoid dropping compiler version support for `ppxlib`: a big part of the `opam` ecosystem depends transitively on `ppxlib` and some package maintainers wouldn't be happy if they'd have to drop support for some compilers because of `ppxlib`. That being said, when there's a good reason to drop old compilers, we do so. Usually the hard lower bound is the min of the versions rescript etc use (I think that's currently still lower than 4.10.0).


> This has resulted in ci only being run on versions of the compiler >= 4.10.

As mentioned above, to fix this I think you could have a compiler >=  4.04.1 bound in the `opam` file and desable the builds of the benchmark files in the CI by adding
```
(enabled_if
  (>= %{ocaml_version} "4.10.0"))
```
to their dune files.

gridbugs:
I've updated all the benchmarking dune files so that they only build on ocaml >= 4.10, and now the tests for older version of the compiler are running again.

panglesd:
Thanks! Sorry for leading you into the wrong direction with the package thing.

I'll review the code soon!

bench/bench.ml
```diff
@@ -0,0 +1,224 @@
...
+    if Sys.file_exists absolute_path then Some absolute_path else None
+
+  let methods = [ via_procfs; via_cwd ]
+
+  let get () =
+    let maybe_path = List.find_map (fun m -> m ()) methods in
+    match maybe_path with
+    | Some path -> path
+    | None -> failwith "couldn't determine the path to the current exe"
+end
```
panglesd, line 198: This code is just to deal with the case where bench is called somewhere else than the root of the project. That's a nice attention to details :+1:

panglesd (approved):
Looks good to me!

I made a suggestion for a more human-readable output for local benchmarking, otherwise the code, and benchmark organization, looks great! It will be easy to add new benchmarks.

gridbugs:
> Thanks! Sorry for leading you into the wrong direction with the package thing.
> 
> I'll review the code soon!

All good! I still ended up making a ppx-bench package to handle extra dependencies required only for benchmarking.

bench/bench.ml
gridbugs: Thanks!

ppxlib-bench.opam
```diff
@@ -0,0 +1,36 @@
...
+authors: ["Jane Street Group, LLC <opensource@janestreet.com>"]
+license: "MIT"
+homepage: "https://github.com/ocaml-ppx/ppxlib"
+doc: "https://ocaml-ppx.github.io/ppxlib/"
+bug-reports: "https://github.com/ocaml-ppx/ppxlib/issues"
+depends: [
+  "dune" {>= "2.7"}
+  "ocaml" {>= "4.04.1"}
+  "ppxlib" {= version}
+  "base"
```
panglesd, line 18: Could you add this for consistency with the `dune-project` file?

```suggestion
  "base" {>= "v0.15" & < "v0.16"}
```

gridbugs: I ended up removing the versions from `dune-project` instead because they were causing CI to limit the versions of ocaml used for testing to just those compatible with that range of versions.

panglesd:
Thanks, @gridbugs! Looking forward to integrating this with `current-bench`.

panglesd:
I tried to enrol this repository in ocaml-benchmark, but "This action must be performed by an organization owner".

Also, there are no public people in the `ocaml-ppx` organization, so I don't know who to ping!

pitag-ha:
I already enrolled it on Monday: https://autumn.ocamllabs.io/ocaml-ppx/ppxlib . I think anyone of us could have done it (we should have the same permissions).

There seems to be a bug, though: notice the "No data for selected interval.". We think it's on the `current-bench` side and @ElectreAAS is already kindly looking into it.

kit-ty-kate:
Hi, this change broke all most of the packages relying on mirage. Is there a particular reason why ppx_sexp_conv is vendored? Why not pulling it up as a normal opam dependency of the ppxlib-bench package?

panglesd:
Ouch, sorry for the mess!

I can't remember or find any reason to vendor instead of pulling it as normal dependency. Maybe it was needed before the `ppxlib-bench` package was created? @gridbugs do you remember?

Without reason to do otherwise, I'll open a PR removing the vendored bench dependency, and make a `0.29.1` release quickly.

ceastlund:
Last I spoke to @pitag-ha about it, it was vendored in order to have more control over its versioning. I don't know any deeper than that what the reasoning was, perhaps Sonja remembers.

kit-ty-kate:
I may understand why it was done this way, after trying to patch this. `ppx_sexp_conv` is used as the main user of ppxlib for the benchmark. So you need it to be compiled using ppxlib inside the project. Installing it from opam would use the latest stable instead (and would make the bench target unbuildable)

I'm not sure what's the best way to fix that. So far I'm leaning into reduce the vendor by removing everything that's not used and removing the public libraries, which should fix the issues when vendoring ppxlib itself. I'm having an issue with the extender which i can't get to compile for some reason tho at the moment so i haven't opened a PR yet.

pitag-ha:
Arrg. Thanks for pointing this out, @kit-ty-kate!

We can choose a different PPX for benchmarking and vendor that one (and get rid of `ppx_sexp_conv`). It doesn't need to be `ppx_sexp_conv`. @kit-ty-kate, do you have an idea/suggestion which PPX we could vendor without breaking anyone on `opam`? If not, one of us will have a look through the PPXs to find a good fit. In case there's none (although I'd be surprised), we'll write our custom benchmark PPX.

gridbugs:
If I remember correctly I chose to vendor `ppx_sexp_conv` because I didn't want it to be possible to change the version of `ppx_sexp_conv ` used in the benchmark experiments as this could impact the results of the benchmarks. Apologies for the breakage! Thanks all for sorting it out.

@kit-ty-kate mind sharing why vendoring `ppx_sex_conv` caused such a breakage? Are there potential dangers of vendoring packages?

kit-ty-kate:
@gridbugs i wrote about it in the description of https://github.com/ocaml-ppx/ppxlib/pull/386. If you need more specific information I can add it there

gridbugs:
Excellent, thanks @kit-ty-kate!

PR #379, opened 2022-10-06
Fix tests for 0.28.0 release

pitag-ha:
There were two independent problems with our tests when running them on the opam-repo CI for the [0.28.0 release](https://github.com/ocaml/opam-repository/pull/22216):
- on archlinux, `egrep` has been deprecated. so `grep -E` has to be used instead
- opam-repo CI now also runs the tests on MacOs. so our test using a binary that was created on Linux has to be disabled on MacOs


panglesd (approved):
Thanks, looks good to me!

Issue #377, opened 2022-10-03
Can't get the current path when using ~impl or ~intf


davesnx:
Hi, I faced an issue that isn't a huge deal but rather something that I would like to report.

`Ppxlib.File_path.get_default_path` doesn't follow the recommendation from Location: https://v2.ocaml.org/releases/4.05/htmlman/libref/Location.html

```
Note on the use of Lexing.position in this module. If pos_fname = "", then use !input_name instead.
```

One of the current workarounds is to use compiler-libs.common and use Location.input_name as a fallback. 

Would it make sense to have it inside ppxlib?

pitag-ha:
> One of the current workarounds is to use compiler-libs.common and use Location.input_name as a fallback.

This is definitely something we should avoid. Thanks for opening the issue, @davesnx!

I have a couple of questions: I'm curious why it's recommended to use `!input_name` when `pos_fname = ""`. Or, more concretely, what are the situations in which `pos_fname = ""`, but ` !input_name=<something interesting>`. Can you give an example in which this has bitten you? Also, have you seen the difference in [`Expansion_context`](https://ocaml.org/p/ppxlib/0.28.0/doc/Ppxlib/Expansion_context/Base/index.html) between `input_name` on one hand and the `file_name` component of `code_path` on the other hand? Would using `input_name` solve the problem in your concrete situation?

davesnx:
I ported a ppx from omp to ppxlib and this trick was implemented there. Haven't struggled with the issue myself. Repo: https://github.com/ml-in-barcelona/react-jsx-ppx

> Also, have you seen the difference in [Expansion_context](https://ocaml.org/p/ppxlib/0.28.0/doc/Ppxlib/Expansion_context/Base/index.html) between input_name on one hand and the file_name component of code_path on the other hand
No I haven't. 

I will try to "force" the issue to appear and report back here.

Even though I opened the PR, seems that it's something that isn't safe to add since there's a chance that this issue rarely (or never) happens. Commit comes from https://github.com/ocaml/ocaml/commit/fc701113aa68d5f787333b7f1d0f9e5d50dafc7f

pitag-ha:
Hey @davesnx, have you found out more about when the issue arises? I'm just asking because I've just gone through the open PRs and have seen your PR about this again.

[...]

Issue #179, opened 2020-09-15
Add Location.absolute_path to ppxlib


jberdine:
When working on [porting](https://github.com/johnwhitington/ppx_blob/pull/20) ppx_blob to ppxlib, it seems that adding Location.absolute_path to ppxlib would be helpful.

PR #380, opened 2022-10-07
[Fix I#377] Ensure fname isn't empty with input_name

davesnx:
Hi, it fixes https://github.com/ocaml-ppx/ppxlib/issues/377.

Installed cinaps and ocamlfind, but couldn't get the tests passing on main.
I'm unable to be sure about this diff, as well it's my first time contributing on a JS library, so all Import isn't clear if Astlib should expose input_name or If I should arrange modules differently. 

Opening this PR to solve both issues.

Side note: Now I'm touching Location maybe I should tackle this issue as well? https://github.com/ocaml-ppx/ppxlib/issues/179

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 unreleased
 ------------------
 
+- Ensure get_default_path isn't empty when loc_start.pos_fname is empty (#380, @davesnx)
 
 0.28.0 (05/10/2022)
 -------------------
--- a/astlib/location.mli
+++ b/astlib/location.mli
@@ -10,6 +10,9 @@ type t = Ocaml_common.Location.t = {
 type 'a loc = 'a Ocaml_common.Location.loc = { txt : 'a; loc : t }
 (** A located type *)
 
+val input_name : string ref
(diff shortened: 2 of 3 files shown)


pitag-ha:
Hi @davesnx, thanks for the PR and thanks for noticing!

About the formalities/the CI: The ocaml-ci failure just seems to be a matter of formatting. Could you format your code (`opam install ocamlformat.0.24.1 -y && dune build @fmt --auto-promote`)? The DCO failure is a bit strange: DCO just checks if you've signed-off your commit correctly, which you seem to have done. How have you signed-off? The CI expects you to use `-s` in `git commit`. And last, but most importantly: could you add a changelog entry, i.e. an entry to the `unreleased` section of `CHANGES.md`?

About the problem that the tests don't seem to pass locally for you: what's the error you're getting?

And about the content of the PR: most likely, it's a good idea to do this change! I just have [some question](https://github.com/ocaml-ppx/ppxlib/issues/377#issuecomment-1283917596) on your issue. And it would also be nice to expose `input_name` in `Astlib` in a different way to keep `Astlib` more easily maintainable (which you couldn't have known. I'll explain after discussing the content question on the issue).

PR #381, opened 2022-11-03
Remove File_path exports.

ceastlund:
The `File_path` submodule is tiny, and shadows the `File_path` library. So I moved its code into `Common` instead.


src/common.mli
```diff
@@ -86,6 +86,12 @@ val mk_named_sig :
 
     It will take care of giving fresh names to unnamed type parameters. *)
 
+(** Return the path used in a location. Strips leading "./", uses "" if no location. *)
+
+val get_default_path : Location.t -> string
```
panglesd, line 91: This doc comment is not associated to the value.

Without ocamlformatting:
```suggestion
(** Return the path used in a location. Strips leading "./", uses "" if no location. *)
val get_default_path : Location.t -> string
```

panglesd (commented):
On one hand, I don't like the current status: opening the `Ppxlib` module brings many values into scope that are most of the time not used; and the doc of the [`Ppxlib` module](https://ocaml-ppx.github.io/ppxlib/ppxlib/Ppxlib/index.html) is hard to understand with so many values.
Moreover, this change seems to break some PPXs: at least [this](https://github.com/janestreet/ppx_inline_test/blob/master/src/ppx_inline_test.ml#L96) and [that](https://github.com/janestreet/ppx_bench/blob/master/src/ppx_bench.ml#L52).

On the other hand, shadowing the `File_path` module isn't great when this library is used at the same time as Ppxlib.

Is there something I am missing in favour of this PR, or another solution with less drawbacks?

pitag-ha:
Hey @panglesd, 

> and the doc of the [Ppxlib module](https://ocaml-ppx.github.io/ppxlib/ppxlib/Ppxlib/index.html) is hard to understand with so many values.

That's true. I didn't think about how long/unorganized the `Ppxlib` module already is when @ceastlund asked whether this change was a reasonable suggestion and I said why not.

> Moreover, [this](https://github.com/janestreet/ppx_inline_test/blob/master/src/ppx_inline_test.ml#L96) change seems to break some PPXs: at least this and [that](https://github.com/janestreet/ppx_bench/blob/master/src/ppx_bench.ml#L52).

@ceastlund will send patch PRs to those two PPXs.

> Is there something I am missing in favour of this PR, or another solution with less drawbacks?

I don't think there's anything you're missing: unless I'm missing something as well, this PR is just meant to remove the inconvenience of shadowing the `File_path` library when opening `Ppxlib`.

Another solution with less drawbacks in my opinion could be the following:
- We can move `get_default_path` to the `Location` module
- We can totally remove `get_default_path_str` and `get_default_path_sig`. By the end of the day, they are very simple utility functions that aren't even used by anyone. And if someone wants to get the file path of a structure/signature, they could write those 4 lines by hand (or use the `file_path` component of `Expansion_context.code_path`, if they're using `map_with_context`). @ceastlund , do you happen to know why we have those to functions in the first place?

panglesd:
@Gopiandcode wrote on [discuss](https://discuss.ocaml.org/t/is-there-a-ppxlib-template-repository-somewhere-should-there-be/11177/7?u=panglesd):
> [...] I’m not sure if this is a very constructive critisism, but in my experience, I’ve found that the API for ppxlib isn’t very “discoverable”:
>
> - [...]
> - another way this arises is in the large number of modules possible to be opened at any time - it’s easy to fall into a rabbit hole of exploring the APIs modules that are completely irrelevant to most cases of writing a ppx.

I agree that removing `File_path` and its content seems to be the best option.

ceastlund:
Pushed a change and retitled the PR. Now removing `File_path` entirely.

pitag-ha (approved):
Sounds good to me. Thanks for the change! Are you going to send the two patch PRs, @ceastlund, or do you want me to do that?

pitag-ha:
Btw, just to make sure, I'll ask the `ocaml-ci` folks whether the CI failure is on the `ocaml-ci` side (which seems to be the case).

ceastlund:
I'm going to make those changes to `ppx_bench` and `ppx_inline_test` internally at janestreet, and they'll go out with our next public release.

pitag-ha:
Do you know when that release of Jane Street is going to take place? I'm asking because we usually also send patches to the public branches of the PPXs we break given that Jane Street tends to do the public releases not that often.

PR #383, opened 2022-12-20
add support for 5.1.0+trunk

avsm:
There are no AST changes yet with respect to 5.0, so this is a boilerplate change.


astlib/config/dune
```diff
@@ -0,0 +1,4 @@
+(executable
+  (libraries ocaml-version)
+  (modules gen)
+  (name gen))
```
pitag-ha, line 4: This is interesting! I didn't know the `ocaml-version` library yet. In general, I think your suggestion is a good idea since it allows the basic `dune` workflow. However, fetching the ocaml version via the `dune` variable (as we're currently doing) has the advantage of keeping our dependency graph minimal. So, even though I like the simplicity of your suggestion, I'd prefer to keep it the way it is. What do you think?

test/expect/dune
```diff
@@ -6,6 +6,7 @@
  (modes byte)
  (libraries
   unix
+  dynlink
```
pitag-ha, line 9: Do you remember why you had to add `dynlink` here? Which version of `dune` are you using?

pitag-ha (commented):
Hey @avsm, thanks for the temporary trunk support! Did you have a concrete use case in mind when adding it? I'm sure the `current-bench` team will be happy to see it (cc @shakthimaan, @ElectreAAS).

Btw, does it build for you? I'm asking because the CI currently seems to be struggling with it.

Also, do I understand correctly that the idea for this PR is that your branch with temporary trunk support will be easy to be found by others? Or would you like the PR to be merged?

Two side-notes (unless you'd like the PR to be merged, in which case one of them would be more than a side-note): The new AST should also be added to the astlib `cinaps` file. And, unrelated, given that the magic numbers haven't been bumped yet (and won't be until the very end of the release phase), it would also be possible to reuse the 5.0 AST completely instead of writing a new AST_501 module (either approach sounds good to me though).

astlib/config/dune
avsm: The problem here was (if I remember right) that a development version of the compiler has a very verbose version that wasn't parsed right.  No need to depend on ocaml_version externally -- if useful, it's just a single file you can pull into this repo.

avsm:
I'll close this one out @pitag-ha, as it solved my need a while back for a ppxlib that worked on 5.1 trunk (where I develop most of my OCaml code these days). I'll move over to #383

PR #385, opened 2023-02-14
Remove vendored dependency

panglesd:
Remove vendored `ppx_sexp_conv` and add it as normal dependency for `ppxlib-bench`.


avsm:
Yes please! The vendoring here also breaks the build of ppxlib in a bigger dune monorepo (where it'll conflict with another version of ppx_sexp_conv elsewhere).

pitag-ha:
Thanks, @panglesd, for the quick PR! To be clear: this breakage is on me.

Notice that if we treat the benchmark PPX as a normal dependency, we'll benchmark some `ppxlib` release (as opposed to its dev version). It's possible that we could find a way around that, but @kit-ty-kate has already found a different great solution to the problem: #386. It's already merged and patch released.

PR #386, opened 2023-02-14
Allow users to vendor ppxlib as-is, as well as ppx_sexp_conv in the same project

kit-ty-kate:
This happens when e.g. using MirageOS 4+

This PR fixes this issue by making the vendored package part of the ppxlib package (no longer vendored from dune's point-of-vue). This way, when vendoring ppxlib, `ppx_sexp_conv` is no longer available in the "public namespace" of library names and there is no longer a clash between ppx_sexp_conv from ppxlib and ppx_sexp_conv from another vendored directory.

Now that the ppx_sexp_conv is in the ppxlib namespace, all of its tests and so on were also triggered as part of ppxlib. This isn't desirable so I simply removed the `test` directory.

Tested successfully with mirage-skeleton (fails with 0.29.0, succeeds with this PR pinned)


pitag-ha (approved):
Thanks a lot for the fix, @kit-ty-kate! This is a very good idea to fix the problem.

PR #388, opened 2023-02-20
Add "ns" and "res" as reserved namespaces

davesnx:
Hi,

This PR adds 2 reserved namespaces "ns" and "res" which enables the possibility to use ppxlib with ReScript.

Further info: https://github.com/zth/rescript-relay/pull/372 and https://github.com/davesnx/styled-ppx/issues/308

### A little bit of context

Using ppxlib to write a ppx for ReScript is quite common (not so much for the parsetree versioning, but rather the ppxlib APIs) but I believe it could conflict with the idea of ppxlib to just be for OCaml/Reason development.

I might agree, but the change seems rather small and harmless. ppxlib can always remove the namespace in a new release and make ReScript team/community fork ppxlib or solve it their own way.

I would rather have ppxlib with this change and enjoy all new work than rather keep a fork with these 2 lines. 

### Commit references from zth/ppxlib
https://github.com/zth/ppxlib/commit/32f83395fb89693a873541298b6367449f23bc4a and https://github.com/zth/ppxlib/commit/b5b48e0bb2bc0a0fcf43ab6d99c0f98df0dda0e3

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 unreleased
 ------------------
 
+- Add "ns" and "res" as reserved namespaces(#388, @davesnx)
+
 0.29.1 (14/02/2023)
 ------------------
 
--- a/src/name.ml
+++ b/src/name.ml
@@ -119,6 +119,8 @@ module Reserved_namespaces = struct
   let () = reserve "merlin"
   let () = reserve "reason"
   let () = reserve "refmt"
+  let () = reserve "ns"
+  let () = reserve "res"
   let () = reserve "metaocaml"
   let () = reserve "ocamlformat"
 
--- a/src/name.ml
+++ b/src/name.ml
@@ -118,9 +118,9 @@ module Reserved_namespaces = struct
   let reserve ns = Hashtbl.add_exn tbl ~key:ns ~data:()
   let () = reserve "merlin"
   let () = reserve "reason"
-  let () = reserve "refmt"
-  let () = reserve "ns"
-  let () = reserve "res"
+  let () = reserve "refmt" (* reason *)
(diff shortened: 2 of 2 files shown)


ceastlund:
I take it the `res` namespace is for "ReScript", what does "ns" refer to?

davesnx:
Good question, I added by inertia and needed to check what it was. It seems like an old name from the syntax called "napskinscript". It's used in a few places from the rescript-lang/syntax: https://github.com/search?q=repo%3Arescript-lang%2Fsyntax%20%22ns.&type=code

pitag-ha:
Hi @davesnx, thanks!

Could you add a changelog entry? Apart from that, LGTM as well!

I'm also thinking that it might be useful to add a comment in the code pointing out that these two namespaces are for rescript compatibility. As @ceastlund has pointed out above, that's not immediately obvious from their name.

As a side-note:

> the idea of ppxlib to just be for OCaml/Reason development

Is that the idea? I remember explicitly adding bucklescript support some time ago: #205 (I know lot's of things have changed in the bucklescript/rescript world since back then, but am not very aware of the details)

davesnx:
Hi @pitag-ha 

- Added comments for the namespaces (added refmt as well just to make it explicit)
- Added the changelog entry
- Force pushed since I forgot to sign the commits (off-topic: is there any git config that I could enable for ppxlib? I couldn't find it)

BuckleScript support with the driver and those changes enable to work well with ReScript. Their core team is against ppxes and they discourage to use them but there's a few popular ones that are widely used. If ever the direction of ReScript is to break completly from ppxlib we can always remove those pieces (I'm happy to be that person!)

pitag-ha:
Thanks!

PR #390, opened 2023-02-24
 Fix failure of 'lift_map_with_context' in traverse by compile-time evaluation of  'fst' and 'snd'

smuenzel:
(This is not really a great solution, but I believe it captures the intent of the code)

Currently, `[@@deriving traverse]` fails on types like
```
type t = | X | Y of { a : string option; b : t; c : t }
```
because the inline record argument in `lift_map_with_context` is wrapped in a `fst (arg, x)` (see test in first commit).

The easiest fix is to evaluate `fst` and `snd` at ppx generation time, but that's not really a clean solution, so I'm making this a draft PR for now.

### DIFF
--- a/test/traverse/test.ml
+++ b/test/traverse/test.ml
@@ -82,3 +82,12 @@ class virtual iter :
     method t : t -> unit
   end
 |}]
+
+type t =
+  | X
+  | Arrow of { label : string option; domain : t; range : t }
+[@@deriving traverse]
+[%%expect{|
+Line _, characters 2-61:
+Error: This constructor expects an inlined record argument.
+|}]
--- a/test/traverse/test.ml
+++ b/test/traverse/test.ml
@@ -88,6 +88,54 @@ type t =
   | Arrow of { label : string option; domain : t; range : t }
 [@@deriving traverse]
 [%%expect{|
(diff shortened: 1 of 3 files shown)


test/traverse/test.ml
```diff
@@ -82,3 +82,60 @@ class virtual iter :
     method t : t -> unit
   end
 |}]
+
+type t =
+  | X
+  | Arrow of { label : string option; domain : t; range : t }
```
panglesd, line 88: Very small nitpick: the test could be made more "to the point".
```suggestion
type t = Inline of { a : string option; b : t }
```

panglesd (approved):
Good catch! And thanks for the fix.

The proposed solution looks good enough for me.

test/traverse/test.ml
smuenzel: Changed, thanks!

panglesd:
Could you add a changelog entry? Thanks! (I can do it if you prefer)

smuenzel:
> The last commit is missing the `Signed-off-by: Stefan Muenzel <source@s.muenzel.net>` sign... (then, it's ready to be merged! Very sorry for the delay!)

This is fixed now

PR #402, opened 2023-03-22
Ast_pattern: augment API with ebool, pbool helper, and a new map.

### LINKED ISSUE #399: Ast_pattern: complete API with `ebool`, `pbool` helper, and a new `map`.
`estring`, `echar`, `efloat`, ... helpers are created using functions generated from the [`constant`](https://ocaml.org/p/ppxlib/latest/doc/Astlib/Ast_500/Parsetree/index.html#type-constant) type.

However, this type does not contain boolean constants, such as `true` and `false`. Indeed, such constants are defined as constructors (without argument, nor uppercase), as seen using `utop -dparsetree`:

```ocaml
utop # true ;;
Ptop_def
  [
    structure_item (//toplevel//[1,0+0]..[1,0+4])
      Pstr_eval
      expression (//toplevel//[1,0+0]..[1,0+4])
        Pexp_construct "true" (//toplevel//[1,0+0]..[1,0+4])
        None
  ]
```

As a consequence, `ebool` was not added to the `Ast_pattern` API, even though it is useful (as illustrated [here](https://github.com/ocaml-ppx/ppx_deriving/pull/263/files#diff-d4f8cdab7fda4870e49b20bfe7e9282ac6c65aa4985721ad501e301d659b8d5eR315)) 

It would be useful to have such values in the API:
```ocaml
val ebool : (bool, 'a, 'b) t -> (expression, 'a, 'b) t
val pbool : (bool, 'a, 'b) t -> (pattern, 'a, 'b) t
```
While looking at this, I noticed that we are also missing two maps:
```ocaml
val map_value : f:('a -> 'b) -> ('b, 'c, 'd) t -> ('a, 'c, 'd) t
val map_value' : f:(location -> 'a -> 'b) -> ('b, 'c, 'd) t -> ('a, 'c, 'd) t
```
with which it is simple to generate `ebool` and `pbool`, and that should be added to the API as well!

Burnleydev1:
fixes #399 

- [x]  Write a function map_value
- [x] Use the `map_value` function to turn a `(bool, 'c, 'd) Ast_pattern.t` into a `(string, c, d) Ast_pattern.t`
- [x] Combine `pexp_construct`, `lident` and the value previously created to define `ebool` and `pbool`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 unreleased
 ------------------
 
+- Ast_pattern now has ebool, pbool helper, and a new map.(#402, @burnleydev1)
+
 - multiple errors are now reported in `metaquot`. (#397, @burnleydev1)
 
 - Add `Attribute.declare_with_attr_loc` (#396, @dvulakh)
--- a/src/ast_pattern.ml
+++ b/src/ast_pattern.ml
@@ -65,6 +65,7 @@ let int32 v = cst ~to_string:Int32.to_string v
 let int64 v = cst ~to_string:Int64.to_string v
 let nativeint v = cst ~to_string:Nativeint.to_string v
 let bool v = cst ~to_string:Bool.to_string v
+let ebool (T func) = T (fun ctx loc x k -> func ctx loc (bool_of_string x) k)
 
 let false_ =
   T
@@ -158,7 +159,6 @@ let alt (T f1) (T f2) =
 
 let ( ||| ) = alt
 let map (T func) ~f = T (fun ctx loc x k -> func ctx loc x (f k))
-let map_value (T func) ~f = T (fun ctx loc x k -> func ctx loc (f x) k)
 let map' (T func) ~f = T (fun ctx loc x k -> func ctx loc x (f loc k))
 let map_result (T func) ~f = T (fun ctx loc x k -> f (func ctx loc x k))
 let ( >>| ) t f = map t ~f
@@ -175,6 +175,9 @@ let map1' (T func) ~f =
 
 let map2' (T func) ~f =
   T (fun ctx loc x k -> func ctx loc x (fun a b -> k (f loc a b)))
+
+let map_value (T func) ~f = T (fun ctx loc x k -> func ctx loc (f x) k)
+let map_value' (T func) ~f = T (fun ctx loc x k -> func ctx loc (f loc x) k)
 
 let alt_option some none =
   alt (map1 some ~f:(fun x -> Some x)) (map0 none ~f:None)
--- a/src/ast_pattern.ml
+++ b/src/ast_pattern.ml
@@ -158,6 +158,7 @@ let alt (T f1) (T f2) =
 
 let ( ||| ) = alt
 let map (T func) ~f = T (fun ctx loc x k -> func ctx loc x (f k))
+let map_value (T func) ~f = T (fun ctx loc x k -> func ctx loc (f x) k)
 let map' (T func) ~f = T (fun ctx loc x k -> func ctx loc x (f loc k))
 let map_result (T func) ~f = T (fun ctx loc x k -> f (func ctx loc x k))
 let ( >>| ) t f = map t ~f
--- a/src/ast_pattern.ml
+++ b/src/ast_pattern.ml
@@ -232,9 +232,8 @@ let pint64 t = ppat_constant (const_int64 t)
 let pnativeint t = ppat_constant (const_nativeint t)
 let single_expr_payload t = pstr (pstr_eval t nil ^:: nil)
 let no_label t = cst Asttypes.Nolabel ~to_string:(fun _ -> "Nolabel") ** t
-
-let ebool t = pexp_construct (lident (bool' t) )none
-let pbool t = ppat_construct (lident (bool' t) )none
+let ebool t = pexp_construct (lident (bool' t)) none
+let pbool t = ppat_construct (lident (bool' t)) none
 
 let extension (T f1) (T f2) =
   T
--- a/src/ast_pattern.ml
+++ b/src/ast_pattern.ml
@@ -65,7 +65,14 @@ let int32 v = cst ~to_string:Int32.to_string v
 let int64 v = cst ~to_string:Int64.to_string v
 let nativeint v = cst ~to_string:Nativeint.to_string v
 let bool v = cst ~to_string:Bool.to_string v
-let ebool (T func) = T (fun ctx loc x k -> func ctx loc (bool_of_string x) k)
+
+let bool' (T func) =
+  T
+    (fun ctx loc x k ->
+      match x with
+      | "true" -> func ctx loc true k
+      | "false" -> func ctx loc false k
+      | _ -> fail loc "Bool")
 
 let false_ =
   T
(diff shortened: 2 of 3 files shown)


panglesd:
The `map_value` and `map_value'` look good!

The second bullet point is almost good: `bool_of_string` raises `Invalid_argument` when the given string is not `"true"` or `"false"`, and that is supposed to happen everytime the pattern fails to match.
You can take inspiration from the `some` function for instance, for what happens when a pattern fails to match.

You don't need to "implement" `Lident`: it is already implemented in `ast_pattern_generated.ml`. You just need to combine `pexp_construct`, `lident`, the value you've called `ebool` in this PR, and `none`, to create the `ebool` we want.

Burnleydev1:
Hello @panglesd, what if I use the `bool_of_string_opt` since it has the `option` which is what I think is what is added by the  `some` function, will it do the job, because `bool_of_string_opt` says that it does this ( `Convert the given string to a boolean.
Return None if the string is not "true" or "false"`)

Burnleydev1:
But it just hit me that `ebool` is supposed to be `(bool, 'a, 'b) t -> (expression, 'c, 'd) t` and not ` (bool option, 'a, 'b) t -> (expression, 'c, 'd) t`

panglesd:
You could have used `bool_of_string_opt` and matched on the result to know if it has worked or not. However, this function is only available since OCaml 4.05, and ppxlib supports OCaml starting from 4.04!

However, those functions are easy to reimplement.

Burnleydev1:
Hello @panglesd,please I looked at `some` and came did something like this 
```
let bool' (T f0) =
  T
    (fun ctx loc x k ->
      match x with
      | Some x0 ->
          ctx.matched <- ctx.matched + 1;
          let k = f0 ctx loc (bool_of_string x0) k in
          k
      | _ -> fail loc "true")
```
which has the signature ` (bool, 'a, 'b) t -> (label option, 'a, 'b) t`
Please I wish to ask if I am in the right direction.
sorry for asking too many questions.

panglesd:
You don't need to take so much of the `some` function: just look at the branch where the given value is _not_ a `Some _`.

Burnleydev1:
please is there documentation on this, please I would like to read it and understand `some` and other functions.

panglesd:
```ocaml
let some (T f0) =           (* some takes as input a pattern [T f0] (supposed to match the inside of the [Some _]*)
  T
    (fun ctx loc x k ->     (* it returns a pattern which: *)
      match x with          (* match whether [x] is [Some _] or [None]: *)
      | Some x0 ->                   (* - in the [Some _] case it *)
          ctx.matched <- ctx.matched + 1;
          let k = f0 ctx loc x0 k in (*    applies the pattern given as input to the value enclosed in the [Some] *)
          k
      | _ -> fail loc "Some"         (* - otherwise, it fails with location [loc], saying that it expects a [Some] (see the definition of [fail] *))
```

Burnleydev1:
Hello @panglesd, thanks for explaining how `some` functions, I tried using the method where we explicitly contain what happens when its "true" and "false", describing what should do when fails. The string I passed in `fail` is not correct yet but I just wanted you to checkout this new definition of `bool'`

panglesd:
Yes, that's perfect! You now have a function turning a `(bool, 'c, 'd) Ast_pattern.t` into a `(string, c, d) Ast_pattern.t`, the task 2. in your first post.

Burnleydev1:
Please how do I use the map, I actually tried using it, but I decided to instead use the structure of the map_value to create the bool' directly.

Burnleydev1:
Hello @panglesd, I am trying to carry out step 2 but I'm facing some difficulties, I took inspiration from this function below since it is the only place in `Ast_pattern` when `pexp_construct` and `Lident` are used
```
let rec parse_elist (e : Parsetree.expression) acc =
  Common.assert_no_attributes e.pexp_attributes;
  match e.pexp_desc with
  | Pexp_construct ({ txt = Lident "[]"; _ }, None) -> List.rev acc
  | Pexp_construct ({ txt = Lident "::"; _ }, Some arg) -> (
      Common.assert_no_attributes arg.pexp_attributes;
      match arg.pexp_desc with
      | Pexp_tuple [ hd; tl ] -> parse_elist tl (hd :: acc)
      | _ -> fail arg.pexp_loc "list")
  | _ -> fail e.pexp_loc "list"
```
and started with something like this 
```let ebool  = Pexp_construct ({txt = Lident "true"; _} ,None) 
```
but I am getting errors like ` Some record fields are undefined: loc`
so I changed it to 
```let ebool  = Pexp_construct ({txt = Lident "true"; loc = Location.none} ,None) 
```
which gives no errors
but then `ebool` here has just the `expression_desc` signature instead of `(bool, 'a, 'b) t -> (expression, 'c, 'd) t` that is expected
I am lost on how to proceed here.

Burnleydev1:
I tried using the method from this [documentation](https://ocaml.org/p/ppxlib/latest/doc/Ppxlib/Ast_pattern/index.html#val-parse_res) as well to get something like this 
```ocaml
let ebool (parse_res : (bool, 'a, 'b) t) : (expression, 'a, 'b) t   = 
 T Pexp_construct ({txt = Lident bool'; loc = Location.none} ,None) 
```
but it did not work either.
as I got the errors `Error (warning 27): unused variable parse_res.` and 
`longident_loc * expression option -> expression_desc
This variant expression is expected to have type (expression, 'a, 'b) t
There is no constructor Pexp_construct within type t`

Burnleydev1:
Lident converts `(label, 'a, 'b) t -> (longident, 'a, 'b) t`
pexp_construct converts `(longident, 'a, 'b) t ->
(expression option, 'b, 'c) t ->
(expression, 'a, 'c) t`

and bool' converts `(bool, 'a, 'b) t -> (label, 'a, 'b) t`
ebool needs to convert `(bool, 'a, 'b) t -> (expression, 'a, 'b) t`
I am currently working on trying to link them

Burnleydev1:
Hello @panglesd, I think I finally got it, I have pushed the changes here.

panglesd:
Very nice! Thank you @Burnleydev1!

Why did you decide to include `bool'` in the `.mli` file? and why naming it like this?

Burnleydev1:
@panglesd, I noticed that other variables in the .ml had their signature in the .mli file and I named it like that because I noticed that each variable eg map has a map’ and int32 also has int32’.I don’t know why that’s the case but I just decide that bool should also have a bool
’

Burnleydev1:
And I noticed that the map’ was the map but with location

Burnleydev1:
I just saw how other vals were defined and used that as a template to define these ones but I can remove bool’ if it is not needed

Burnleydev1:
Or change it to something more conventional

Burnleydev1:
But the main reason is that `bool` already existed and I did not know a better name, I removed it from .mli and it still built without any errors, so I will push a commit for the changes

Burnleydev1:
@panglesd, could we name it `bool_l` indicating that it converts bool to label?

ceastlund:
If the existing helpers are called `<type>'`, then I think it's sensible for this feature to continue that, rather than introduce a competing idiom. I agree the idiom is not great, but I think we should fix it wholesale in a separate PR rather than make it this PR's responsibility.

panglesd (approved):
Yes, `bool'` is coherent with other naming!
For some reason I missed the `int32'`, ... convention, I only knew the `map'`, ... convention (which is for "with location" variants).

With the removal of `ebool'` from the mli, It's now perfect, and ready to be merged!

Burnleydev1:
Thanks @panglesd, @ceastlund for the review.

PR #411, opened 2023-04-04
Reflect comptibility with stable OCaml

pitag-ha:
Since the AST reader was biased towards the current OCaml version, the trunk-support branch also supports stable OCaml versions.


panglesd (approved):
Looks good!

panglesd:
Actually, I forgot to look at the CI when reviewing... You need to add `open Stdlib0` at the beginning of the `migrate_500_501.ml` and `migrate_501_500.ml` files to have those constraints.

PR #415, opened 2023-04-13
5.2 support

shym:
Building on #413, this sets up all the deep copy mechanism to track `trunk` evolution.
This should probably be cleaned up before merging:
- I didn’t know how to format long line for the copy_poly_constraint. I wasn’t sure either whether there is a specific order for the functions when I inserted `copy_poly_constraint`,
- I’m not sure about the drop of version constraints.

In CI tests, it seems to compile fine on both 5.1.0+trunk and 5.2.0+trunk packages. Test suite succeeds on 5.2.0+trunk locally.


astlib/migrate_501_502.ml
```diff
@@ -0,0 +1,1262 @@
...
+    Ast_502.Parsetree.pdira_desc = copy_directive_argument_desc pdira_desc;
+    Ast_502.Parsetree.pdira_loc = copy_location pdira_loc;
+  }
+
+and copy_directive_argument_desc :
+    Ast_501.Parsetree.directive_argument_desc ->
+    Ast_502.Parsetree.directive_argument_desc = function
+  | Ast_501.Parsetree.Pdir_string x0 -> Ast_502.Parsetree.Pdir_string x0
+  | Ast_501.Parsetree.Pdir_int (x0, x1) ->
+      Ast_502.Parsetree.Pdir_int (x0, Option.map (fun x -> x) x1)
```
panglesd, line 40: If that were not automatically generated, I would advise to remove the useless `map` :)

astlib/migrate_502_501.ml
```diff
@@ -0,0 +1,1262 @@
+module From = Ast_502
```
panglesd, line 1:
```suggestion
open Stdlib0
module From = Ast_502
```

astlib/migrate_501_502.ml
```diff
@@ -0,0 +1,1262 @@
+module From = Ast_501
```
panglesd, line 1: `Option.map` is not available on OCaml `4.04`:

```suggestion
open Stdlib0
module From = Ast_501
```

panglesd (commented):
Thanks!
You need to rebase, without the modifications of the version constraint (they have been fixed in another PR).

astlib/migrate_501_502.ml
shym: This should be applied to quite a few other `Migrate` modules then, as I created those two files by copying `astlib/migrate_501_500.ml` and hand-editing the only interesting cases.

shym:
To help with reviewing, this gives a better view of what this PR brings (that's how it was written, at least).
`sed 's/500/502/g' astlib/migrate_501_500.ml > /tmp/a; git diff /tmp/a astlib/migrate_501_502.ml` gives:
```diff
diff --git 1/tmp/a 2/astlib/migrate_501_502.ml
index 705db5022f..8f44fab63c 100644
--- 1/tmp/a
+++ 2/astlib/migrate_501_502.ml
@@ -224,47 +224,10 @@ and copy_value_binding :
        Ast_501.Parsetree.pvb_attributes;
        Ast_501.Parsetree.pvb_loc;
      } ->
-  let pvb_pat = copy_pattern pvb_pat and pvb_expr = copy_expression pvb_expr in
-  let pvb_pat, pvb_expr =
-    match pvb_constraint with
-    | None -> (pvb_pat, pvb_expr)
-    | Some { locally_abstract_univars; typ } ->
-        let typ = copy_core_type typ in
-        let typ_poly =
-          {
-            typ with
-            ptyp_attributes = [];
-            ptyp_desc =
-              Ast_502.Parsetree.Ptyp_poly (locally_abstract_univars, typ);
-          }
-        in
-        let pvb_pat =
-          {
-            pvb_pat with
-            ppat_attributes = [];
-            ppat_desc = Ast_502.Parsetree.Ppat_constraint (pvb_pat, typ_poly);
-          }
-        and pvb_expr =
-          List.fold_left
-            (fun expr var ->
-              {
-                expr with
-                pexp_attributes = [];
-                Ast_502.Parsetree.pexp_desc =
-                  Ast_502.Parsetree.Pexp_newtype (var, expr);
-              })
-            {
-              pvb_expr with
-              pexp_attributes = [];
-              pexp_desc = Pexp_constraint (pvb_expr, typ);
-            }
-            (List.rev locally_abstract_univars)
-        in
-        (pvb_pat, pvb_expr)
-  in
   {
-    Ast_502.Parsetree.pvb_pat;
-    Ast_502.Parsetree.pvb_expr;
+    Ast_502.Parsetree.pvb_pat = copy_pattern pvb_pat;
+    Ast_502.Parsetree.pvb_expr = copy_expression pvb_expr;
+    Ast_502.Parsetree.pvb_constraint = Option.map copy_poly_constraint pvb_constraint;
     Ast_502.Parsetree.pvb_attributes = copy_attributes pvb_attributes;
     Ast_502.Parsetree.pvb_loc = copy_location pvb_loc;
   }
@@ -334,6 +297,17 @@ and copy_pattern_desc :
   | Ast_501.Parsetree.Ppat_open (x0, x1) ->
       Ast_502.Parsetree.Ppat_open (copy_loc copy_Longident_t x0, copy_pattern x1)
 
+and copy_poly_constraint : Ast_501.Parsetree.poly_constraint -> Ast_502.Parsetree.poly_constraint
+    =
+ fun {
+        Ast_501.Parsetree.locally_abstract_univars;
+        Ast_501.Parsetree.typ;
+     } ->
+  {
+    Ast_502.Parsetree.locally_abstract_univars = List.map (copy_loc (fun x -> x)) locally_abstract_univars;
+    Ast_502.Parsetree.typ = copy_core_type typ;
+  }
+
 and copy_core_type : Ast_501.Parsetree.core_type -> Ast_502.Parsetree.core_type
     =
  fun {

```

shym:
I rebased and added the necessary `open Stdlib0` in the `astlib` modules where they were missing.
Some modules in `src` use `Option.map` without `Stdlib0`: I suppose that’s not a problem?

astlib/migrate_501_502.ml
panglesd: Yes, I've opened #417 with those modifications (sorry, I should have mentioned it here!)

shym:
I just updated this PR so that it doesn’t conflict with #417, so it opens `Stdlib0` only in the new migration modules.

pitag-ha (approved):
Ouups, I really thought we had merged this a long time ago. Sorry for that!

I was just about to get rid of the `Stdlib` qualification in `copy_location` for old compiler compatibility and, doing so, noticed that we shouldn't have the deep copy in `copy_location` at all and have gotten rid of it. With that, it all looks good to me.

@panglesd , now that this has a commit of mine, could you quickly review and then merge? @shym has provided us the diff to the last migrations (in which we should also get rid of the deep dopies in `copy_location`) above, so the review is very quick!

panglesd (approved):
Thanks @shym and sorry for the delay! (I think I was too focused on waiting for the 5.1 parsetree to stabilize and missed the fact that this PR should still be merged...)

PR #417, opened 2023-04-14
Fix trunk compatibility on 4.04 and 4.05

panglesd:
Those versions do not have an `Option.map`.

It's important to have `trunk-support` support all versions, since we are likely to merge (part of) it when `5.1` is released!

### DIFF
--- a/astlib/migrate_500_501.ml
+++ b/astlib/migrate_500_501.ml
@@ -1,3 +1,4 @@
+open Stdlib0
 module From = Ast_500
 module To = Ast_501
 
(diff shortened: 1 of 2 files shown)


panglesd:
Accidentally added the fix for inline tests in this PR instead of opening another one...

pitag-ha (approved):
Thanks for fixing this, @panglesd! The way you've fixed compatibility for 4.14 and 4.05 is uncontrersial.

A quick note about the way you've fixed the expect tests: duplicating the tests for different compilers -as opposed to making one test cross-compiler compatible- always has downsides. E.g. when running into several test failures, it isn't clear whether that's really several failures or the same one duplicated. However, given that the tests are automatically generated, we don't have the typical maintenance problems of duplication here. And, in fact, I agree that this idea considerably decreases maintenance burden: Making the expect tests more cross-compiler compatible has always been a pain. So your idea sounds good to me. Thanks for coming up with with it and implementing it!

Issue #418, opened 2023-04-16
ppxlib#trunk-support on 5.1.0~alpha1 breaks `let f: type a.` syntax


copy:
Consider:

```
let f: type a. a option -> _ = assert false
```

```
(executable
 (name test)
 (flags :standard -dsource)
 (preprocess (pps ppx_here)))
```

On OCaml 5.1.0~alpha1 it results in the following error:

```
File "dune", line 1, characters 0-82:  
1 | (executable
2 |  (name test)
3 |  (flags :standard -dsource)
4 |  (preprocess (pps ppx_here)))
[@@@ocaml.ppx.context
  {
    tool_name = "ppx_driver";
    include_dirs = [];
    load_path = [];
    open_modules = [];
    for_package = None;
    debug = false;
    use_threads = false;
    use_vmthreads = false;
    recursive_types = false;
    principal = false;
    transparent_modules = false;
    unboxed_types = false;
    unsafe_string = false;
    cookies = []
  }]
let (f : 'a . a option -> _) = fun (type a) -> (assert false : a option -> _)
File "test.ml", line 1, characters 15-16:
1 | let f: type a. a option -> _ = assert false
                   ^
Error: Unbound type constructor a
```

While on 5.0:

```
let f : type a. a option -> _ = assert false
```

(any ppx that uses ppxlib demonstrates this issue, ppx_here is just a small example)

copy:
Here's a simpler one:

```
File "test.ml", line 1, characters 4-10:
1 | let (x, y) : (int * int) = assert false
        ^^^^^^
Error: This pattern matches values of type 'a * 'b
       but a pattern was expected which matches values of type int * int
```

panglesd:
Thanks for the report. This is related to the parsetree migration (see https://github.com/ocaml-ppx/ppxlib/pull/407). I'll have a look.

panglesd:
Should be fixed by #420.

pitag-ha:
Thanks again for reporting, @copy! It's fixed now.

PR #420, opened 2023-04-17
[trunk-support] Fix migration 501 to 500

panglesd:
Fix https://github.com/ocaml-ppx/ppxlib/issues/418.

501 let-binding constraints are using constructs, we need to "varify" them when putting the constraint on the expression for the 500 representation of let-binding constraints.

I think we need to add a convenient way to add tests to Parsetree migration. Debugging this without proper printing support and diff is quite tedious...


copy:
There still seem to be some issues with this, e.g.:

```ocaml
let (x, y) : (int * int) = assert false
```

and 

```ocaml
type _ foo = X : int foo | Y : int64 foo
let sizeof : type a. a foo -> _ = function
  | X -> 1
  | Y -> 8
```

copy:
Yes, this seems to work now, cheers!

Note that there's some further changes coming up in the next alpha, that might be relevant to the changes here: https://github.com/ocaml/ocaml/issues/12188#issuecomment-1511617170

anmonteiro:
Perhaps this is ocaml/ocaml#12188, but this code doesn't compile anymore:

https://github.com/janestreet/accessor_base/blob/27c1ff07b01e761ec3a034f598b81c0175b26601/src/accessor_list.ml#L115-L116

panglesd:
I don't think it compile in 5.1, unrelatedly to ppxlib (or, my test was wrong, but with `main.ml` being:

```ocaml
module Bounded_length_list : sig
  type 'a t = private 'a list
end = struct
  type 'a t = 'a list
end

type 'a t =
  | Leaf of 'a Bounded_length_list.t
  | Branch of 'a Bounded_length_list.t t

let rec to_list : type a. a t -> a list = function
  | Leaf xs -> (xs :> a list)
  | Branch t -> List.concat (to_list t :> a list list)
```
I have
```
/tmp/test_ppxlib/bin ⌚ 9:01:29 🐪 5.1.0+trunk 

$ ocaml main.ml 
File "./main.ml", line 12, characters 22-23:
12 |   | Leaf xs -> (xs :> a list)
                           ^
Error: Unbound type constructor a
```

it might be worth adding this example to issue https://github.com/ocaml/ocaml/issues/12188, to check that Octachron's https://github.com/ocaml/ocaml/pull/12191 is fixing also this issue?

panglesd:
(In any case, I'm waiting for the fix to be merged before resuming working on this PR!)

Octachron:
@panglesd  Don't hesitate to ping me if you want me to have a look at the PR once the upstream fix is merged (now that there is one more line to your table).

@anmonteiro , that was another bug: the newtype definition was no longer propagated in one case, this is fixed (in the PR).

Octachron:
The last version of the change has been merged upstream. I am thinking of releasing an alpha2 with it soonish (and probably switch to the beta once ppxlib is ready).

pitag-ha:
I've adapted this branch to the compiler parsetree change https://github.com/ocaml/ocaml/pull/12191. I've rebased the branch over `trunk-support` to handle the 502 migrations as well. I've also added a couple of tests. One test (`forth_back_migrations.t`) checks whether there's a diff between the
AST's resulting from
1. parsing <file> on 5.1.0 directly
2. parsing <file> on 5.1.0, migrating down to 5.0.0 and migrating back to 5.1.0

As expected, that's usually not the case. However, unexpectedly, that's literally never the case, even not in the example where we expect a diff:
```
let f : 'a . 'a = (fun (type a) -> (assert false : a))
```

The other test (`one_migration.t`) checks whether whether parsing on 5.0.0 (result of test running on 5.0.0) is the same as parsing on 5.1.0 and then migrating down to 5.0.0 (result of test running on 5.1.0). That test shows that we mess up the locations once in a while. The test also gives more info about what's failing in
```
let f : 'a . 'a = (fun (type a) -> (assert false : a))
```
(we somehow add two `Ptyp_poly` instead of one).

What's missing:
- [ ] Look into the locations. The diff we get on the locations might be acceptable. I haven't looked into it.
- [ ] Decide whether we care about the `let f : 'a . 'a = (fun (type a) -> (assert false : a))` example or consider it an edge case that doesn't happen.
- [ ] Add examples to the tests with attributes.
- [ ] Fix the dune file of the tests (my `enabled_if`-logic seems to be wrong)

pitag-ha:
I've just added another test to check whether the way we modify the locations is a problem. It is a problem: in several cases, our modifications don't preserve the location invariants: https://github.com/ocaml-ppx/ppxlib/pull/420/files#diff-a534d23967c350c025acd3da48f35901ef9da9f8ea5d3f079c26e5f7ee5cc462R204

@Octachron, let me give more context to ask my questions then: I'm talking about the 5.0 -> 5.1 -> 5.0 migrations. The problem here is: 5.1 contains a more compact representation than 5.0, so location information gets lost when upward migrating. In particular, in the simplest example `let x : int = 5`, the location of the whole pattern `x : int` as well as the location of the whole type node (i.e. `Ptyp_poly`), as well as the location of the whole expression constraint, get lost. In more interesting cases, such as with polymorphic types for example, the location of each nested `Pexp_newtype` expression is lost as well. Similarly, for coercion.

Do you happen to know if there's a way to recover those locations as a combination of locations we still have in 5.1.0? If not: Are those locations actually relevant for the compiler/merlin? If not, I'm inclined to make them ghost locations. If yes, th only idea I have is to store their locations in a value binding attribute when upward migrating.

test/501_migrations/normal_migrations.t
```diff
@@ -0,0 +1,53 @@
...
+  $ ./compare_on.exe file.ml ./identity_driver.exe
+
+  $ echo "let f: type a. a option -> _ = assert false" > file.ml
+  $ ./compare_on.exe file.ml ./identity_driver.exe
+
+  $ echo "let f : 'a . 'a = (fun (type a) -> (assert false : a))" > file.ml
+  $ ./compare_on.exe file.ml ./identity_driver.exe
+
+Here we expect a diff (downwards migrating should yield the same as in the example right above).
+However, something is wrong.
```
Octachron, line 36: The tests are in the wrong direction: we expect that the round trip `r = 5.1 ⇒ 5.0 ⇒ 5.1` is a normalization.

Consequently, for every syntactic category `s` for which the down migration `down` is not onto, the reverse `up` migration should pick a best representative among the pre-images of `s` by `down` .

Since in our case, the best representative is the new syntactic construct for `let pat: constraint = exp`, we should expect a diff whenever we transform a non-normalized 5.1 syntactic form into a normalized form.

In other words, the diff should appear for
```
   $ echo "let ((x, y) : (int * int)) = assert false" > file.ml
   $ ./compare_on.exe file.ml ./identity_driver.exe


   $ echo 'let (x : [`A | `B]) = (`A :> [`A | `B])' > file.ml
   $ ./compare_on.exe file.ml ./identity_driver.exe

   $ echo 'let (x : [`A]) = (`A : [`A] :> [< `A | `B])'  > file.ml
   $ ./compare_on.exe file.ml ./identity_driver.exe
```
because the those syntactic forms and the new value constraint form map to the same 5.0 parsetree node, and the `up` transform will normalize then into the new syntactic form.

test/501_migrations/normal_migrations.t
```diff
@@ -0,0 +1,53 @@
...
+  $ echo "let f : type a . a = assert false" > file.ml
+  $ ./compare_on.exe file.ml ./identity_driver.exe
```
Octachron, line 38: For this case, what happens is that on the surface language the `down` transform is injective:
```ocaml
let f: type a.a = assert false
```
is mapped onto
```ocaml
let (f: 'a . 'a) = (fun (type a) -> assert false : a)
```
but the type annotation in the pattern `(f:'a . 'a)` is not allowed in the surface language.
Which means that in order to observe a diff in the roundtrip, one need to start with ppx-constructed AST in 5.1
```ocaml
let (f: 'a . 'a) = (fun (type a) -> assert false : a)
```
which will end up being normalized into
```ocaml
let f: type a. a = assert false
```
by the roundtrip. I am not sure if we want to test those cases, however?

astlib/migrate_501_500.ml
```diff
@@ -227,16 +227,72 @@ and copy_value_binding :
...
         let typ_poly =
           {
             typ with
             ptyp_attributes = [];
             ptyp_desc =
-              Ast_500.Parsetree.Ptyp_poly (locally_abstract_univars, typ);
+              Ast_500.Parsetree.Ptyp_poly
+                ( locally_abstract_univars,
+                  varify_constructors locally_abstract_univars typ );
           }
```
Octachron, line 296: This was a ghost node in 5.0.0 .

astlib/migrate_501_500.ml
```diff
@@ -262,6 +318,44 @@ and copy_value_binding :
...
         in
         (pvb_pat, pvb_expr)
+    | Some (Pvc_constraint { locally_abstract_univars = []; typ }), _ ->
+        let typ = copy_core_type typ in
+        let pvb_pat =
+          {
+            pvb_pat with
+            ppat_attributes = [];
+            ppat_desc = Ast_500.Parsetree.Ppat_constraint (pvb_pat, typ);
+          }
```
Octachron, line 328: This is a ghost node in 5.0.0 with the right end location extended to the right end location of typ.

astlib/migrate_501_500.ml
```diff
@@ -262,6 +318,44 @@ and copy_value_binding :
...
+        in
+        (pvb_pat, pvb_expr)
```
Octachron, line 330: To be completely faithful at the parsetree level, the expression part should have a constraint added to the `pvb_expr` part when `typ` is not a `Ptyp_poly` .

However, `let (x:typ) = y` and `let (x:typ) = (y:typ)` are equivalent after typechecking, so this this is only visible at the parsetree level.

astlib/migrate_501_500.ml
```diff
@@ -262,6 +318,44 @@ and copy_value_binding :
...
+    | Some (Pvc_coercion { ground; coercion }), _ ->
+      let coercion = copy_core_type coercion in
+      let typ =
+        {
+          coercion with
+          ptyp_attributes = [];
+          (* location is wrong. attrs might also be wrong. *)
+          ptyp_desc = Ast_500.Parsetree.Ptyp_poly ([], coercion);
```
Octachron, line 338: This is a ghost location (with the right location?).

astlib/migrate_501_500.ml
```diff
@@ -262,6 +318,44 @@ and copy_value_binding :
...
+        }
+      in
+      let pvb_pat =
+        {
+          pvb_pat with
+          ppat_attributes = [];
+          ppat_desc = Ast_500.Parsetree.Ppat_constraint (pvb_pat, typ);
+        }
```
Octachron, line 346: This is a ghost node, with location that start with the pattern and ends with the type.

astlib/migrate_501_500.ml
```diff
@@ -262,6 +318,44 @@ and copy_value_binding :
...
+      in
+      let ground = Option.map copy_core_type ground in
+      let pvb_expr =
+        {
+          pvb_expr with
+          pexp_attributes = [];
+          (* location is wrong. attrs might also be wrong. *)
+          pexp_desc = Ast_500.Parsetree.Pexp_coerce (pvb_expr, ground, coercion);
+        }
```
Octachron, line 355: The location used in 5.0.0 for those nodes (which are not ghosts) starts with the pattern and ends with the expression.

astlib/migrate_501_500.ml
```diff
@@ -227,16 +227,72 @@ and copy_value_binding :
      } ->
   let pvb_pat = copy_pattern pvb_pat and pvb_expr = copy_expression pvb_expr in
   let pvb_pat, pvb_expr =
-    match pvb_constraint with
-    | None -> (pvb_pat, pvb_expr)
-    | Some { locally_abstract_univars; typ } ->
+    match (pvb_constraint, pvb_pat) with
+    | ( Some (Pvc_constraint { locally_abstract_univars; typ }),
```
Octachron, line 231: This case perform a supplementary deep copy, it is better to make sure that it only happens when `locally_abstract_univars` is not empty.

astlib/migrate_501_500.ml
```diff
@@ -262,6 +318,44 @@ and copy_value_binding :
...
+      let pvb_expr =
+        {
+          pvb_expr with
+          pexp_attributes = [];
+          (* location is wrong. attrs might also be wrong. *)
+          pexp_desc = Ast_500.Parsetree.Pexp_coerce (pvb_expr, ground, coercion);
+        }
+      in
+      (pvb_pat, pvb_expr)
+    | _ -> (pvb_pat, pvb_expr)
```
Octachron, line 358: I would propose to split the two cases hidden behind this `_` wild card:
```ocaml
| Some Pvc_constraint {locally_abstract_types = _ :: _ ; _ }, _ ->
     pvb_pat, pvb_expr (* forbidden in the language *)
| None; _ -> pvb_pat, pvb_expr (* no constraints to translate *)
```

astlib/migrate_501_500.ml
Octachron: Same with pvb_pat with location ending at the end of the type below.
Contrarily, the location for the expression part are not ghosts but they all start at the beginning of the pattern (and ends on the expression).

astlib/migrate_501_500.ml
Octachron: I had missed that this case also cover the special encoding `let ident: typ ` in general. It is probably thus simpler to avoid calling `varify_constructors` when `locally_abstract_univars` are empty.

astlib/migrate_501_500.ml
```diff
@@ -262,6 +318,44 @@ and copy_value_binding :
             (List.rev locally_abstract_univars)
         in
         (pvb_pat, pvb_expr)
+    | Some (Pvc_constraint { locally_abstract_univars = []; typ }), _ ->
+        let typ = copy_core_type typ in
```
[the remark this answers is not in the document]
Octachron: Disregards the above comment, that doesn't apply for non-identifier patterns.

astlib/migrate_500_501.ml
```diff
@@ -297,7 +287,12 @@ and copy_value_binding :
              { ptyp_desc = Ptyp_poly (args_tyvars, rt) } );
        ppat_attributes = [];
       } ->
-          Some (pat, args_tyvars, rt)
+          Some (pat, args_tyvars, rt, `Var)
+      | {
+       Ast_500.Parsetree.ppat_desc = Ppat_constraint (pat, rt);
+       ppat_attributes = [];
+      } ->
+          Some (pat, [], rt, `NonVar)
```
Octachron, line 295: This change is too lax: we really want this branch to only be taken when `pat.ppat_desc` is not `Ppat_var`. Otherwise, we group the case
```ocaml
let (x,y) : int * int = ...
```
(which may have come from a value constraint or not) with 
```ocaml
let (x:int) = ...
```
which would have been desugared to `let (x: . int) = ... `  (with a `Ptyp_poly([], _)`) if it was a value constraint in 5.0.0 .

test/501_migrations/normal_migrations.t
pitag-ha: Thanks a lot for the in-depth review, @Octachron, this is super helpful for us! And also thanks for all the explanations, among others the ones here.

I'm a bit confused about the statement

> The tests are in the wrong direction: we expect that the round trip `r = 5.1 ⇒ 5.0 ⇒ 5.1` is a normalization.

What do you mean with "The tests are in the wrong direction"? The tests in this file do test the round trip ` r = 5.1 ⇒ 5.0 ⇒ 5.1`. That's the round trip, for which I expect that there might be diffs. For the other round trip ` r2 = 5.0 ⇒ 5.1 ⇒ 5.0`, I don't expect diffs (up to the location problem I've mentioned) for the following reason: `down: 5.1 ⇒ 5.0` is onto, so we can choose `up: 5.0 ⇒ 5.1` to be its right inverse: `r2 = down \circ up = id`.

Btw, that's exactly the choice we've made: we've chosen `up` to be `down`'s right inverse. Unfortunately, that means that `r` isn't a normalization: the example you're mentioning `let ((x, y) : (int * int)) = assert false` is mapped to its syntactic form in `5.1` rather than its normal form. From a practical point of view that makes a lot of sense, I think: `r2` is by far the more important round trip than `r`: `r` will only happen during the compiler's `5.1` release phase. At the end of that phase, we'll bump the Ppxlib AST to `5.1` and from that moment on only `r2` can happen (and _will_ happen for everyone using ocmal <= 5.0 and a PPX). Do you think that was a bad choice of ours?

(As a side-note: I'm assuming there's a typo in

> Consequently, for every syntactic category `s` for which the down migration `down` is not onto

Did you want to say that `down` is _not_ onto or that it's onto? If you indeed want to say _not_ onto, could you give an example in `5.0` that isn't hit?)

test/501_migrations/normal_migrations.t
pitag-ha: > on the surface language the `down` transform is injective

I've double-checked again and for me,
```
let f : 'a . 'a = (fun (type a) -> (assert false : a))
```
is accepted syntax and is parsed into the same AST in `5.0` (up to locations) as
```
let f : type a . a = assert false
```
Those two examples are parsed into different ASTs in `5.1`. Am I misunderstanding something? If not, isn't this an example showing that also on the surface language `down` is not injective?

pitag-ha:
There seems to be an ocaml-ci problem now. Apart from that, the tests should be fixed: I expect a diff only in `reverse_migrations.t` and only on `5.1.0~alpha2`, which should show the current location diffs in the `5.0.0 => 5.1.0 => 5.0.0` round trip.

test/501_migrations/normal_migrations.t
Octachron: First, yes, I agree that
``ocaml
let f : 'a . 'a = (fun (type a) -> (assert false : a))
```
and 
```ocamll
let f : type a . a = assert false
```
are parsed to the same node in 5.0 . I was mislead by the absence of diff and the problem with  ppx built parsetree exists that should ignore for now.

What's really at play here is that the `down` function is using a new encoding for the 5.1 term
```ocaml
let f : 'a . 'a = (fun (type a) -> (assert false : a))
```
which is transformed into 5.0
```ocaml
let (f: ø . ('a . 'a)) = ((fun (type a) -> (assert false : a)) : ('a . 'a))
```
because we are in the 
```
| ( Some (Pvc_constraint { locally_abstract_univars; typ }),
        { Ast_500.Parsetree.ppat_desc = Ppat_var _; ppat_attributes = [] } )
```
branch (with `locally_abstract_univars=[]`).
Then the `up` function pick this term as a pattern `let (x: ... . typ ) = (exp : typ)` and 
transform it back to 5.1
```ocaml
let f : 'a . 'a = (fun (type a) -> (assert false : a))
```
A good way to test this issue is to assert in the `up` translation that the `typ` term in  `(pat: ... . typ)` is not a `Ptyp_poly` itself with
```diff
--- a/astlib/migrate_500_501.ml
+++ b/astlib/migrate_500_501.ml
@@ -287,6 +287,7 @@ and copy_value_binding :
              { ptyp_desc = Ptyp_poly (args_tyvars, rt) } );
        ppat_attributes = [];
       } ->
+          assert (match rt.ptyp_desc with Ptyp_poly _ ->  false | _ -> true );
           Some (pat, args_tyvars, rt, `Var)
       | {
        Ast_500.Parsetree.ppat_desc = Ppat_constraint (pat, rt);
```
which currently fails on the `normal_migrations.t` test.

pitag-ha: I see, that makes a lot of sense! Are you currently fixing this as well?

Octachron: Yep, I should have a fix ready soon.

test/501_migrations/normal_migrations.t
Octachron: The two last commits on https://github.com/Octachron/ppxlib/tree/pr420_non_abstract_poly_fix fix the down and upward translation to avoid going through an invalid 5.0 parsetree. However, with the two commits we are still able to keen `r(⭨⭧)` the identity for
```ocaml
let f: 'a. 'a = (fun (type a) -> assert false:a)
```
because the test https://github.com/Octachron/ppxlib/blob/pr420_non_abstract_poly_fix/astlib/migrate_500_501.ml#L317 fails due to the difference in locations between the pattern and expression constraint.

pitag-ha:
Thank you for the fixes, @Octachron, you're saving us a lot of time here!

I've looked quickly at your fix so far, and everything seems very good to my eyes. The fix reveals another location invariants breakage  (which seems pretty much the same as the other ones):

```
  $ echo "let f : 'a . 'a = (fun (type a) -> (assert false : a))" > file.ml
  $ ./reverse_migrations.exe -check -locations-check file.ml > /dev/null
  File "file.ml", line 1, characters 4-5:
  1 | let f : 'a . 'a = (fun (type a) -> (assert false : a))
          ^
  Error: invalid output from ppx:
         this pattern is built from a core type whose location is outside of this node's.
  Child core type found at:
  File "file.ml", line 1, characters 13-15:
  [1]
```


Btw, also thanks for the new explanation in the test file! I was always thinking about the transformations "up to locations". Pretty interesting that locations can actually make a structural difference in the AST!

Octachron:
That's not that surprising when the ghost indication is missing on the pattern.

I have the the fix for the location is in the two commits in  https://github.com/Octachron/ppxlib/tree/pr420_location_fixes , beware that the first commit extend the special handling in `src/location_check.ml` to cover not only `let x :> typ = exp` but also `let x : ty :> cty = exp`.

With this location fix, the reverse migration test except for one case where there is a real loss of information between 5.01 and 5.0 : when the type constraint ends up on terminal that's not part of the AST

```
let x : ((((((int)))))) = 0
```
we cannot reconstruct the ghost pattern range `x : ((((((int))))))` because we don't have the location of the last `)` which is a part of the concrete syntax not tracked by the abstract syntax.

That's annoying but fortunately this only happens on this ghost pattern location.

pitag-ha:
This is merged in https://github.com/ocaml-ppx/ppxlib/pull/424 now. Thanks again!

pitag-ha:
Please, don't delete your branch, @panglesd.

PR #422, opened 2023-05-22
Update the opam description

pitag-ha:
Avoid confusion around the topic that ppxlib isn't only a library, but also contains the driver and metaquot.


panglesd (approved):
I'm not sure why, but "infrastructure" sounds hardware-related to me.
In this particular case, there is no possible confusion (we can't download hardware through `opam` yet) so it's not a problem.
(I personally think "framework" describes `ppxlib` best, but I've heard people thinking this word is too widely used, so I would not dare to argue for it!)

pitag-ha:
> (I personally think "framework" describes ppxlib best, but I've heard people thinking this word is too widely used, so I would not dare to argue for it!)

In fact, I was also thinking of putting "framework" before deciding to put "infrastructure". I wasn't really sure what a framework is, though, so I had a look at the [Wikipedia article about frameworks](https://en.wikipedia.org/wiki/Software_framework) yesterday. However, it's too object-oriented for me to really understand it. I think the reason why I think "framework" doesn't really fit here is that with my understanding of "framework", a framework is meant to help build one concrete type of software, e.g. web apps, phone apps, tests or similar. I also have the impression that frameworks are usually used explicitly, i.e. the user is aware they're using the framework. When beginners derive `show`, they aren't aware that `dune` is creating and calling a `ppxlib` driver when building the project. Let me know if you have a different understanding of "framework"!

Apart from "framework" and "infrastructure", I've also considered "system" or "library+tool" (this last one doesn't capture that `metaquot` and `traverse` come bundled with `ppxlib` as well, but I think that's ok). What do you think about those options?

panglesd:
From the article and your explanation, I admit I tend to still think "framework" is a good fit...

But I think infrastructure is also good, let's not nitpick too much on the choice of the term :) It's the one you end up with, and it's good, let's use it!

PR #424, opened 2023-06-12
Pr420 with clean history

pitag-ha:
This is the same as #420, but with a clean git history and with signed commits. I care about the git history here because we'll need to cherry-pick these changes over to `main` (and I didn't want to clean up the history on the #420 branch, because people are using that branch).

When cherry-picking this over to `main`, we'll need changelog entries for both the 5.1.0 support, mentioning that this is incompatible with `5.1.0~alpha1`, and for the location-check change.


panglesd (approved):
Thanks a lot @pitag-ha and @Octachron for all the hard work!

PR #425, opened 2023-06-12
Remove misleading attribute hint

ceastlund:
Fixes an odd case where ppxlib suggests you've used an attribute in the wrong context, even when the attribute is in fact declared in that context.

Previously ppxlib had reported cases where the attribute was declared in other contexts, whether or not it was also declared in the current context. In the first commit, we add a test demonstrating this behavior. In the second commit, we fix the error message and the test.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 unreleased
 ------------------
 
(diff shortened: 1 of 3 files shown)


pitag-ha:
Thanks! Could you add a changelog entry?

ceastlund:
Added. Look good?

PR #427, opened 2023-06-14
Replace Caml with Stdlib

ceastlund:
Use `Stdlib` instead of the now-unnecessary alias `Caml`.

### DIFF
--- a/stdppx/stdppx.ml
+++ b/stdppx/stdppx.ml
@@ -1,3 +1,5 @@
+module Caml = Stdlib [@@deprecated "[since 2023-06] use Stdlib instead"]
+open Stdlib
 open StdLabels
 module Sexp = Sexplib0.Sexp
 module Sexpable = Sexplib0.Sexpable
--- a/stdppx/stdppx.ml
+++ b/stdppx/stdppx.ml
@@ -1,5 +1,3 @@
-module Caml = Stdlib
-open Caml
 open StdLabels
 module Sexp = Sexplib0.Sexp
 module Sexpable = Sexplib0.Sexpable
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,8 @@ unreleased
 
 - Adopt the OCaml Code of Conduct on the repo (#426, @pitag-ha)
 
+- Replace `Caml` with `Stdlib`. (#427, @ceastlund)
+
 - Clean up misleading attribute hints when declared for proper context. (#425, @ceastlund)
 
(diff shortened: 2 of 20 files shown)


stdppx/stdppx.ml
```diff
@@ -1,5 +1,5 @@
-module Caml = Stdlib
-open Caml
+module Caml = Stdlib [@@deprecated "[since 2023-06] use Stdlib instead"]
```
pitag-ha, line 1: By default, the deprecation alert is an error, isn't it? If so, I'd wait for the creation of the `5.1.0` bump PPX universe (i.e. not the upcoming release, but the one after). What do you think?

pitag-ha (commented):
Thanks, looks probably good. However, I'd possibly wait for the `5.1.0` bump PPX universe to get a better overview over how many PPXs will be affected.

For context: What was `Caml` needed for in the first place?

ceastlund:
`Caml` existed before there was a `Stdlib` alias provided by the runtime.

stdppx/stdppx.ml
ceastlund: sounds fine

ceastlund: Are we ready for me to release this now?

pitag-ha: With "to release this" you mean "to merge this", right?

We've decided not to bump the AST to `5.1.0` and instead to wait for `5.2.0` since `5.1.0` hasn't introduced any new syntax feature that's reflected in the AST. So, if we wanted to wait for the next creation of a PPX universe, we'd have to wait for the `5.2.0` bump.

I'd say, instead let's just merge this now and be careful in the next opam release.

Just to make sure I'm not missing anything: The `Caml` module was just needed for something Jane Street internal, right? I think it got introduced when removing the dependencies on `base` and `stdio`, so it's probably related to that. Do you remember?

pitag-ha (approved):
I'd say we can just merge this now and be careful during the next opam release (see comment above).

Just to make sure I'm not missing anything: The `Caml` module was just needed for something Jane Street internal, right? I think it got introduced when removing the dependencies on `base` and `stdio`, so it's probably related to that. Do you remember?

ceastlund:
`Caml` was for being able to explicitly refer to `Stdlib` before there was a name for it. So we could shadow values from the default environment, but still have a way to explicitly refer to them.

pitag-ha:
Ah, you mean as a standard runtime module? Thanks for explaining!

PR #428, opened 2023-06-16
Add support for OCaml 5.1

tmattio:
* Update trunk AST for pvb_constraint
* Add migration from 501 to 500
* Bump the 501 AST magic numbers
* Binary AST reader: use latest possible AST version

When the magic numbers aren't unique, before this commit, the binary AST reader would read into the first AST version matching the magic number. With this commit, it takes the last such version. That's important for trunk support.

* Fix inline tests' dependency on OCaml version

Done by making the expected output also dependent on the OCaml version

---

This is cherry-picking 4ebc7d4b8c77d516676d7a71be9911fabdfcb90b to 9d95d42754f49d9f7b8f94bf402da0cb18bf8b4a. (with a small difference from 4ebc7d4b8c77d516676d7a71be9911fabdfcb90b to remove the README update)

I've squashed the commits to make the history cleaner as most of the commits seemed to be fixes of the intial commit, but I can keep the history intact if you prefer.

I'll rebase trunk-support on top of main once merged.


pitag-ha (commented):
Thanks for taking this on! There seem to be a few important commits from `trunk-support` missing (see below).

tmattio:
Thanks for the review @pitag-ha and you're right I had omitted commits that came after the support for 5.2 that fixed 5.1. I added 83a371f5d81831980cf1978bb1920c81cda37337 and 0bc7034da32ae7b7971a2e7c83d1defee80a6cd6.

So the only commits from trunk-support that are included now should be:
- b7eddd29805661d43b01d4f54ea53fc9610277c5
- b044667a7ee10f80d517108738a7ba3948f78f79
- d0e7d17738a7e1a27e4bbc0f52690830cab6aed4
- 4c456d5bb4d864295b1734b662e393fba2f5b456
- 61818cec6bf1e4f4925c1c0d1af0f89ce0a26076

Do any of these seem relevant for 5.1?

CHANGES.md
```diff
@@ -25,6 +25,8 @@ unreleased
 - Update description to reflect that `ppxlib` contains more than a library
   (#422, @pitag-ha)
 
+- Add support for OCaml 5.1 (#428, @shym, @Octachron , @pitag-ha, @panglesd)
```
pitag-ha, line 28:
```suggestion
- Add support for OCaml 5.1, excluding OCaml `5.1.0~alpha1` (#428, @shym, @Octachron , @pitag-ha, @panglesd)
```

Since it isn't common to be incompatible with alpha-releases, I think it's worth pointing that out.

CHANGES.md
```diff
@@ -25,6 +25,8 @@ unreleased
...
+
```
pitag-ha, line 29:
```suggestion
- Add support for OCaml 5.1 (#428, @shym, @Octachron , @pitag-ha, @panglesd)
- Driver: Fix `-locations-check` option for coercions with ground  (#428, @Octachron)
```

astlib/migrate_501_500.ml
```diff
@@ -0,0 +1,1405 @@
...
+        in
+        let pvb_pat = ghost_constraint pvb_pat typ in
+        let ground = Option.map copy_core_type ground in
+        let pexp_loc = merge_loc pvb_pat.ppat_loc pvb_expr.pexp_loc in
+        let pvb_expr =
+          {
+            pvb_expr with
+            pexp_attributes = [];
+            pexp_loc;
+            (* location is wrong. attrs might also be wrong. *)
```
pitag-ha, line 384:
```suggestion
```

I forgot to remove this from the draft when fixing it.

pitag-ha (commented):
Looks almost good, thanks! We just need to run and promote the tests to fix them on 5.1. Could you do that?

tmattio:
> Looks almost good, thanks! We just need to run and promote the tests to fix them on 5.1. Could you do that?

Done

pitag-ha (approved):
Thank you, looks all good (up to the still running CI)! Once the CI finishes and passes, let's merge this and then cut a release right away.

tmattio:
The cluster is having a hard time running the jobs - macOS x86 and Debian s390x have been hanging since yesterday.

Seeing all the other jobs are passing and we don't have an immediate solution for unlocking the hanging jobs, I'm merging now.

PR #439, opened 2023-06-27
Upgrade to ocamlformat 0.26.0

Julow:
Unnecessary escaping is removed from doc-comments, which has a huge impact on `ast.ml`. The indentation of attributes after a `let`/`val` is also changed, which are present in an unusual amount in ppxlib ! The class are more compact.

The aim of this commit is to gather feedback.

Changelog can be found here: https://github.com/ocaml-ppx/ocamlformat/blob/main/CHANGES.md

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -6,3 +6,5 @@
 
 #The commit upgrading to ocamlformat 0.24.1
 0970c3a7f91291bd92eb277331b5b6af20b608e9
+# Upgrade to OCamlformat 0.26.0
+4134f18de87cd5bad7a42f12b5ef43132d09f57f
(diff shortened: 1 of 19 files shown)


Julow:
OCamlformat 0.26.0 has been released! This PR can be merged (not squashed due to `.git-blame-ignore-revs`) or closed if it's a bit too early to upgrade.

Julow:
OCamlformat will be released with compatibility with 5.1 in a few weeks. Let's skip 0.26.0.

Should I open a similar PR for future versions ?

panglesd:
Sorry for not answering earlier...

I am personally in favour of upgrading ocamlformat, so I would welcome a similar PR! Being able to give feedback on changes that affect our codebase is also valuable.

 I don't know what other maintainers think, though.

Julow:
I'll create a similar PR next time :)

PR #440, opened 2023-06-28
Added ?warning flag to Deriving.Generator.make to enable warnings in derived code

ceastlund:
Adds the ability for ppx derivers to opt in to unused code warnings, which ppxlib would otherwise disable for derived code. This will help, over time, clean up unused code from derivers.

We made this property opt-in because currently, many ppxes are not compatible with these warnings. For example, `ppx_hash` derives `hash` and `hash_fold_t`. If one were used and not the other, there's no way to derive "half" of the ppx. So we can't opt `ppx_hash` in until we have a way to request it piecemeal. Over time, we'll have to make more ppxes compatible with this option in order to clean up unused code.

### DIFF
--- a/src/deriving.ml
+++ b/src/deriving.ml
@@ -133,6 +133,8 @@ type parsed_args =
   | Args of (string * expression) list
   | Unknown_syntax of Location.t * string
 
+type 'item derived_code = { items : 'item list; unused_code_warnings : bool }
+
 module Generator = struct
   type deriver = t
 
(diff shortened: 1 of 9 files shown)


ceastlund:
@pitag-ha and @panglesd, this is part of an intern project here at janestreet by @jacksonzou123. Let us know how this looks. We should also think how upgrading will go for external clients as we opt ppxes into this. I mostly think it won't be that hard, because manually silencing warnings is not hard if people need to.

src/deriving.ml
```diff
@@ -654,6 +671,12 @@ let wrap_sig ~loc ~hide sg =
   in
   if wrap then wrap_sig ~loc ~hide sg else sg
 
+let wrap_sig ~loc ~hide list =
+  List.concat_map list ~f:(fun { items; unused_code_warnings } ->
+    if List.is_empty items
+    then []
```
panglesd, line 677: This test is to avoid adding removing warnings in an empty structure, in case no code is generated, such as in:
```
include struct [@@@ocaml.warning "-32"] end[@@ocaml.doc "@inline"][@@merlin.hide]
```
That was not the original behaviour: the useless `inline struct` was added in any case. So this is an unrelated bonus feature from this PR!

However, those chains of shadowing can be hard to read: we now have three `wrap_sig` and `wrap_str` one after the other, and we don't know which function does what.
Maybe you could add at least a comment on what are the invariant the function expect, and what it does?
For instance, the last `wrap_sig` function is handling the empty case, and the middle `wrap_sig` invariant is that it takes a non-empty signature.

src/deriving.ml
```diff
@@ -676,6 +699,19 @@ let types_used_by_deriving (tds : type_declaration list) : structure_item list =
 let merge_generators field l =
   List.filter_map l ~f:(fun x -> x) |> List.concat |> Deriver.resolve_all field
 
+(* This function merges ['a derived] if they have the same [unused_code_warnings]. This
+   reduces the number of times we add [include struct ... end] to disable warnings. *)
+let merge_derived lists =
+  List.fold_right
```
panglesd, line 705: For some reason, I always thought `fold_left` and `fold_right` were the same function, with different order of arguments. So I was confused with this function. Finally, I learnt the right/left is about the traversal entry of the list!

panglesd (approved):
Code looks good to me, with minor comments!

I'm curious about the reason for willing to restrict the generated code.
Is it to improve compile time performances?

(I expect the compiler to remove unused values anyway, so this should not affect runtime performances).

ceastlund:
> I'm curious about the reason for willing to restrict the generated code.
> Is it to improve compile time performances?

I don't understand what you're asking here, @panglesd.

ceastlund:
If you mean, why do we want to remove unused code, then yes, it's for compilation performance. It's also to restrict final binary size, which can even improve runtime performance.

pitag-ha:
> We should also think how upgrading will go for external clients as we opt ppxes into this. I mostly think it won't be that hard, because manually silencing warnings is not hard if people need to.

Do I understand correctly that you're planning to opt the public releases of the Jane Street derivers into this? Have you had a look at how big the impact would be?

ceastlund:
No we haven't. We can leave it off in the public release if need be. But probably we're not the only ones who want the benefit.

ceastlund:
It's definitely worth some thought about how we proceed with this rollout, though.

PR #441, opened 2023-06-30
Preserve quoted attributes on antiquoted payloads

### LINKED ISSUE #406: metaquot does not transfer attributes
`[%expr [%t e] [@foo]]` produces an expression that does not have the `[@foo]` attribute on it. It seems pretty clear the writer's intention here would be to include that attribute.

ncik-roberts:
Fixes #406 by preserving attributes on antiquoted constructs.

Example:
```ocaml
let e = [%expr (() [@attr1])] in
[%expr [%e] [@attr2]]

(* Old output: *)
() [@attr1]

(* New output: *)
() [@attr1] [@attr2]
```

You can see more examples in the tests.

Some non-obvious design decisions/weirdness:
  * `attr2` goes at the end of the attribute list, not the start, because it's an "outer" attribute. This is consistent with other places that ppxlib merges attributes, e.g. `merge_attributes` in `src/extension.ml`.
  * However, this design decision means that we can't support attributes in patterns, e.g. there's no way to put `[@attr1]` at the end of `quoted_expr.pexp_attributes` in the pattern `[%expr [%e? quoted_expr] [@attr1]]`.
  * I merge the attributes by inserting a runtime call to `Stdlib.List.append`. This is susceptible to shadowing, but I couldn't think of a better idea.
  * Some type errors are worse. See the example below.

```ocaml
let _ = [%expr [%e ()] [@attr]]
[%%expect{|
Line _:
Error: This expression has type unit which is not a record type.
|}]
```

That's because this gets expanded to something like:
```ocaml
let v = () in
{ v with pexp_attributes = v.pexp_attributes @ [ attr ] }
```
This error message could be improved if we put a type annotation on `v` earlier on, e.g. `let v : Parsetree.expression = () in ...`, but that's also susceptible to the shadowing issue. Could a maintainer weigh in?

### DIFF
--- a/metaquot/ppxlib_metaquot.ml
+++ b/metaquot/ppxlib_metaquot.ml
@@ -239,10 +239,6 @@ module Patt = Make (struct
         (match attrs with
         | None -> ()
         | Some { quoted_attributes; field_name = _ } ->
-            (* We can't construct a pattern that searches for [quoted_attributes]
-               at the end of [p]'s attribute list -- the pattern language isn't
-               expressive enough. Instead, we fail.
-            *)
             assert_no_attributes quoted_attributes);
         self#typed p type_name
     | PPat (_, Some e) ->
--- a/metaquot/ppxlib_metaquot.ml
+++ b/metaquot/ppxlib_metaquot.ml
@@ -244,15 +244,15 @@ module Patt = Make (struct
   let cast self ext attrs ~type_name =
     match snd ext with
     | PPat (p, None) ->
-      (match attrs with
-       | None -> ()
-       | Some { quoted_attributes; field_name = _ } ->
-         (* We can't construct a pattern that searches for [quoted_attributes]
-            at the end of [p]'s attribute list -- the pattern language isn't
-            expressive enough. Instead, we fail.
-         *)
-         assert_no_attributes quoted_attributes);
-      self#typed p type_name
+        (match attrs with
+        | None -> ()
+        | Some { quoted_attributes; field_name = _ } ->
+            (* We can't construct a pattern that searches for [quoted_attributes]
+               at the end of [p]'s attribute list -- the pattern language isn't
+               expressive enough. Instead, we fail.
+            *)
+            assert_no_attributes quoted_attributes);
+        self#typed p type_name
     | PPat (_, Some e) ->
         Ast_builder.Default.(
           ppat_extension ~loc:e.pexp_loc
--- a/metaquot/ppxlib_metaquot.ml
+++ b/metaquot/ppxlib_metaquot.ml
@@ -157,7 +157,7 @@ module Expr = Make (struct
   let fresh_name =
     let counter = ref 0 in
     fun () ->
-      let var = "_ppx_metaquot_helper_var%d" ^ Int.to_string !counter in
+      let var = "_ppx_metaquot_helper_var%d" ^ string_of_int !counter in
       incr counter;
       var
 
@@ -177,7 +177,8 @@ module Expr = Make (struct
     | [] -> self#typed e type_name
     | _ :: _ ->
         let loc = { loc with loc_ghost = true } in
-        let open (val Ast_builder.make loc) in
+        let module Ast_builder_with_loc = (val Ast_builder.make loc) in
+        let open Ast_builder_with_loc in
         let var = fresh_name () in
         let var_expr = pexp_ident (Located.mk (Lident var)) in
         let field_name = Located.mk (Lident field_name) in
--- a/metaquot/ppxlib_metaquot.ml
+++ b/metaquot/ppxlib_metaquot.ml
@@ -3,11 +3,26 @@ open Ast_builder.Default
 module E = Extension
 module A = Ast_pattern
 
+type quoted_attributes =
+  { quoted_attributes : attributes
+  (* The attributes that appear quoted, e.g. [@foo] in [%expr [%e e] [@foo]] *)
+  ; field_name : string
+  (* The field name where attributes are stored for the kind of ASt the quoted attributes
+     are placed on, e.g. pexp_attributes. *)
+  }
+
 module Make (M : sig
   type result
 
   val annotate : result -> core_type -> result
-  val cast : extension -> result
+
+  val cast
+    :  < attributes : attributes -> result; .. >
+      (* The instance of the [std_lifters] class being used. *)
+    -> extension
+    -> quoted_attributes option
+    -> result
+
   val location : location -> result
   val location_stack : (location -> result) option
   val attributes : (location -> result) option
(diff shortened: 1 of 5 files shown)


ceastlund:
Talked with Nick about why attributes in pattern positions don't work. It's because the pattern language doesn't let us "scan" to find the attribute in a list of potentially arbitrary order/length.

As for injected code -- ppxes are inherently unhygienic, but I think references to `Stdlib` are safe enough. For annotated types, I think `Ppxlib_ast.expression` should also be safe enough. Anyone shadowing that name is asking for trouble, much more than if they happened to name something `Parsetree`.

ncik-roberts:
That makes good sense. And, I missed the fact that metaquot already does this type constraining in other places, so this PR really ought to move the place the type constraint is inserted to keep the error message equally good. And now it does.

That's now:

```ocaml
let _ = [%expr [%e ()] [@attr]]
[%%expect{|
Line _, characters 19-21:
Error: This expression should not be a unit literal, the expected type is
       Ppxlib_ast.Ast.expression
|}]
```

ncik-roberts:
I'm not familiar with ppxlib's CI, but it looks like `ocaml-benchmarks` is failing with `Sys_error("No space left on device")`. I don't think that's related to my PR, so please shout if you think otherwise.

antalsz:
We need this to fully support Jane Syntax changes; @ceastlund, what's the review status of this?

metaquot/ppxlib_metaquot.ml
```diff
@@ -102,11 +154,73 @@ module Expr = Make (struct
 
   let annotate e core_type = pexp_constraint ~loc:core_type.ptyp_loc e core_type
 
-  let cast ext =
+  let fresh_name =
+    let counter = ref 0 in
+    fun () ->
+      let var = "_ppx_metaquot_helper_var%d" ^ string_of_int !counter in
```
ceastlund, line 160: There is already a `gen_symbol` function available from ppxlib. Let's not mint a new one here.

metaquot/ppxlib_metaquot.ml
```diff
@@ -102,11 +154,73 @@ module Expr = Make (struct
...
+     [@attr]) at the end of the list is consistent with other parts of ppxlib
+     that accumulate attributes.
+  *)
+  let add_quoted_attributes self e { quoted_attributes; field_name } ~type_name
+      ~loc =
+    match quoted_attributes with
+    | [] -> self#typed e type_name
+    | _ :: _ ->
+        let loc = { loc with loc_ghost = true } in
+        let module Ast_builder_with_loc = (val Ast_builder.make loc) in
```
ceastlund, line 180: Instantiating all of `Ast_builder` here feels heavyweight. Why not just open `Ast_builder.Default` for the whole file, and pass in `~loc` as necessary?

ceastlund (changes requested):
Looks good, just two minor comments to address.

metaquot/ppxlib_metaquot.ml
ncik-roberts: Updated, thanks.

metaquot/ppxlib_metaquot.ml
ncik-roberts: Done.

ceastlund (approved):

ceastlund:
@pitag-ha and @panglesd - can one of you give this a quick skim before we merge?

panglesd:
Just got back from holiday. I'll have a look today or tomorrow!

metaquot/ppxlib_metaquot.ml
```diff
@@ -128,9 +233,18 @@ module Patt = Make (struct
...
     match snd ext with
-    | PPat (p, None) -> p
+    | PPat (p, None) ->
+        (match attrs with
+        | None -> ()
+        | Some { quoted_attributes; field_name = _ } ->
+            (* We can't construct a pattern that searches for [quoted_attributes]
+               at the end of [p]'s attribute list -- the pattern language isn't
+               expressive enough. Instead, we fail.
+            *)
```
panglesd, line 245: I'm not sure about this comment...
I think the reason for refusing attributes here, is that we are already refusing any attribute in a pattern quotation (see [line 62](https://github.com/ocaml-ppx/ppxlib/pull/441/files#diff-f0e37284b1140251946b048782d09f22023de852392a992294464806eb4fc9afR62) of your version).
For instance `let [%expr 1 [@attr]] = [%expr 1 [@attr]]` will fail, because of the attribute in the pattern extension node.
I guess the reason for this is that taking the attribute into account when making a pattern would make the pattern a bit too restrictive.

However, I don't think there is any technical impossibility with this, it's only a feature we want. We could accept attributes and make the pattern "attribute-sensitive". In this case, we could always build at compile-time the pattern with the merged attributes:
```
let [%expr [%e? { pexp_attributes = [ x ] ; _ }] [@attr2]] = [%expr 1 [@attr1] [@attr2]]
```
would have `x` be the `attr1` attribute.

The reason we can do this (despite the fact that the language of patterns is very limited) is that since patterns are not first class OCaml object (unlike expressions), there cannot be any variable containing a pattern. So, we always have the concrete pattern that we can examine. For instance this has no "pattern" equivalent:

```
let f x =
  [%expr [%e x [@attr1]] [@attr2]]
```

panglesd: My concrete suggestion is just to remove the comment:
```suggestion
```
or to modify it along the lines of:
```suggestion
            (* We don't consider attributes in pattern quotations, since that
               would make the pattern too restrictive. We thus require the
               quotation payload to be free from attributes. *)
```
(feel free to improve!)

panglesd (approved):
Looks good to me!

metaquot/ppxlib_metaquot.ml
ncik-roberts: It may well be true that some patterns are "mergeable", like `[%e? { pexp_attributes = [ x ]; _ }]`. But I think this only works for literal list patterns. For example, I can't think of a way that this could work:
```
let [%expr [%e? { pexp_attributes = x; _ }] [@attr3]] = [%expr 1 [@attr1] [@attr2] [@attr3]]
```

`[%expr [%e? { pexp_attributes = x; _ }] [@attr3]]` would need to expand to a pattern that matches the *prefix* of a list to `x` and its last element to `[@attr3]`. I can't see how this could work: OCaml's pattern language isn't expressive enough for this.

That said, I agree the current comment isn't clarifying so I will delete it.

panglesd: Thanks, that explains it! You are right.
I still think that it is a good thing to remove the comment, since (even if we could) we don't want to merge attributes on patterns.

ncik-roberts:
Thanks for your review! Does it look ready to merge?

panglesd:
Yes. Thanks for the contribution!

PR #443, opened 2023-07-06
Give semantics to reserving attribute namespace with dots

### LINKED ISSUE #442: Reserving an attribute namespace with a dot has unexpected semantics
The `-reserve-namespace` flag and the `Reserved_namespaces.reserve` function will gladly accept a value like "foo.bar", but no attribute will ever actually match this namespace. (The current implementation of reserved-namespace-checking just extracts the first dot-separated component of the attribute and checks against the set of reserved strings, and a component never contains a dot.)

I'd like to be able to reserve a namespace like "foo.bar" and have it match these attributes:
- `[@foo.bar]`
- `[@foo.bar.baz]`
- `[@foo.bar.baz.qux]`

etc.

ncik-roberts:
Resolves #442.

This PR gives different semantics to reserving attribute namespaces that contain dots, like "foo.bar". The previous semantics: reserving this namespace has no effect. The new semantics: reserving this namespaces causes attributes of the form `[@foo.bar]` or `[@foo.bar.*]` to be treated as reserved.

This is technically a backward-incompatible change, as existing reservations of "foo.bar" will start to have an effect. But it's hard to imagine anyone actually intending these reservations to have no effect, and I claim the new semantics is a more reasonable way to handle this case.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,8 @@ unreleased
 
 - Preserve quoted attributes on antiquotes in metaquot (#441, @ncik-roberts)
 
+- Attribute namespaces: Fix semantics of reserving multi-component namespaces (#443, @ncik-roberts)
+
 0.30.0 (20/06/2023)
 -------------------
 
--- a/src/name.ml
+++ b/src/name.ml
@@ -44,10 +44,15 @@ module Pattern = struct
   let matches t matched = String.Set.mem matched t.dot_suffixes
 end
 
-let get_outer_namespace name =
+(* On the namespace "a.NAMESPACE", return the pair ("a", NAMESPACE) *)
+let split_outer_namespace name =
   match String.index_opt name '.' with
   | None -> None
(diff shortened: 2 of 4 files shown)


panglesd:
Thanks a lot for the PR! I haven't looked at the implementation, but from the description, the change seems an improvement!

Apologies for the very long delay in reviewing. I have very few time to allocate to ppxlib, but I hope to be able to review this week.

pitag-ha (approved):
Thanks! Both semantics and implementation look good to me. Also thanks for the code comments, they've made review a lot fatster!

pitag-ha:
PD: I've just added a changelog entry:
```
Attribute namespaces: Fix semantics of reserving multi-component namespaces (#443, @ncik-roberts)
```

Let me know if you'd prefer something else.

ncik-roberts:
Thanks for the review. That changelog entry looks good to me.

PR #444, opened 2023-07-06
Add command-line flag opting in to unused code warnings in derived code.

ceastlund:
Gates the new `?unused_code_warnings` feature on a command-line flag. This will make it easier for opam packages to opt in to it piecemeal, so that a change in upstream packages like `ppx_jane` will not break lots of downstream packages.


src/deriving.ml
```diff
@@ -59,6 +59,17 @@ let () =
 
 let keep_w60_impl () = !keep_w60_impl || Driver.pretty ()
 let keep_w60_intf () = !keep_w60_intf || Driver.pretty ()
+let allow_unused_code_warnings = ref Options.default_allow_unused_code_warnings
+
+let () =
+  Driver.add_arg "-deriving-allow-unused-code-warnings"
+    (Set allow_unused_code_warnings)
+    ~doc:" allow ppx derivers to enable unused code warnings";
```
pitag-ha, line 67:
```suggestion
    ~doc:" Allow ppx derivers to enable unused code warnings";
```

src/deriving.ml
```diff
@@ -59,6 +59,17 @@ let () =
...
+  Driver.add_arg "-deriving-disallow-unused-code-warnings"
+    (Clear allow_unused_code_warnings)
+    ~doc:" force ppx derivers to disable unused code warnings"
```
pitag-ha, line 70:
```suggestion
    ~doc:" Force ppx derivers to disable unused code warnings"
```

src/deriving.ml
```diff
@@ -59,6 +59,17 @@ let () =
 
 let keep_w60_impl () = !keep_w60_impl || Driver.pretty ()
 let keep_w60_intf () = !keep_w60_intf || Driver.pretty ()
+let allow_unused_code_warnings = ref Options.default_allow_unused_code_warnings
+
+let () =
+  Driver.add_arg "-deriving-allow-unused-code-warnings"
```
pitag-ha, line 65: I'm wondering if we can come up with a shorter name for this flag. Would `-deriving-allow-w32-w60` be clear?

test/deriving_warning/dune
```diff
@@ -1,4 +1,4 @@
 (library
  (name ppx_warning_flags_testing_on_tests)
  (preprocess
-  (pps ppx_warning_flags_testing)))
+  (pps ppx_warning_flags_testing -deriving-allow-unused-code-warnings)))
```
pitag-ha, line 4: To also add a test for this new cmd-line flag, we could have two tests, one with this flag and one without.

pitag-ha (approved):
Thanks, @ceastlund! I think this is a very good solution for a smooth adoption to https://github.com/ocaml-ppx/ppxlib/pull/440.

src/deriving.ml
ceastlund: I think unused type warnings might also be included. I don't really want to limit the flag to certain warning numbers because we might want to extend it if more get added. Also the warning numbers are not super readable. But something like `-unused-code-warnings=true` and `-unused-code-warnings=false` would be fine if that looks better.

pitag-ha: Yes, that sounds good to me. It also makes a better cli-design IMO: With `-deriving-allow-unused-code-warnings` and `-deriving-disallow-unused-code-warnings`, one could pass in both those flags, and it would be unclear which one should dominate.

test/deriving_warning/dune
ceastlund: Done.

CHANGES.md
```diff
@@ -1,13 +1,15 @@
 unreleased
 -------------------
 
+- Add `-unused-code-warnings` command-line flag. (#444, @ceastlund)
```
pitag-ha, line 4:
```suggestion
- Driver: Add `-unused-code-warnings` command-line flag. (#444, @ceastlund)
```

Talking about "command-line" makes it quite clear that the change is about the driver, not the library, but it's still helpful to be explicit.

pitag-ha (approved):
Thanks for the improvements, @ceastlund! Looks all good to me.

PR #447, opened 2023-07-11
Driver: Multiple errors can be reported.

Burnleydev1:


CHANGES.md
```diff
@@ -1,6 +1,8 @@
 unreleased
 -------------------
 
+- Multiple errors can be reported. (#447, @burnleydev1)
```
panglesd, line 4: This description is not very informative for someone reading the changelog. It could be nice if you can explain more precisely your changes.

src/driver.ml
```diff
@@ -526,68 +526,79 @@ let get_whole_ast_passes ~hook ~expect_mismatch_handler ~tool_name ~input_name =
...
-      t list
-      * label loc list
-      * (location * label) list
-      * exn
-      * Location.Error.t list
-  in
+let apply_transforms ~tool_name ~file_path ~field ~lint_field ~dropped_so_far
+    ~hook ~expect_mismatch_handler ~input_name ~embed_errors ast =
+  (* Define a custom exception to wrap the state of the transformation process
+     when an exception occurs *)
```
panglesd, line 532: Many of the comments, such as this one, are out of date!
```suggestion
```

src/driver.ml
```diff
@@ -526,68 +526,79 @@ let get_whole_ast_passes ~hook ~expect_mismatch_handler ~tool_name ~input_name =
...
+  let acc =
+    List.fold_left cts ~init:(ast, [], [], [])
+      ~f:(fun (ast, dropped, (lint_errors : _ list), errors) (ct : Transform.t)
+         ->
+        (* Get the input name *)
+        let input_name =
+          match input_name with
+          | Some input_name -> input_name
+          | None -> "_none_"
+        in
```
panglesd, line 562: I would say the comment is redundant with the self-explanatory code?
```suggestion
        let input_name =
          match input_name with
          | Some input_name -> input_name
          | None -> "_none_"
        in
```
That's the case for other comments!

panglesd (commented):
Thanks @Burnleydev1!

I left some comments, but it looks good!

Before merging this PR, I will write an issue explaining why we need your changes: most of the discussion happened offline, and it is good to have a log of the reason of the change.

src/driver.ml
```diff
@@ -526,68 +526,79 @@ let get_whole_ast_passes ~hook ~expect_mismatch_handler ~tool_name ~input_name =
...
+                debug_dropped_attribute ct.name ~old_dropped:dropped
+                  ~new_dropped;
+                new_dropped)
+              else []
+            in
+            (ast, dropped, lint_errors, errors @ more_errors))
+  in
+  (* If there were no errors, return an Ok value *)
+  Ok (finish acc)
+(* If an exception was raised, recover from the exception*)
```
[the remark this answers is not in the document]
Burnleydev1: Hello @panglesd, thanks for the review, 
this error causes this error in the call site, 
```
This pattern should not be a constructor, the expected type is
Ppxlib__.Import.Parsetree.structure * Ppxlib__.Import.attribute list *
Ppxlib__.Location.Error.t list
```
in 
```
  | Error (st, lint_errors, errors) ->
      Error (st |> lint lint_errors |> with_errors errors)
  | Ok (st, lint_errors, errors) ->
      Ok (st |> lint lint_errors |> cookies_and_check |> with_errors errors)
```

src/driver.ml
```diff
@@ -526,68 +526,79 @@ let get_whole_ast_passes ~hook ~expect_mismatch_handler ~tool_name ~input_name =
...
     get_whole_ast_passes ~tool_name ~hook ~expect_mismatch_handler ~input_name
   in
-  let finish (x, _dropped, lint_errors, errors) =
-    ( x,
+  (* call this function when the transformation is finished *)
+  let finish (ast, _dropped, lint_errors, errors) =
+    ( ast,
       List.map lint_errors ~f:(fun (loc, s) ->
           Common.attribute_of_warning loc s),
       errors )
```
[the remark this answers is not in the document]
Burnleydev1: @panglesd  Thank you, I am not sure how to implement this yet, but it sounds like a very logical approach to me.

test/driver/exception_handling/whole_file_multiple_errors.ml
```diff
@@ -1,16 +1,18 @@
...
-        Location.raise_errorf ~loc "A located error in a whole file transform")
-      "a_raise_exc")
+  let instrument =
+    let transformation ctxt str =
+      let loc =
+        match str with
+        | [] -> Location.in_file (Expansion_context.Base.input_name ctxt)
+        | hd :: _ -> hd.pstr_loc
+      in
+      Location.raise_errorf ~loc "SHOULD APPEAR FIRST"
```
panglesd, line 11: Maybe a sentence that explains at which stage the error happens would make it more informative

```suggestion
      Location.raise_errorf ~loc "Raising an error during the first instrumentation phase"
```
and similarly for the other errors!

src/driver.ml
```diff
@@ -695,7 +695,8 @@ let map_structure_gen st ~tool_name ~hook ~expect_mismatch_handler ~input_name
       ~expect_mismatch_handler ~input_name ~embed_errors
   with
   | st, lint_errors, errors ->
```
panglesd: I cannot locate the comment on the whole relevant part.

Note that the following:
```
match expr
with pattern -> expr2
```
with only one pattern is equivalent to
```
let pattern = expr in
expr2
```
but the latter is more explicit and easier to read!

In this case, there were two cases (`Error` and `Ok`) and there is now only one, so it would make sense to turn the `match` into a `let` as explained above.

test/driver/exception_handling/run.t
```diff
@@ -124,3 +124,21 @@ and the whole AST is prepended with an error extension node.
...
   Fatal error: exception Failure("An exception in a whole file transform")
   [2]
+
+
+4. Reporting Multiple Exceptions
+
+Just one exception is raised when -embed-errors is not added.
+  $ ./whole_file_multiple_errors.exe impl.ml
+  File "impl.ml", line 1, characters 0-12:
+  Error: SHOULD APPEAR FIRST
```
panglesd, line 134: Some more explanation is needed here!

```
When the `-embed-error` flag is not set, exceptions stop the rewriting process. Therefore, only the first exception is reported to the user:
[...]
When the `-embed-error` flag is set, located exceptions thrown during the rewriting process are caught, and collected. The "throwing transformations" are ignored. After all transformations have been applied, the collected errors are appended at the beginning of the AST.

File: test/driver/exception_handling/run.t [hunk 9]
```diff
@@ -124,3 +124,21 @@ and the whole AST is prepended with an error extension node.
...
+  [1]
+
+When the argument `-embed-errors` is added, the exception is added to the existing errors and prepended to the ast.
+  $ ./whole_file_multiple_errors.exe -embed-errors impl.ml
+  [%%ocaml.error "SHOULD APPEAR FIRST"]
+  [%%ocaml.error "A second located error in a whole file transform"]
+  [%%ocaml.error "SHOULD APPEAR LAST"]
+  type a = int
+  type b = int[@@deriving deriver_raised_exception]
```
panglesd, line 143: Maybe a new `impl.ml` should be created, without this line.

panglesd (approved):

test/driver/exception_handling/run.t
panglesd: The file `impl.ml` is defined at lines [113 and 114 of this file](https://github.com/ocaml-ppx/ppxlib/pull/447/files/b04e47ed9d2e7706785f88fe13bf7c84ea13f74c#diff-264139ab557e4aad6831cc4158977efce83cb19d7079a2bce2bfa5d51c5e3f49R113-R114). Such a file is created once per example, overwriting the previous one. For instance an `impl.ml` file was already created at line [84-86](https://github.com/ocaml-ppx/ppxlib/pull/447/files/b04e47ed9d2e7706785f88fe13bf7c84ea13f74c#diff-264139ab557e4aad6831cc4158977efce83cb19d7079a2bce2bfa5d51c5e3f49R84-R86), overwritten at line 113-114.

It would be ideal to overwrite `impl.ml` one more time to have the file tailored to the example (e.g., without a mention to `deriver_raised_exception`)

panglesd (commented):
I noticed that we also need an update to the documentation for this PR!

At least [the part on handling errors](https://ocaml-ppx.github.io/ppxlib/ppxlib/good-practices.html#handling_errors) should be updated to reflect the new behaviour.

NathanReb (commented):
This looks good to me!

I have only one concern regarding documentation. I think the good practice section should only contain the recommended way to deal with errors to avoid confusion. It dives pretty deep into how we handle raised exceptions and I don't think this is useful for ppx authors.

Specifying this behaviour is important but I think it should probably live elsewhere for clarity's sake.

@panglesd what should be done regarding https://github.com/ocaml-ppx/ppxlib/pull/457? There's a bit of documentation here and I think it will conflict with your work. I'm happy to cut out the documentation changes so we can merge this and discuss how to best document the behaviour elsewhere.

NathanReb:
@Burnleydev1 do you want to take care of rebasing and fixing the last bits? I'm happy to take care of it if you don't have time for this.

panglesd:
> I'm happy to cut out the documentation changes so we can merge this and discuss how to best document the behaviour elsewhere.

I agree! Also, it would not really make sense to write documentation separately for this PR and #453, it's more efficient to do it once for both (e.g. in #457).

NathanReb:
I just rebased and dropped the documentation changes. This should be good to go!

panglesd:
Great, thanks @NathanReb!

Thanks a lot @Burnleydev1 for your work!

Issue #449, opened 2023-07-19
trunk-support does't build with ocaml-trunk


hhugo:
```
#=== ERROR while compiling ppxlib.0.29.1 ======================================#
# context     2.1.0 | linux/x86_64 |  | pinned(git+https://github.com/ocaml-ppx/ppxlib.git#trunk-support#cc54c517)
# path        ~/.opam/trunk/.opam-switch/build/ppxlib.0.29.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p ppxlib -j 7 @install
# exit-code   1
# env-file    ~/.opam/log/ppxlib-149928-37f429.env
# output-file ~/.opam/log/ppxlib-149928-37f429.out
### output ###
# (cd _build/default && /home/hugo/.opam/trunk/bin/ocamlc.opt -w -9 -g -bin-annot -I astlib/.astlib.objs/byte -I /home/hugo/.opam/trunk/lib/ocaml-compiler-libs/common -I /home/hugo/.opam/trunk/lib/ocaml/compiler-libs -no-alias-deps -open Astlib__ -o astlib/.astlib.objs/byte/astlib__Ast_502.cmo -c -impl astlib/ast_502.pp.ml)
# File "astlib/ast_502.ml", lines 113-195, characters 2-33:
# Error: This variant or record definition does not match that of type
#          "Parsetree.core_type_desc"
#        An extra constructor, "Ptyp_open", is provided in the original definition.
```

panglesd:
Thanks for the report!

This seems to be due to https://github.com/ocaml/ocaml/pull/12044 being merged in trunk. Usually, we are tagged in PRs that modify the parsetree, so that we can react and update the `trunk-support` branch, but this time we were not. We'll try to update the branch!

(FTR, there is also https://github.com/ocaml/ocaml/pull/12236 which modifies the parsetree and certainly broke trunk-support).

panglesd:
Small update: I opened #451 for re-adding trunk support.

The migration is not completely sound yet, but I'll try to improve this before I go in holiday. But it compiles, and depending on what you want to do, it might be "sufficient" (I am not confident in saying this).

hhugo:
I use trunk support to regularly test trunk or open ocaml PRs with jsoo. It is useful to give early feedback.
In the past I've been contributing to migrate-parsetree or patching ppxlib myself. It seems to me that dropping manual support for trunk is an overreaction, but well. 

[...]

PR #451, opened 2023-08-01
`trunk-support` supports trunk

panglesd:
This PR is unfinished (locations are wrong, migration is untested, migrations don't roundtrip...), but should hopefully allow users to compile ppxlib against `trunk`.
Note that you will need to pin `ocaml-compiler-libs` to `git@github.com:art-w/ocaml-compiler-libs.git#ocaml-5.2-trunk` since the dependency is not compatible with 5.02.

-----

Ocaml pretty recently added two parsetree changes in its parsetree:
- https://github.com/ocaml/ocaml/pull/12044
- https://github.com/ocaml/ocaml/pull/12236

### Local open types

I migrated local open in types as we migrate local open in patterns (4.04 -> 4.03): by raising an error. This means that any code using local open in types anywhere will be unable to use ppxlib, until ppxlib bumps its AST to 5.02. This will require a lot of work, since I suspect many PPXes directly pattern-match/create function nodes.

### Syntactic arities

For syntactic arities, it seems that there are some problems. Let's collect any issue found when implementing the migrations in this thread (I'll update the list and add more information as we make progress).

#### For 5.02 -> 5.01:

- some locations (for "subfunctions") are missing. We can use the "whole location" maybe. Similarly for`pexp_loc_stack`.
- We need to check to which node to attach the attributes to.

#### For 5.01 -> 5.02:

- I have only done the migration "naively", without using the new feature where a function has syntactic arity greater than 1.
- Some location information might also be missing? (Need investigation)
- Doing it properly requires some focus to do it right!

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Install ppxlib dependencies
         run: |
-          opam install ./ppxlib.opam --deps-only
+          opam install ./ppxlib.opam --deps-only --with-test
 
       - name: Show configuration
         run: |
@@ -42,4 +42,4 @@ jobs:
           opam list
 
       - name: Build the ppxlib
-        run: opam exec -- dune build -p ppxlib
+        run: opam exec -- dune test -p ppxlib
--- a/astlib/ast_502.ml
+++ b/astlib/ast_502.ml
@@ -147,7 +147,7 @@ module Parsetree = struct
               - [T #tconstr]             when [l=[T]],
               - [(T1, ..., Tn) #tconstr] when [l=[T1 ; ... ; Tn]].
            *)
-    | Ptyp_alias of core_type * string  (** [T as 'a]. *)
+    | Ptyp_alias of core_type * string loc  (** [T as 'a]. *)
     | Ptyp_variant of row_field list * closed_flag * label list option
         (** [Ptyp_variant([`A;`B], flag, labels)] represents:
               - [[ `A|`B ]]
--- a/astlib/ast_502.ml
+++ b/astlib/ast_502.ml
@@ -456,7 +456,7 @@ module Parsetree = struct
       pbop_loc : Location.t;
     }
 
-  and function_param (*IF_CURRENT = Parsetree.function_param *) =
+  and function_param_desc (*IF_CURRENT = Parsetree.function_param_desc *) =
     | Pparam_val of arg_label * expression option * pattern
     (** [Pparam_val (lbl, exp0, P)] represents the parameter:
         - [P]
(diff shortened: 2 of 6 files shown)


hhugo:
I was able to use this with trunk and run most of the jsoo test-suite.
The naive ***5.01 -> 5.02*** migration is responsible for one breakage in the test-suite.

pitag-ha:
Thanks, @panglesd, for this PR and all your work on trunk support!

Would it sound good to you to separate the two changes? I'd merge the support for local open types as is: It's a new syntax feature represented by a new node and your migrations do what expected.

The syntactic arity support seems a lot more complex. We'll have trouble, both getting the migrations semantically as right as possible and updating all the PPXs it'll break once we bump the AST. First things first: To get the migrations right, we should add tests. Btw, @hhugo could you point us to the jsoo test that breaks? Once the migrations are isolated and with tests, let's ask the author of the compiler PR @ncik-roberts if they'd be willing to give us a hand, similarly to how Octachron helped with the support of value binding type constraints. By the end of the day, the original author understands the change the best.

ncik-roberts:
Yes, when you are in that state please let me know. I hope I can be helpful, both with how to proceed with the OCaml 5.2 release and for a good long-term API for ppxlib here.

panglesd:
I opened this PR quickly, to allow for compilation on trunk of (hopefully most) programs that depend on ppxlib. I did not add test or anything because I just did it quickly and it was not finished.

We have made some progress on migrations with @Julow! But the changes are not pushed yet, since it's WIP.

About local open types: I think we can do better. This naive migration will get us unable to use ppxlib in conjunction with local open types. We can use an attribute with a carefully crafted payload to remember the local open type, and put it back in the upward migration.

About syntactic arity, we are also using attributes to distinguish between `fun x -> fun y -> 1` and `fun x y -> 1`.

I will:
- finish the migrations with @Julow
- write the tests
- separate in two PRs

Hopefuly soon!

pitag-ha:
Re local open types: IMO, it's not necessary to spend time improving those migrations. With your current migrations, the lack of local open types support will only be there before we bump the AST to 5.2. That will be at the end of the 5.2 RC phase / beginning of 5.2 stable release. Before getting into those release phases, people tend to only test and benchmark `trunk` on existing code. So before those release phases, it's important that we make Ppxlib compatible with 5.2 for existing code, but we don't need to support its new syntax features. Once the new compiler is stably released, we bump the AST and people will be able to use the new compiler features -including local open types- in conjunction with Ppxlib. Let me know if you have a different opinion, @panglesd.

hhugo:
> Btw, @hhugo could you point us to the jsoo test that breaks?

The broken tests all live in https://github.com/ocsigen/js_of_ocaml/blob/master/lib/tests/test_fun_call.ml
It checks the arity of functions at runtime in javascript code generated by jsoo.

Here is the failure reported by the CI https://github.com/ocsigen/js_of_ocaml/actions/runs/6326376239/job/17179814876?pr=1487

panglesd:
@hhugo could you try again with this branch? I just pushed the work we did with @Julow.

It needs testing (we haven't added tests yet) but if everything goes well, arity should be better preserved than in the previous naive version.

hhugo:
> @hhugo could you try again with this branch? I just pushed the work we did with @Julow.
> 
> It needs testing (we haven't added tests yet) but if everything goes well, arity should be better preserved than in the previous naive version.

The latest version of this PR fixes the issue. Thanks

hhugo:
https://github.com/ocaml/ocaml/pull/12639 broke support for trunk again

pitag-ha:
See https://github.com/ocaml-ppx/ppxlib/issues/449#issuecomment-1845194866.

panglesd:
I cherry picked a commit from @hhugo, and then made the trunk CI be run again (by pinning ocaml-compiler-libs).

I planned to merge but [the tests failed](https://github.com/ocaml-ppx/ppxlib/actions/runs/7138795708/job/19440961173?pr=451) with some cryptic:

> ```  
> $ cat binary_ast | ../identity_driver.exe -impl -
> -  let b = 2
> +  File "_none_", line 1:
> +  Error: Internal error: invalid [@@ocaml.ppx.context { load_path }] pair syntax
> +  [1]
> ```

I wanted to investigate the issue but again, got busy on other things. So in the light of https://github.com/ocaml-ppx/ppxlib/issues/449#issuecomment-1846000476, either I have time to investigate today, or I'll merge this evening and let users of the branch fix errors they might encounter (or fix it myself later, but at least this PR is merged).

hhugo:
After a very quick look:
- The test failure was caused by https://github.com/ocaml/ocaml/pull/12246
- updating test/driver/stdin_input/test.t/binary_ast using a trunk should fix the issue.
- I think we can't expect ppxlib to be able to read old marshalled ast.

panglesd:
If I remember correctly, there is some alternative toolchain (related to reason) that requires ppxlib to read an AST built with a different version than what ppxlib was built with, so we might want to do something here.

I think @pitag-ha explains that here: https://github.com/ocaml-ppx/ppxlib/pull/409#issuecomment-1508756493

In the meantime, let's merge this.

panglesd:
> The test failure was caused by https://github.com/ocaml/ocaml/pull/12246

Thanks for the link!

PR #453, opened 2023-08-03
Allowing multiple errors to be reported in one pass of the context_free phase

Burnleydev1:


test/driver/exception_handling/run.t
```diff
@@ -44,21 +44,42 @@ caught, so no AST is produced.
...
+ when the -embed-errors flag is not passed  
   $ ./extender.exe impl.ml
   File "impl.ml", line 2, characters 8-34:
   Error: A raised located error
   [1]
 
+
+ when the -embed-errors flag is  passed
+  $ ./extender.exe -embed-errors impl.ml
+  [%%ocaml.error "A raised located error"]
```
panglesd, line 59: We can see that it's not working yet (likely only due to the value of `embed_errors`, which is always `false` in your implementation).

If it were working, both errors("A raised located error" and "A second raised located errors" would be seen)

panglesd (commented):
Making good progress!

I reviewed only the "Catching extenders' exceptions". Except for minor comments, it looks good!

The `embed_errors` value is not passed from the driver yet, which is probably why the tests do not work yet...

src/context_free.ml
```diff
@@ -459,9 +476,11 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
 
     method! core_type base_ctxt x =
       map_node EC.core_type core_type super#core_type x.ptyp_loc base_ctxt x
+        ~embed_errors:false
```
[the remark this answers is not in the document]
Burnleydev1: What comes to mind is using open `Context-free` but I do not know if that is the best way to go about it.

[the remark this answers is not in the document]
Burnleydev1: SIr calling using `open Context-free` causes a circular depedency.

src/context_free.ml
```diff
@@ -484,14 +503,16 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
...
+              e.pexp_loc base_ctxt e ~embed_errors
         | _ -> return e
       in
       expanded >>= fun e ->
       let expand_constant kind char text =
         match Hashtbl.find_opt constants (char, kind) with
         | None -> super#expression base_ctxt e
-        | Some expand -> self#expression base_ctxt (expand e.pexp_loc text)
+        | Some expand -> (
+            try self#expression base_ctxt (expand e.pexp_loc text)
```
panglesd, line 514: Only the call to `expand`, and not the call to `self#expression`, should be inside the `try ... with`.
Otherwise, we would risk catching exceptions that were not meant to be caught here. Also, I think it is more readable to `try ... with` only the expression that can raise.

src/context_free.ml
```diff
@@ -503,18 +524,24 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
               | None ->
                   self#pexp_apply_without_traversing_function base_ctxt e func
                     args
-              | Some e -> self#expression base_ctxt e))
+              | Some e -> (
+                  try self#expression base_ctxt e
```
panglesd, line 528: Why do you catch exceptions here? I would not expect this to raise, these are just calls to "continue the traverse", not an expansion. (There are similar `try with` on such expression later!)

test/driver/exception_handling/run.t
```diff
@@ -112,12 +157,48 @@ and the whole AST is prepended with an error extension node.
...
+  File "impl.ml", line 1, characters 8-10:
+  Error: rewriter 2 failed
+  [1]
+
+ When embed-errors is not passed
+  $ ./constant_type.exe -embed-errors impl.ml
+  [%%ocaml.error "rewriter 2 failed"]
+  [%%ocaml.error "rewriter 3 failed"]
+  [%%ocaml.error "rewriter 2 failed"]
+  [%%ocaml.error "rewriter 3 failed"]
```
panglesd, line 181: Maybe the "constant rewriting" error message could be improved! It is not trivial to link the list of error messages to the original code...

CHANGES.md
```diff
@@ -1,6 +1,8 @@
 unreleased
 -------------------
 
+- Allowing multiple errors to be reported in one pass of the context_free transformation.(#453, @burnleydev1)
```
panglesd, line 4: Allowing multiple error to be reported is a nice effect of this PR, but I think the best description is that raising an exception does not cancel the whole context free phase anymore!

panglesd (commented):
Nice progress on the "constant" rewriting!

I left some small comments.

src/context_free.ml
```diff
@@ -459,9 +476,11 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
 
     method! core_type base_ctxt x =
       map_node EC.core_type core_type super#core_type x.ptyp_loc base_ctxt x
+        ~embed_errors
```
panglesd, line 479: The diff can be made quite a lot smaller by redefining `map_node` and similar functions with a fixed value for `embed_errors`. Similar to what is done [here](https://github.com/ocaml-ppx/ppxlib/pull/453/files#diff-f74171f29027fe3de9bf4383816c1a4599ab4851ecf6181abb5f41fbd39b06b9R468-R469) with the `hook` value!

src/context_free.ml
```diff
@@ -511,9 +530,12 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
...
               | Some e -> self#expression base_ctxt e))
-      | Pexp_constant (Pconst_integer (s, Some c)) ->
-          expand_constant Integer c s
-      | Pexp_constant (Pconst_float (s, Some c)) -> expand_constant Float c s
+      | Pexp_constant (Pconst_integer (s, Some c)) -> (
+          try expand_constant Integer c s
+          with exn when embed_errors -> (e, [ exn_to_error exn ]))
+      | Pexp_constant (Pconst_float (s, Some c)) -> (
+          try expand_constant Float c s
+          with exn when embed_errors -> (e, [ exn_to_error exn ]))
```
panglesd, line 538: I had something else in mind, but catching the error here is also a possibility!

panglesd (commented):
Nice improvements! Only the "special function" rewriting is missing.

I left some comments.

src/context_free.ml
Burnleydev1: Thank you Sir, this link directs back to this pr.

src/context_free.ml
```diff
@@ -518,18 +518,30 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
           | None ->
               self#pexp_apply_without_traversing_function base_ctxt e func args
           | Some pattern -> (
-              match pattern e with
+              let pat_expr =
```
panglesd, line 521: `pattern` was already a strange name for a function transforming an expression. Let's not use `pat_expr` as name for the generated code, but something more explicit such as `generated_code`.

src/context_free.ml
```diff
@@ -518,18 +518,30 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
...
+                try (pattern e, [])
```
panglesd, line 522: You can use `return` from `With_errors`.

src/context_free.ml
```diff
@@ -518,18 +518,30 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
...
+                with exn when embed_errors -> (None, [ exn_to_error exn ])
+              in
+              pat_expr >>= fun expr ->
+              match expr with
               | None ->
                   self#pexp_apply_without_traversing_function base_ctxt e func
                     args
-              | Some e -> self#expression base_ctxt e))
+              | Some e -> self#expression base_ctxt e
+              (* with exn when embed_errors -> (e, [ exn_to_error exn ]) *)))
```
panglesd, line 531: Let's avoid keeping old commented out code!

src/context_free.ml
```diff
@@ -518,18 +518,30 @@ class map_top_down ?(expect_mismatch_handler = Expect_mismatch_handler.nop)
...
+              let pat_exp =
+                try (pattern e, [])
+                with exn when embed_errors -> (None, [ exn_to_error exn ])
+              in
+              pat_exp >>= fun expr ->
+              match expr with
               | None -> super#expression base_ctxt e
-              | Some e -> self#expression base_ctxt e))
+              | Some e -> self#expression base_ctxt e
+              (* with exn when embed_errors -> (e, [ exn_to_error exn ]) *)))
```
panglesd, line 544: Same remarks!

test/driver/exception_handling/special_functions.ml
```diff
@@ -1,17 +1,9 @@
 open Ppxlib
 
-let expand e =
-  let return n = Some (Ast_builder.Default.eint ~loc:e.pexp_loc n) in
-  match e.pexp_desc with
-  | Pexp_apply (_, _arg_list) -> Location.raise_errorf "error special function"
-  | _ -> return 0
+let expand e = Location.raise_errorf ~loc:e.pexp_loc "error special function "
```
panglesd, line 3:
```suggestion
let expand e = Location.raise_errorf ~loc:e.pexp_loc "error special function"
```

panglesd (commented):
A small batch of comments!

Burnleydev1:
Thank you @panglesd, I have implemented the changes requested.

NathanReb (commented):
This looks good, great change!

Do we really need to tie this to `embed_errors`? Are there good reasons not to do this? It's already on by default when used through dune if I'm not mistaken. I guess this could break tests that rely on the output of the driver. Do you have any other uses cases in mind?

NathanReb:
The tests are broken it seems. The error extension points are printed backward and some locations were off when running the driver without `-embed-errors`.

I think this comes from a change in #447. I suggest we sort errors by location before inserting them in a separate PR so that the order of errors is a bit more reliable than what it is right now.

src/context_free.ml
```diff
@@ -197,43 +197,52 @@ module Generated_code_hook = struct
...
   in
   match EC.get_extension context x with
   | None -> super_call base_ctxt x
   | Some (ext, attrs) -> (
-      E.For_context.convert_res ts ~ctxt ext
-      |> With_errors.of_result ~default:None
+      (try
+         E.For_context.convert_res ts ~ctxt ext
+         |> With_errors.of_result ~default:None
+       with exn when embed_errors -> (None, [ exn_to_error exn ]))
```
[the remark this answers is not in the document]
NathanReb: It's also unclear to me what would be the best way to deal with an exception from the actual rewriting function here. The errors returned by `convert` are `Ast_pattern` parsing errors which are slightly different from the ones we're catching here.

NathanReb:
This should be good to go except for a CI failure with tests on 4.05 and 4.04, I'm looking into it.

We discussed a slightly different approach regarding context free transformations throwing exceptions where we would inject the error node directly in the AST in place of the node that needed to be rewritten instead of leaving the original node and inserting the error at the beginning of the AST. I still believe this would be a cleaner solution but at the moment I think the solution we have is good enough to be merged. We can still improve this later on.

NathanReb:
Should be good now!

panglesd (approved):
Thanks @NathanReb for the final work on this PR! I believe it is ready to be merged.

About the other approach of replacing extension nodes by error nodes in case of exceptions: I agree this is a very valid approach, and would definitely accept a PR with the changes. I would also think the current situation is good enough.

For reference, here are the pros and cons of each approach, as discussed in a dev meeting (quoting from the notes):
> [Main advantages of current approach:] Coherence with error collecting from other phases, less potential for later coming whole-file transformations to mess up embedded errors. We all agree that the latter shouldn't be taken too much into consideration.
> [Main advantages of replacing the extension node:] Coherence with what we ask PPX authors to do (embed the errors themselves) and no duplication of error messages in case of extension nodes, where the compiler will also emit an error if the node doesn't get replaced.

PR #454, opened 2023-08-10
Expose migration/copying/etc. functions for all AST types needed by `Pprintast`

antalsz:
This moves some cinaps stuff around, but otherwise is just a widening of the exposed APIs -- no new functionality is defined (except for a few new identity functions for the 4.14 <-> 5.00 nonmigration).

The motivation for this is that we would like to be able to link up `Pprintast` with the [flambda-backend](https://github.com/ocaml-flambda/flambda-backend) pretty-printer, and this requires the broader API surface area this PR exposes; to be clear, though, this PR doesn't do anything flambda-backend specific.

### DIFF
--- a/.ocamlformat-ignore
+++ b/.ocamlformat-ignore
@@ -18,10 +18,13 @@ astlib/ast_413.ml
 astlib/ast_414.ml
 astlib/ast_501.ml
 
-# Files that use cinaps to generate bode blocks from other code blocks work well,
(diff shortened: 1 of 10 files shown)


ceastlund (approved):

ceastlund:
Looks like it needs a changelog entry, can you add one? Otherwise, once the CI signs off on the tests, looks good to me.

antalsz:
Done, I used the PR title

PR #456, opened 2023-09-20
Fix minor error in README

nicovank:
Currently learning OCaml so I may be missing something. This confused me for a few seconds.
If `MY_VAR="foo"`, it makes sense to `[%get_env "MY_VAR"]`, but not really `[%get_env "foo"]`.


panglesd:
Yes, of course, thanks for the PR!

PR #457, opened 2023-10-24
Update documentation with new behaviour wrt exceptions

panglesd:
This is the PR that updates the documentation taking into account the changes made in #447 and #453.

Pinging @Burnleydev1 if you want to review!

This PR makes it "more acceptable" to raise for PPXs, since it is supported and does not harm the other ppx execution. So most of the changes are in the "raising" section of the manual.

### DIFF
--- a/doc/good-practices.mld
+++ b/doc/good-practices.mld
@@ -83,31 +83,6 @@ error, the embedding mechanism has many advantages. For instance, embedding
 allows to report multiple errors, and to output the part of the code that could
 be generated successfully.
 
-{2 Ppxlib and exceptions}
-
-In general, raising an exception in a registered transformation will make the
-ppxlib driver crash with an uncaught exception error. However, when spawned with
-the [-embed-errors] or [-as-ppx] flags (that's the case when Merlin calls the
-driver) the ppxlib driver still handles a specific kind of exception: Located
-exceptions. They have type {{!Ppxlib.Location.exception-Error}[Location.Error]}
-and contain enough information to display a located error message.
-
-During its {{!page-driver.driver_execution}execution}, the driver will run many
-different rewriters. In the case described above, it will catch any located
-exception thrown by a rewriter. When catching an exception, it will collect the
-error in a list, take the last valid AST (the one that was given to the raising
-rewriter) and continue its execution from there.
-
-At the end of the rewriting process, the driver will prepend all collected errors
-to the beginning of the AST, in the order in which they appeared.
-
-The same mechanism applies for the context-free rewriters: if any of them raises,
-the error is collected, the part of the AST that the rewriter was responsible to
-rewrite remains unmodified, and the context-free phase continues.
-
-The function provided by the API to raise located errors is
-{{!Ppxlib.Location.raise_errorf}[Location.raise_errorf]}.
-
 {2 Embedding the Errors in the AST}
 
 It is better to always return a valid AST, as complete as possible, but with
--- a/doc/good-practices.mld
+++ b/doc/good-practices.mld
@@ -93,22 +93,20 @@ exceptions. They have type {{!Ppxlib.Location.exception-Error}[Location.Error]}
 and contain enough information to display a located error message.
 
 During its {{!page-driver.driver_execution}execution}, the driver will run many
-different rewriter. In the case described above, it will catch any located
+different rewriters. In the case described above, it will catch any located
 exception thrown by a rewriter. When catching an exception, it will collect the
 error in a list, take the last valid AST (the one that was given to the raising
 rewriter) and continue its execution from there.
 
-At the end of the rewriting process, the driver will append at collected errors
-at the beginning of the AST, in the order in which they appeared.
+At the end of the rewriting process, the driver will prepend all collected errors
+to the beginning of the AST, in the order in which they appeared.
 
-The same mechanism applies for the context-free rewriters: if any of them raise,
+The same mechanism applies for the context-free rewriters: if any of them raises,
 the error is collected, the part of the AST that the rewriter was responsible to
 rewrite remains unmodified, and the context-free phase continues.
 
 The function provided by the API to raise located errors is
-{{!Ppxlib.Location.raise_errorf}[Location.raise_errorf]}. But keep in mind that
-in many case, there is a better AST to continue from than the one that was
-initially passed to the rewriter.
+{{!Ppxlib.Location.raise_errorf}[Location.raise_errorf]}.
 
 {2 Embedding the Errors in the AST}
 
--- a/doc/good-practices.mld
+++ b/doc/good-practices.mld
@@ -76,7 +76,7 @@ There are mainly two ways to report errors when writing a PPX.
 
 - By embedding special extensions nodes, called "error nodes", inside the
   generated code.
-- By raising a sepcific exception, letting the ppxlib driver handle the error.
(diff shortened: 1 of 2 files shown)


doc/good-practices.mld
```diff
@@ -64,24 +64,61 @@ children. Some helpers for this are provided in {{!Ppxlib.Merlin_helpers}[Merlin
...
+
+It is better to always return a valid AST, as complete as possible, but with
+"error extension nodes" at every place where successful code generation was
+impossible. Error extension nodes are special extension nodes
+[[%ocaml.error "error_message"]] that can be embedded into a valid AST and are interpreted later
+as errors, e.g., by the compiler or Merlin. As all extension nodes, they can be
+put {{:https://ocaml.org/manual/extensionnodes.html}at many places in the AST}
+to replace structure items, expressions, or patterns, for example.
 
 So whenever you're in doubt whether to throw an exception or if to embed the error as
```
Burnleydev1, line 123: Hello @panglesd , I was thinking we update this if there is no longer a preferred way of reporting errors, except the new behaviour does not still make raising as good as embedding.

doc/good-practices.mld
```diff
@@ -64,24 +64,61 @@ children. Some helpers for this are provided in {{!Ppxlib.Merlin_helpers}[Merlin
...
+In order to achieve this, errors that happen during rewriting should be handled
+in a way that do not prevent a meaningful AST to be passed to Merlin.
+
+There are mainly two ways to report errors when writing a PPX.
+
+- By embedding special extensions nodes, called "error nodes", inside the
+  generated code.
+- By raising a sepcific exception, letting the ppxlib driver handle the error.
+
+Let us already emphasize that, while exceptions can be practical to quickly fail
```
Burnleydev1, line 81: I am thinking we could maybe remove `already` in this sentence.

Burnleydev1 (commented):
I just had some minor questions, but I think it looks good to me.

doc/good-practices.mld
panglesd: Thanks!

doc/good-practices.mld
```diff
@@ -64,24 +64,61 @@ children. Some helpers for this are provided in {{!Ppxlib.Merlin_helpers}[Merlin
...
+
+In general, raising an exception in a registered transformation will make the
+ppxlib driver crash with an uncaught exception error. However, when spawned with
+the [-embed-errors] or [-as-ppx] flags (that's the case when Merlin calls the
+driver) the ppxlib driver still handles a specific kind of exception: Located
+exceptions. They have type {{!Ppxlib.Location.exception-Error}[Location.Error]}
+and contain enough information to display a located error message.
+
+During its {{!page-driver.driver_execution}execution}, the driver will run many
+different rewriter. In the case described above, it will catch any located
```
pitag-ha, line 96:
```suggestion
During its {{!page-driver.driver_execution}execution}, the driver will run many
different rewriters. In the case described above, it will catch any located
```

doc/good-practices.mld
```diff
@@ -64,24 +64,61 @@ children. Some helpers for this are provided in {{!Ppxlib.Merlin_helpers}[Merlin
...
+exception thrown by a rewriter. When catching an exception, it will collect the
+error in a list, take the last valid AST (the one that was given to the raising
+rewriter) and continue its execution from there.
 
-A common way to report an error is to throw an exception. However, this method
-interrupts the execution flow of the [ppxlib] driver and leaves later PPXs
-unexpanded when handing the AST over to Merlin.
+At the end of the rewriting process, the driver will append at collected errors
+at the beginning of the AST, in the order in which they appeared.
```
pitag-ha, line 102:
```suggestion
At the end of the rewriting process, the driver will prepend all collected errors
to the beginning of the AST, in the order in which they appeared.
```

doc/good-practices.mld
```diff
@@ -64,24 +64,61 @@ children. Some helpers for this are provided in {{!Ppxlib.Merlin_helpers}[Merlin
...
 
-Instead, it is better to always return a valid AST, as complete as possible, but
-with "error extension nodes" at every place where successful code generation was
-impossible. Error extension nodes are special extension nodes [[%ocaml.error
-error_message]] that can be embedded into a valid AST and are interpreted
-later as errors, e.g., by the compiler or Merlin. As all extension nodes,
-they can be put {{:https://ocaml.org/manual/extensionnodes.html}at many places
-in the AST} to replace structure items, expressions, or patterns, for example.
+The same mechanism applies for the context-free rewriters: if any of them raise,
```
pitag-ha, line 104:
```suggestion
The same mechanism applies for the context-free rewriters: if any of them raises,
```

doc/good-practices.mld
```diff
@@ -64,24 +64,61 @@ children. Some helpers for this are provided in {{!Ppxlib.Merlin_helpers}[Merlin
...
+the error is collected, the part of the AST that the rewriter was responsible to
+rewrite remains unmodified, and the context-free phase continues.
+
+The function provided by the API to raise located errors is
+{{!Ppxlib.Location.raise_errorf}[Location.raise_errorf]}. But keep in mind that
+in many case, there is a better AST to continue from than the one that was
+initially passed to the rewriter.
```
pitag-ha, line 111: What do you mean with the "keep in mind" side-note here?

pitag-ha (approved):
Perfect! Thank you, @panglesd and @Burnleydev1, both for the update of the docs and the implementation of this! Btw, do you want me to have a look at the PRs of the implementation, @panglesd?

doc/good-practices.mld
panglesd: I wanted to say that `raise_errorf` should only be used when there is no advantages to embed the error in the AST. But that's not even true: a PPX can raise a located error, and catch it at the point where it wants to embed the error, instead of using `result` type.
I'll remove the side-note.

panglesd, line 111:
```suggestion
The function provided by the API to raise located errors is
{{!Ppxlib.Location.raise_errorf}[Location.raise_errorf]}.
```

panglesd:
> do you want me to have a look at the PRs of the implementation, @panglesd?

If you have time, yes!
Otherwise, I'll have a look myself (since it's been a long time since I last looked at it, it is like if I were a new reviewer.

NathanReb (commented):
It looks great! Very clear!

My only concern is that it goes quite a bit into detail on how we handle exceptions while we recommend not using them.

Do you have another section in mind where we could document this behaviour and keep this section about what we recommend?

If not I think I would at least swap the order of the sub sections and put the recommendation first and only then explain how we deal with exceptions now!

panglesd:
Thanks for the review!

> My only concern is that it goes quite a bit into detail on how we handle exceptions while we recommend not using them.

That's a good point!
If we prefer to move the explanation on how the driver handles exceptions, I think a natural place would be in the "[driver execution explanation](https://ocaml-ppx.github.io/ppxlib/ppxlib/driver.html#driver_execution)" of the docs.
Would you like to do it yourself?
Otherwise, for a smaller change, putting this explanation after the better practice (or even after the documented example) seems a good improvement already.

NathanReb:
Sure, happy to take care of it!

NathanReb:
Just moved the exception handling section to `driver.mld`.

NathanReb:
@panglesd if this looks good to you, I'll go ahead and merge!

NathanReb (approved):
Thanks for the added links, it looks much better with them indeed!

PR #460, opened 2024-01-02
Fix hash used by `Attribute.get`

ceastlund:
Updated `Attribute.get` and `Attribute.has_flag` to ignore `loc_ghost` field of attribute locations. This allows flags to be used in code that is manipulated by other ppxes and has `loc_ghost` set to `false` along the way, when the original attribute in the source file had `loc_ghost` set to `true`.

The diff to the tests is probably slightly more readable commit-by-commit.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 unreleased
 -------------------
 
+- Update `Attribute.get` to ignore `loc_ghost`. (#460, @ceastlund)
+
 - Add API to manipulate attributes that are used as flags (#408, @dianaoigo)
 
 - Update changelog to use ISO 8061 date format: YYYY-MM-DD. (#445, @ceastlund)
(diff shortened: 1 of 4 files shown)


NathanReb (approved):
Looks good to me!

Just to make sure I understand this correctly, this isn't specific to flags is it? The issue is that if an attribute present in the source tree was "ghosted" by a first ppx but consumed by a second one it wasn't correctly marked as seen because `Attribute_table.remove` wouldn't find it.

NathanReb:
Out of curiosity, what was the ppx chain that lead you to this bug?

ceastlund:
You are correct that this is not specific to flags. Flags just had a test that was easy to use for this purpose.

As for what lead to this, someone here at jane street was working with a ppx, and if they ghosted their locations they got an attribute not used error, but if they didn't ghost the locations they got a location check error. So it was a catch 22, and this change is needed to resolve it.

PR #462, opened 2024-01-04
adding ocaml 4.12 + TRUNK

jmikedupont2:
Compiling on 4.12+trunk

### DIFF
--- a/astlib/ast_412.ml
+++ b/astlib/ast_412.ml
@@ -294,9 +294,13 @@ module Parsetree = struct
     *)
     | Ppat_exception of pattern
     (* exception P *)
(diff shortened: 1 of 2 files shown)


pitag-ha:
Hello @jmikedupont2, you're using the `ocaml-4.12.0+domains+effects` compiler? Is that to try out the old design on effects? Nowadays, almost everyone who wants to use domains or effects uses OCaml 5, which is stably released and works very well, e.g. `ocaml.5.1.1`. Is that an option for you?

Btw thanks for being pro-active and opening this PR! Unfortunately, we can't merge it as it would break support with the stably released OCaml 4.12.

PR #465, opened 2024-01-16
Set appropriate binary mode writing stdout

jonahbeckford:
TLDR: Windows needs the binary mode set.

Possible fix for https://github.com/ocaml-ppx/ppxlib/issues/466

### Context

I'm using Dune in an unusual way that is exposing this ppxlib bug. Basically, since I want to use PPX-es with Dune from a bytecode-only environment, and because Dune requires a C compiler to build a `(staged_pps)` or `(pps)` driver, I have to resort to:

```scheme
 (preprocess
  (action
   (run %{lib:TheNameOfTheLibrary:ppx.exe} -as-pp %{input-file})))
```

That in turn runs `-as-pp` which writes to `stdout`.

So:

```powershell
# Works
Y:\source\dksdk-type\tests\_std\lib\DkSDKCoder_PPX\ppx.exe -as-pp .\tests\_std\expression.ml -o a.pp.ml

# Fails
Y:\source\dksdk-type\tests\_std\lib\DkSDKCoder_PPX\ppx.exe -as-pp .\tests\_std\expression.ml
Caml1999M031Fatal error: exception Failure("output_value: not a binary channel")
Raised by primitive operation at Stdlib.output_value in file "stdlib.ml", line 386, characters 26-54
Called from Ppxlib__Utils.Ast_io.write.(fun) in file "src/utils.ml", line 187, characters 8-34
Called from Ppxlib__Driver.process_file in file "src/driver.ml", line 1169, characters 6-159
Called from Ppxlib__Driver.standalone in file "src/driver.ml", line 1541, characters 9-27
Re-raised at Location.report_exception.loop in file "parsing/location.ml", line 938, characters 14-25
Called from Ppxlib__Driver.standalone in file "src/driver.ml", line 1544, characters 4-59
Called from Dune__exe___ppx in file ".ppx/6241d9fed22cd957d799b2636a4b50f0/_ppx.ml-gen", line 1, characters 9-36
```

### DIFF
--- a/src/utils.ml
+++ b/src/utils.ml
@@ -154,7 +154,11 @@ module Ast_io = struct
   let read input_source ~input_kind =
     try
       match input_source with
-      | Stdin -> from_channel stdin ~input_kind
+      | Stdin ->
+          (match input_kind with
+          | Necessarily_binary -> set_binary_mode_in stdin true
+          | _ -> ());
+          from_channel stdin ~input_kind
       | File fn -> In_channel.with_file fn ~f:(from_channel ~input_kind)
     with exn -> (
       match Location.Error.of_exn exn with
--- a/src/utils.ml
+++ b/src/utils.ml
@@ -168,9 +168,7 @@ module Ast_io = struct
     try
       match input_source with
       | Stdin ->
-          (match input_kind with
-          | Necessarily_binary -> set_binary_mode_in stdin true
-          | _ -> ());
+          set_binary_mode_in stdin true;
           from_channel stdin ~input_kind
       | File fn -> In_channel.with_file fn ~f:(from_channel ~input_kind)
     with exn -> (
--- a/src/utils.ml
+++ b/src/utils.ml
@@ -3,6 +3,15 @@ open Import
 let with_output fn ~binary ~f =
   match fn with
   | None | Some "-" ->
+      (* Flipping back and forth from binary to text is not
+         a good idea, so we'll make two simplifying assumptions:
+         1. Assume that nothing is buffered on stdout before
+            entering [with_output]. That means we don't need to
+            flush the stdout on entry.
+         2. Assume that nothing else is sent to stdout after
+            [with_output]. That means it is safe to leave stdout
+            channel in binary mode (or text mode if [binary=true])
+            after the function is done. *)
       set_binary_mode_out stdout binary;
       f stdout
   | Some fn -> Out_channel.with_file fn ~binary ~f
@@ -114,6 +123,10 @@ module Ast_io = struct
           parse_source_code ~kind ~input_name ~prefix_read_from_source ch
       | Necessarily_binary -> Error Not_a_binary_ast
     in
+    (* Marshalled AST must be read in binary mode. Even though we don't know
+       before reading the magic number when the file has a marshalled AST,
+       it is safe to read source files in binary mode. *)
+    set_binary_mode_in ch true;
     match read_magic ch with
     | Error s -> handle_non_binary s
     | Ok s -> (
--- a/src/utils.ml
+++ b/src/utils.ml
@@ -2,7 +2,9 @@ open Import
 
 let with_output fn ~binary ~f =
   match fn with
-  | None | Some "-" -> f stdout
+  | None | Some "-" ->
+      set_binary_mode_out stdout binary;
+      f stdout
   | Some fn -> Out_channel.with_file fn ~binary ~f
 
 module Kind = struct
--- a/test/driver/run_as_ppx_rewriter_preserve_version/print_magic_number.ml
+++ b/test/driver/run_as_ppx_rewriter_preserve_version/print_magic_number.ml
@@ -1,9 +1,5 @@
 let magic_length = String.length Astlib.Config.ast_impl_magic_number
 let buf = Bytes.create magic_length
-
-let len =
-  set_binary_mode_in stdin true;
-  input stdin buf 0 magic_length
-
+let len = input stdin buf 0 magic_length
 let s = Bytes.sub_string buf 0 len
 let () = Printf.printf "Magic number: %s" s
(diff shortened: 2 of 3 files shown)


src/utils.ml
```diff
@@ -2,7 +2,9 @@ open Import
 
 let with_output fn ~binary ~f =
   match fn with
-  | None | Some "-" -> f stdout
+  | None | Some "-" ->
+      set_binary_mode_out stdout binary;
+      f stdout
   | Some fn -> Out_channel.with_file fn ~binary ~f
```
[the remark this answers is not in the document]
jonahbeckford: Binary mode should be set when writing binary to a file. I thought the `~binary` in `Out_channel.with_file fn ~binary ~f` already handled that.

I could be very wrong. Please remember this is the first time I have seen this codebase and I barely understand it.

src/utils.ml
```diff
@@ -152,7 +154,11 @@ module Ast_io = struct
   let read input_source ~input_kind =
     try
       match input_source with
-      | Stdin -> from_channel stdin ~input_kind
+      | Stdin ->
+          (match input_kind with
+          | Necessarily_binary -> set_binary_mode_in stdin true
+          | _ -> ());
+          from_channel stdin ~input_kind
```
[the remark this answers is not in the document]
jonahbeckford: The default text mode translates `\r\n` to `\n`, so a magic number like `Caml1999M033` should have no problem being read in text mode. In general, as long as there is no byte `0x0d` (CR) in the magic number it should be safe to read in text mode, and then switch to binary mode.

The only thing I am worried about is buffering. Would the entire buffer be translated (big problem), rather than just the magic number field (no problem). Perhaps @nojb can comment?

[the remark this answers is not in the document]
nojb: > The only thing I am worried about is buffering. Would the entire buffer be translated (big problem), rather than just the magic number field (no problem). Perhaps @nojb can comment?

As mentioned above, by the time the bytes get into the input buffer, translation has already ocurred, so it is too late to switch modes after checking the magic number. Switching `stdin` to binary mode before reading the magic seems the only way. And anyway, binary mode works just fine for reading source code.

test/driver/run_as_ppx_rewriter_preserve_version/print_magic_number.ml
```diff
@@ -1,5 +1,9 @@
 let magic_length = String.length Astlib.Config.ast_impl_magic_number
 let buf = Bytes.create magic_length
-let len = input stdin buf 0 magic_length
+
+let len =
+  set_binary_mode_in stdin true;
+  input stdin buf 0 magic_length
+
```
nojb, line 7: Strictly speaking this is probably not needed because the magic number does not contain CRs. Also this test is not run on Windows :)

nojb (commented):
Thanks for the ping; I left some comments.

test/driver/run_as_ppx_rewriter_preserve_version/print_magic_number.ml
jonahbeckford: reverted

jonahbeckford:
Addressed feedback. Changed from Draft to normal PR. The first problem mentioned in #466 is fixed, but the "Second Attempt" is still not fixed (and I have no good way of troubleshooting it).

src/utils.ml
```diff
@@ -152,7 +167,11 @@ module Ast_io = struct
   let read input_source ~input_kind =
     try
       match input_source with
-      | Stdin -> from_channel stdin ~input_kind
+      | Stdin ->
+          (match input_kind with
+          | Necessarily_binary -> set_binary_mode_in stdin true
+          | _ -> ());
```
nojb, line 173: `input_kind` can be either `Necessarily_binary` or `Possibly_source`. I think binary mode should be set on both cases, because both cases may end up reading binary data.

jonahbeckford: Changed and tested. (The "Second" problem still remains open)

nojb:
@jonahbeckford do you confirm that you still have the "Second Attempt" issue even after the latest changes to this PR?

nojb:
(Yes, you do; I had missed your last message).

nojb:
@jonahbeckford for the "second" problem, there is a suspicious `open_in` in https://github.com/ocaml/merlin/blob/05e90d823dd5b5acaaa70d2b2095b5bde2e395c4/src/ocaml/driver/pparse.ml#L180

Did you try changing it to `open_in_bin`?

jonahbeckford:
> @jonahbeckford for the "second" problem, there is a suspicious `open_in` in https://github.com/ocaml/merlin/blob/05e90d823dd5b5acaaa70d2b2095b5bde2e395c4/src/ocaml/driver/pparse.ml#L180
> 
> Did you try changing it to `open_in_bin`?

Wow, you deserve an Olympics medal. The "Second Attempt" works! I'll submit a PR for merlin for that bit.

src/utils.ml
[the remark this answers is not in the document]
pitag-ha: Thanks, both of you, for the clarifications, they're very useful for me!

I agree with both things @nojb has said: Fully switching `stdout` to binary mode should not be a problem as the driver either outputs a binary AST or text (e.g. errors, lint results, a pprinted AST or similar). Mixing both would be a bug. Even if we implemented a server mode for the driver some day, the output kind should remain the same between different server calls. However, a comment about the impact on `stdout`'s buffer and about leaving `stdout` in binary mode is useful, thanks for already writing it, @jonahbeckford.

test/driver/run_as_ppx_rewriter_preserve_version/print_magic_number.ml
pitag-ha: The way it is now, this test doesn't fail for you on Windows, @jonahbeckford , does it?

jonahbeckford: Everything now passes on Windows using the latest ppxlib PR and two merlin PRs (one closed, one open).

pitag-ha (approved):
This all looks good to me. Thanks again, @jonahbeckford!

If there are no objections (@NathanReb , @panglesd , @ceastlund), I'll go ahead and merge it.

NathanReb:
Looks good to me too! Let's merge!

PR #467, opened 2024-02-01
Fix a bug in bench that was running the driver on dune generated folders

NathanReb:


panglesd (commented):
There is another call to `readdir_full_path`, don't we want to also filter dot files there?

I guess so, in which case it would be better to modify `readdir_full_path` to filter the dot files. It would also allow to avoid calling `Filename.concat` followed by `Filename.basename`!

NathanReb:
Done!

panglesd (approved):
Let's merge!

PR #468, opened 2024-02-02
Make `Context_free.map_top_down`'s `embed_errors` arg optional

NathanReb:
This extra argument that was added in #453 caused breakage in `ocsigen-i18n` that was using `Context_free.map_top_down`.

Although it's undocumented, this function is still part of the public API. Adding an extra optional argument is not strictly speaking non-breaking but is good enough of an effort considering the status of this function and in particular, the fact that it only has a single public external user at the moment, for which it does prevent breakage.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 unreleased
 -------------------
 
+- Add an optional `embed_errors` argument to `Context_free.map_top_down` that
+  controls how to deal with exceptions thrown by context-free rules.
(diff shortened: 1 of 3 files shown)


NathanReb:
I added a changelog entry in case it does introduce breakage outside opam-repo somewhere, to help users track it down to the next release!

pitag-ha (approved):
Thanks a lot for noticing the breakage during the release and for the quick fix!

PR #469, opened 2024-02-05
Remove ocaml-compiler-libs pin

NathanReb:
This pin isn't required anymore since the release of `ocaml-compiler-libs.v0.17.0`


pitag-ha (approved):
Thanks!

PR #470, opened 2024-02-05
Simplify expect tests

NathanReb:
It seems that in 5.1.0 the output of the toplevel was sometime prefixed with an empty line and that led us to have seperate expect test output reference files for 5.1.0.

This PR slightly modifies the expect test runner to trim the toplevel output, getting rid of said lines on the way.

There is also a small bonus: I'm suggesting a rewriting of one of the deriving test so it relies on a positive output of the toplevel rather than on error reporting from the compiler. I maybe lacking context but it seems to me the two test are equivalent and the updated one won't trigger a failure due to what appears to be a change in the error reporting in OCaml 5.2.

Please let me know what you think!


NathanReb:
I should've opened this as draft, I wanted to see the CI run to test the 5.1.0 thing and make sure it was working alright!

NathanReb:
It seems to be working fine! This should be ready to review!

NathanReb:
There is something incredibly fishy going on here, even with the trimming we end up with a leading empty line in some cases. I'll look into it!

NathanReb:
Got it, errors are reported by another printer, I'll try fix this one as well!

NathanReb:
Let's wait for the CI but I fixed the error reporting as well and manually tested it on 5.1.1 so it should be good this time.

test/expect/expect_test.ml
```diff
@@ -24,6 +24,12 @@ let run_expect_test file ~f =
     if Sys.file_exists corrected_file then Sys.remove corrected_file;
     exit 0)
 
+let capture_trimmed_fmt printer arg =
+  let buf = Buffer.create 1024 in
+  let buf_fmt = Format.formatter_of_buffer buf in
+  let _ = printer buf_fmt arg in
```
panglesd, line 30: That makes me more confident we are not ignoring a value!
```suggestion
  let () = printer buf_fmt arg in
```

panglesd (approved):
Nice! The trimming removes many version-specific tests.

Issue #448, opened 2023-07-12
Ppxlib’s behaviour in case of raised exception.


panglesd:
There has been several "private" discussions about ppxlib’s behaviour in case a transformation throws an exception. Those conversations lead to decisions, but the reasons behind the decisions were not always easy to find. This issue is an attempt at solving this!

## Past and current behaviour

When I started working on `ppxlib`, when a transformation raised an exception, the whole AST was removed and replaced by a single item, an error extension node containing the exception.

This was not a problem for the compiler, which reports the first error it encounters and stops. However, it was a problem for Merlin, which had no way to "recover from the error". None of its features could work on such an empty AST (see also https://github.com/ocaml-ppx/ppxlib/issues/314).

So, we needed a way to recover from the error, to give a reasonable AST to Merlin.
After some discussions, we implemented in https://github.com/ocaml-ppx/ppxlib/pull/315 the following behaviour:
- When a transformation raises, the last valid AST is used (that is, the one given to the transformation), with an error extension node appended at its beginning.
- For context-free transformations, a raise would cancel the context-free phase: the last valid AST would be the one when starting the context-free phase.

This behaviour was discussed, also seeking advices from Merlin people. I think the reasonning behind this was to be able to give PPX authors the possibility to distinguish between a "hard" error, where continuing with the unrewritten AST would not make sense, and "recoverable" errors where it is possible to continue. Recoverable errors would be embedded manually in the AST by the rewriter.

However, there are problems with this approach:
- PPXs were written with exception raising even in the case of recoverable errors, and it is not scalable to change them all
- We could not find a single example of "non-recoverable" PPX error, where it is better to use the last valid AST
- Not running the upcoming transformations leads to many unrelated errors (unbound values, unrewritten extension node...). Those unrelated errors were difficult to understand: they could be in completely different part of the file, and have no links with the actual error.

For instance in:

```ocaml
module A = struct
  type t = [%ext failing] [@@deriving show]
end

module B = struct
  type t = A | B [@@deriving show]
  let f = print A
end
```
there would be the "Unbound value `print`" error, which is hard to link with the failure of expanding `[%ext A]`.

## New behaviour

After starting to patch some rewriters to have them embed recoverable errors, doubts with the current approach were raised.
We discussed the approach, and think the following is better:
- When a transformation raises, the last valid AST is used, as input to the upcoming transformations. All errors are collected. At the end, the errors are appended, as extension nodes, at its beginning.
[...]

PR #472, opened 2024-02-09
Insert caught exception in place of the rewritten or generated item

NathanReb:
This should wrap up the work on #448 !

When a context free rule raises, the exception was caught and turned into an error extension node prepended to the whole AST. This changes this behaviour to instead insert the error extension where the generated code would be, had the rule not raised.


NathanReb:
Fixed the build failure and enabled this behaviour for special functions and constants as well!

src/context_free.ml
```diff
@@ -197,6 +197,48 @@ module Generated_code_hook = struct
...
+  | EC.Class_type_field -> Ast_builder.Default.pctf_extension ~loc extension
+  | EC.Core_type -> Ast_builder.Default.ptyp_extension ~loc extension
+  | EC.Expression -> Ast_builder.Default.pexp_extension ~loc extension
+  | EC.Module_expr -> Ast_builder.Default.pmod_extension ~loc extension
+  | EC.Module_type -> Ast_builder.Default.pmty_extension ~loc extension
+  | EC.Pattern -> Ast_builder.Default.ppat_extension ~loc extension
+  | EC.Signature_item -> Ast_builder.Default.psig_extension ~loc extension []
+  | EC.Structure_item -> Ast_builder.Default.pstr_extension ~loc extension []
+  | EC.Ppx_import ->
+      (* Insert the error in the type decl manifest *)
```
panglesd, line 218: We've had the [same idea](https://github.com/ocaml-ppx/ppxlib/pull/294/files#diff-1039f50ffb020bed079f732077cd7111c5abbf9845a47564d28af250f0e95b98R133) :slightly_smiling_face: 

This function could be in `extension.ml`, in my opinion, but I'm not even sure and that might be bikeshedding: feel free to ignore the suggestion...

panglesd (approved):
Thanks! It will be great to be sure not to have an "unrewritten extension node" error displayed in a list of errors. And it's a very natural place to put the error in.

src/context_free.ml
NathanReb: I moved it in context_free because putting it in extension meant to make it public and I'm not sure we want to.

It is also only used for errors and is therefore very specific.

PR #474, opened 2024-02-15
Generate binary AST on the fly in stdin input driver test

NathanReb:
The point of this test is not to ensure we can read binary ASTs from a different version than the one the driver is compiled with. There is already another test for this use case.


pitag-ha (approved):
Yes, that sounds good and solves the problem with the change in the PPX context, thanks!

Have you already found out why the test about preserving the AST version isn't affected by the PPX context change?

panglesd (approved):

PR #479, opened 2024-02-27
Fix migration of Pexp_function type_constraints

NathanReb:
In 5.2 the representation of function expressions has been simplified. Among the simplifications, a `Pexp_function` now has a `type_constraint` argument that captures a type constraint or coercion on the function's return type.

E.g. in 5.01 the following:
```ocaml
let f x : string = x
```
was roughly:
```ocaml
Pexp_fun
  Nolabel
  None
  (Ppat_var "x")
  (Pexp_constraint (Pexp_ident "x") (Ptyp_constr "t")
```

In 5.02 it becomes:
```ocaml
Pexp_function
  [(Pparam_val Nolabel None (Ppat_var "x"))]
  (Some (Pconstraint (Ptyp_constr "t")))
  (Pfunction_body (Pexp_ident "x"))
```

It's worth noting that it's still possible to have the coercion or constraint in the function body. We chose to always extract it from the body and promote it to the `Pexp_function` argument when migrating from 5.1 to 5.2.

There was a bug in the coercion case where we were promoting it AND keeping it in the body which lead to compilation errors as the expression wasn't typing correctly anymore.

### DIFF
--- a/astlib/migrate_501_502.ml
+++ b/astlib/migrate_501_502.ml
@@ -141,7 +141,7 @@ and copy_expression_desc loc :
               (* Same as above, might not roundtrip but hopefully OK. *)
               let c1 = Option.map copy_core_type c1
               and c2 = copy_core_type c2 in
-              (acc, Some (Ast_502.Parsetree.Pcoerce (c1, c2)), take_body e)
+              (acc, Some (Ast_502.Parsetree.Pcoerce (c1, c2)), take_body exp)
           | _ -> (acc, None, take_body e)
       and take_arguments_fun acc arg_label opt_default pat expr =
         let acc =


NathanReb:
I'm trying to figure out if there is any value in trying to make this round-trip properly as it could easily be achieved with attributes but I don't want to go down that road if it's not necessary.

NathanReb:
cc @Octachron @kit-ty-kate 

This fixes the bug in https://github.com/Octachron/bug-reproduction-archives/pull/1.

I'm still a bit puzzled as to why it didn't trigger when manually compiling but at least now it's not triggering with dune anymore and the migration is correct!

panglesd:
Thanks @NathanReb !

I would appreciate a little help to review, if you still have the issue freshish in your mind.

In particular, your example treats about constraints, but your modification is on the coercion migration... Could you confirm that the actual code where the migration was wrong, and was fixed by this PR is:

```
let f x :> string = x
```
?

(We need a way to easily add tests for migrations to review those kind of changes...)

NathanReb:
The issue was with with coercion which is one of the two possible type of type_constraints, see:
- [Pexp_function definition](https://github.com/ocaml-ppx/ppxlib/blob/trunk-support/astlib/ast_502.ml#L327)
- [type_constraint definition](https://github.com/ocaml-ppx/ppxlib/blob/trunk-support/astlib/ast_502.ml#L514)

What happens in the migration is that we want to move those constraint up from the function body to the `Pexp_function` as described in the PR description above.

We were doing it correctly in the `Pexp_constraint` case, see [here](https://github.com/ocaml-ppx/ppxlib/blob/trunk-support/astlib/migrate_501_502.ml#L126) and [here](https://github.com/ocaml-ppx/ppxlib/blob/trunk-support/astlib/migrate_501_502.ml#L139, you can see we're calling `take_body` on the expression wrapped in the `Pexp_constraint`.

In the Pexp_coerce case though we were not. It's likely just a typo but you can see [here](https://github.com/ocaml-ppx/ppxlib/blob/trunk-support/astlib/migrate_501_502.ml#L144) that we call `take_body e`, `e` not being the expression wrapped in the `Pexp_coerce` but instead the expression we are matching over, i.e. the `Pexp_coerce` itself.

Roughly, instead of producing:
```ocaml
let f x : t :> string = body
```
We were produing:
```ocaml
let f x : t :> string = (body : t :> string)
```

Which was not compiling because one of those coercion happened first and made the other incorrect as `t` was not visible as the return type anymore.

panglesd (approved):
Thanks a lot for the explanation! It's much clearer.

I had a problem: I remembered very strongly thinking deeply about those constraints being both on the expression and the pattern. I remember discussing that but could not find it back...

But actually, the discussion was about the 5.0 <-> 5.1 migrations! See for instance https://github.com/ocaml-ppx/ppxlib/pull/420#issuecomment-1512514602

I can now approve the PR with a zen mind. The issue is clear and the fix too, and I know the bug had no deep thinking behind it.

(I was curious, since in 5.0 the constraint is added to both the expression and the pattern, how it could deal with this case. 
```ocaml
OCaml version 5.0.0
Enter #help;; for help.

# type t1 = [ `A ] ;;
# type t2 = [ `A | `B ] ;;
# let f : t1 :> t2 = `A ;;
Ptop_def
  [
    structure_item (//toplevel//[1,0+0]..[1,0+21])
      Pstr_value Nonrec
      [
        <def>
          pattern (//toplevel//[1,0+4]..[1,0+16]) ghost
            Ppat_constraint
            pattern (//toplevel//[1,0+4]..[1,0+5])
              Ppat_var "f" (//toplevel//[1,0+4]..[1,0+5])
            core_type (//toplevel//[1,0+14]..[1,0+16]) ghost
              Ptyp_poly
              core_type (//toplevel//[1,0+14]..[1,0+16])
                Ptyp_constr "t2" (//toplevel//[1,0+14]..[1,0+16])
                []
          expression (//toplevel//[1,0+4]..[1,0+21])
            Pexp_coerce
            expression (//toplevel//[1,0+19]..[1,0+21])
              Pexp_variant "A"
              None
            Some
              core_type (//toplevel//[1,0+8]..[1,0+10])
                Ptyp_constr "t1" (//toplevel//[1,0+8]..[1,0+10])
                []
            core_type (//toplevel//[1,0+14]..[1,0+16])
              Ptyp_constr "t2" (//toplevel//[1,0+14]..[1,0+16])
              []
      ]
  ]

val f : t2 = `A
```
so in 5.0 `` let f : t2 = (`A : t1 :> t2)` `` and `` let f : t1 :> t2 = `A `` have exactly the same parsetree... I wonder if we deal with that well in the 5.0 <-> 5.1 migration, but let's say yes, since no-one complained...)

kit-ty-kate:
Should the preview package in opam-repository be updated? If so I can do if you want.

I've tested `trunk-support` locally and it seems to work just fine and has fixed the two issues reported

NathanReb:
Sorry I was off on friday! Yes I will release a new preview version! I just need to check if the `ocaml.ppx.context` thing has been fixed upstream!

PR #480, opened 2024-03-04
Fix source quotation in OCaml 5.2

NathanReb:
The compiler removed some logic that was reopening the source file based on `Location.input_name` and instead only rely on `Location.input_lexbuf` for source quotation.

This PR first adds a test that rely on source quotation. Our test suite failed to catch this bug because all our tests use `OCAML_ERROR_STYLE=short`.
It then changes the way we read source files to stick to what the compiler does in `Pparse` while preserving our features such read from stdin. We now set the input_lexbuf properly.

### DIFF
--- a/astlib/location.ml
+++ b/astlib/location.ml
@@ -1,6 +1,7 @@
 include Ocaml_common.Location
 
 let set_input_name name = input_name := name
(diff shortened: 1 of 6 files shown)


NathanReb:
See https://github.com/ocaml/ocaml/pull/12991 for context!

panglesd (approved):
Looks good to me!

PR #483, opened 2024-03-22
fix: set input_lexbuf when reading binary AST too

anmonteiro:
see https://github.com/ocaml/ocaml/pull/12991


NathanReb:
@anmonteiro I'm not super familiar with `melange` but I'm guessing this is necessary for reason's syntax support right?

With reason you're using the driver on the marshalled AST resulting from parsing the reason syntax, whereas with regular OCaml syntax, you're using the driver directly on the source files. Are my assumptions correct?

anmonteiro:
I don’t think so. There’s no dialect present in my repro https://github.com/anmonteiro/no-source-quotation-repro

I think this will rather happen on all staged_pps (the instructions in my repro pass the `-ppx` flag)

NathanReb:
Ok thanks for the details!

I'll add a test for this and merge!

NathanReb (approved):
Thanks for the fix!

I just fixed the formatting and added a test to show the before/after the fix.

anmonteiro:
Thank you!

PR #486, opened 2024-04-18
Add OCaml 5.2 support

NathanReb:
This PR simply consists in merging `trunk-support` into main.

Most of the of the work here except for the first commit has been reviewed separately as we opened PRs to get those changes merged into the `trunk-support` branch itself.

It has been tested with the beta release of the compiler already as the 0.32.1~5.2preview release has been kept up to date with `trunk-support` as we were adding and fixing the 5.2 support.


NathanReb:
One thing I'm not quite sure how to handle: the changelog.

There is a `0.32.1~5.2preview` release but I think we're supposed to mark it as unavailable once we have an actual release with the 5.2 support. I'm not quite sure whether we should keep a `0.32.1~5.2preview` section in the changelog or rename it as `unreleased` on main until we cut an "actual" release.

I'd instinctively go for the latter as otherwise, if we release right after merging this, the changelog section for that release would be empty which does not make much sense.

Let me know what you think!

Issue #473, opened 2024-02-13
Unused-type declaration warning and bisect_ppx coverage points


mbarbin:
While using `[@@deriving sexp_of]` and `[@@deriving equal]`, I've observed that
an unused function is being generated:

```ocaml
module Blah = struct
  type t = G [@@deriving_inline sexp_of]

  let _ = fun (_ : t) -> ()
  let sexp_of_t = (fun (G) -> Sexplib0.Sexp.Atom "G" : t -> Sexplib0.Sexp.t)
  let _ = sexp_of_t
  [@@@end]
end

module Blih = struct
  type t = H [@@deriving_inline equal]

  let _ = fun (_ : t) -> ()
  let equal = (Stdlib.(=) : t -> t -> bool)
  let _ = equal
[@@@end]
end
```

The specific part I'm referring to is `let _ = fun (_ : t) -> ()`.

Despite my efforts to identify the code generating this part within
`ppx_sexp_conv` and `ppx_equal`, I've been unsuccessful. I'm wondering if it
could be a part of `ppxlib` itself?

Would you be willing to assist me in locating the code that generates this
binding, and perhaps help me understand its purpose? 

My inquiry stems from the observation that this construct interacts with
`bisect_ppx`. When this code is instrumented, a non-visitable coverage point is
inserted, thereby reducing the reported code coverage and preventing it from
reaching 100%.

Here's an example of how this code gets instrumented:

```ocaml
let _ = fun (_ : t) -> ___bisect_visit___ 0; ()
```

I would be interested in looking for a solution to this issue, but first, I think I'd
need to understand the rationale behind the `let _ = fun (_ : t) -> ()` 
construct.

Thank you for your time!

panglesd:
The `let _ = fun (_ : t) -> ()` line is added by ppxlib in [this function](https://github.com/ocaml-ppx/ppxlib/blob/main/src/deriving.ml#L703).

The reason for it is to remove any `Error (warning 34 [unused-type-declaration]): unused type t` if the type is not used.

I guess that the reason we do that is that the type is, in some ways, used: if not by OCaml, it is used by ppxlib to generate some code.

In my opinion, this could be removed: the deriving should not "modify" the type it derives from, which includes its "usedness". If defining a type only for the functions it will generate, one can use an extension node `type%ext_name t = ...`. but I might be missing something!

mbarbin:
Hi @panglesd,

[...]

PR #490, opened 2024-05-25
Add driver flag to force opting in to unused code warnings in derived code

mbarbin:
To quote the documentation recently added for the disable warning flags:

> Ppxlib driver has a variety of ways to disable warnings that can be triggered when using `[@@deriving ...]`. These are all enabled by default but we added flags to let driver users disable them. To allow smooth transition from always adding them to never do so and let individual ppx-es do what they must to avoid triggering warnings, we also added optional arguments to `Deriving.make` so that the ppx themselves can declare whether they need the driver to disable warnings or not.
>
> One such flag and optional argument pair is the `-unused-code-warnings` flag and `?unused_code_warning` `Deriving.V2.make` argument. Both of those default to `false` and control whether we disable warnings 32 and 60 for generated code.

This PR adds a new value for the flag, in addition to `true|false`, it adds `force` which allows to force enabling unused code warnings even when the deriver arg was set to `false`. See the 2 new lines in the table documenting the combination of `Deriver arg` and `Driver Flag`:

```diff
 | Deriver arg | Driver Flag | Unused Code Warnings
 |-------------|-------------|----------------------
 | true        | true        | Enabled
 | true        | false       | Disabled*
+| true        | force       | Enabled
 | false       | true        | Disabled
 | false       | false       | Disabled
+| false       | force       | Enabled
```

Motivation:

This is a PR that was prompted by the discussion in #473 

In particular, I am interested to see what is the result of enabling unused code warnings locally without requiring the PPX author opt-in part of it.

### DIFF
--- a/src/options.ml
+++ b/src/options.ml
@@ -1,4 +1,8 @@
-let default_allow_unused_code_warnings = false
+module Forcable_bool = struct
+  type t = True | False | Force
+end
+
+let default_allow_unused_code_warnings : Forcable_bool.t = False
 let perform_checks = false
 
 (* The checks on extensions are only to get better error messages
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,8 @@ details.
 
 ### Other changes
 
+- Driver: Add `-unused-code-warnings=force` command-line flag argument. (#490, @mbarbin)
+
 - new functions `Ast_builder.{e,p}list_tail` that take an extra tail
   expression/pattern argument parameter compared to `Ast_builder.{e,p}list`, so
   they can build ASTs like `a :: b :: c` instead of only `[ a; b ]`.
--- a/doc/driver.mld
+++ b/doc/driver.mld
@@ -91,8 +91,21 @@ driver.exe [extra_args] [<files>]
   -no-merge                   Do not merge context free transformations (better for debugging rewriters). As a result, the context-free transformations are not all applied before all impl and intf.
   -cookie NAME=EXPR           Set the cookie NAME to EXPR
   --cookie                    Same as -cookie
-  -help                       Display this list of options
-  --help                      Display this list of options
+  -deriving-keep-w32 {impl|intf|both}
+                               Do not try to disable warning 32 for the generated code
+  -deriving-disable-w32-method {code|attribute}
+                               How to disable warning 32 for the generated code
+  -type-conv-keep-w32 {impl|intf|both}
+                               Deprecated, use -deriving-keep-w32
+  -type-conv-w32 {code|attribute}
+                               Deprecated, use -deriving-disable-w32-method
+  -deriving-keep-w60 {impl|intf|both}
+                               Do not try to disable warning 60 for the generated code
+  -unused-code-warnings {true|false|force}
+                               Allow ppx derivers to enable unused code warnings (default: false)
+  -help                        Display this list of options
+  --help                       Display this list of options
+
 v}
 {%html: </details>%}
 
--- a/src/deriving.ml
+++ b/src/deriving.ml
@@ -63,10 +63,20 @@ let allow_unused_code_warnings = ref Options.default_allow_unused_code_warnings
 
 let () =
   Driver.add_arg "-unused-code-warnings"
-    (Bool (( := ) allow_unused_code_warnings))
-    ~doc:"_ Allow ppx derivers to enable unused code warnings"
+    (Symbol
+       ( [ "true"; "false"; "force" ],
+         function
+         | "true" -> allow_unused_code_warnings := True
+         | "false" -> allow_unused_code_warnings := False
+         | "force" -> allow_unused_code_warnings := Force
+         | _ -> assert false ))
+    ~doc:" Allow ppx derivers to enable unused code warnings (default: false)"
 
-let allow_unused_code_warnings () = !allow_unused_code_warnings
+let allow_unused_code_warnings ~ppx_allows_unused_code_warnings =
+  match !allow_unused_code_warnings with
+  | Force -> true
+  | False -> false
+  | True -> ppx_allows_unused_code_warnings
 
 module Args = struct
   include (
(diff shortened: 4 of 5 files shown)


mbarbin:
By the way, something that didn't occur to me initially, if you prefer this can be merged into the existing flag, by making the cli ui `-unused-code-warnings=force` rather than adding a new flag.

mbarbin:
Noting some action item for me to look into next on this PR:

1. Rebase on top of #492 
2. Look into merging the value `force` into the existing flag, making sure not to break when the flag is passed with no value (look into how stdlib's arg works, `Arg.Bool`, `Arg.String`, etc.)

mbarbin:
@NathanReb thanks for #492 ! I like you proposed new tests a lot, and found it makes adding tests for this PR easier.

By the way, do you like to use cram tests as [characterization tests](https://en.wikipedia.org/wiki/Characterization_test) in the ppxlib project? If so, I'd like to propose to include to the new `run.t` examples characterizing the use of `deriving-keep-w32` and how it combines with the existing `-unused-code-warnings`. I've added one section for it in this PR but if you'd like I can make this bit separate.

NathanReb:
I think the `-deriving-keep-w32` is indeed completely untested and characterization tests for it would be a welcome addition! `test/deriving_warning/run.t` also seems like the right place for such tests.

If it's not asking too much though, I would prefer if those were added in a separate PR from this one!

NathanReb:
@mbarbin do you want to keep working on this? I'd like to include this into the next release!

I'm of course happy to take over if you're too busy at the moment!

mbarbin:
@NathanReb One more thing while I am thinking about it : do as it is the most convenient for you, please do not worry about keeping or rebasing my commits, especially if it is just simpler to start over from a clean slate in another PR. I can close my draft and/or salvage things I'd like to extract later. Thank you!

mbarbin:
Hi @NathanReb just the heads up I found some time to make progress on the action items and pushed an update.

NathanReb:
Awesome, I'll take a look!

NathanReb:
@mbarbin could you rebase on top of main? I'm actually thinking about dropping the `-unused-type-warnings` flag and instead disable all warnings through this single flag.

The type one is likely to be harmless to the vast majority of users so I'm not sure how worth it is to complexify this whole thing with yet another flag.

Some users want the same transition scenario than with the other warnings so it also allows to reuse the `Deriving.make` argument without modifying the API.

src/options.ml
```diff
@@ -1,4 +1,8 @@
-let default_allow_unused_code_warnings = false
+module Allow_unused_code_warnings = struct
```
NathanReb, line 1: What do you think about using a more generic name here, such as `Force_flag`?

NathanReb (commented):
The code looks good to me! Thanks for taking the time to add this feature!

Except for the `Options` submodule name this is good to go once we've rebased it on top of main. I'll work on adding the unused type warning control under this same flag!

mbarbin:
@NathanReb thank you for the review!

> @mbarbin could you rebase on top of main? I'm actually thinking about dropping the `-unused-type-warnings` flag and instead disable all warnings through this single flag.

The latest update rebased on top of the `-unused-type-warnings` commit. I can create a version with these parts left off, however please read on.

> The type one is likely to be harmless to the vast majority of users so I'm not sure how worth it is to complexify this whole thing with yet another flag.

I agree with the harmless part of the type one, however I am under the impression that disabling the code warning is likely to require more work on the user side of things. For example, I tried on my code base to use `-deriving-keep-w32=both` and noticed this would require some non trivial number of tweaks to get it adoptable (I'm thinking that `unused-code-warnings=force` is also going to require some similar work). I was hoping I would be able to close #473 with the next release of ppxlib, as an incremental step without having to go on the journey to solve all ppxlib related unused code warnings just yet (I still hope to do this as future work!).

If you are thinking about dropping the ability to control the unused type warning in isolation, then, I would advocate for resurrecting the PR #495 that changes the nature of the code, so we can close #473.

> Some users want the same transition scenario than with the other warnings so it also allows to reuse the `Deriving.make` argument without modifying the API.

This sounds interesting. Can you say a bit more about this, I'd like to understand this in more details.

Thanks a lot!

src/options.ml
mbarbin: I like it, thanks! Done.

mbarbin:
> @mbarbin could you rebase on top of main?

@NathanReb I removed the part about `-unused-type-warnings` and rebased on top of main.

NathanReb:
Ok I undestand your point about merging the two flags and I agree user will have an easier time disabling the unused type warning silencing than the other two.

How about making the `-unused-code-warnings` flag control all three of warnings 32, 34 and 60 silencing and adding a separate `-unused-type-warnings` flag that controls only the warning 34 one?

> This sounds interesting. Can you say a bit more about this, I'd like to understand this in more details.

It's simply the feature where both the user and the derivers being used need to opt out of the warning silencing (except when `force` is passed of course) for it to be disabled, as you might have read in the tests suite. I think it'll be simpler to control all warnings from that same argument.

The idea is that users can pass the flag with `true` and it will initially have no impact. As they patch the derivers they use so that they opt out, the driver won't silence warnings for code generated for by those but will keep generating it for other derivers. This allow to quickly opt out for the ones that weren't causing any troubles while taking the time to fix the derivers that need changes in order not to cause unwanted warnings. Ultimately once all derivers have been fixed and have opted out, only the user decides whether to silence warnings or not. At that point, we can decide to change the default, both for the flag and for the `Deriving.make` argument (potentially in two separate steps) so that we don't silence warnings unless the user or deriver opt in.

NathanReb (approved):
Looks good, thanks!

mbarbin:
> How about making the `-unused-code-warnings` flag control all three of warnings 32, 34 and 60 silencing and adding a separate `-unused-type-warnings` flag that controls only the warning 34 one?

If I understand correctly, this is a slight twist to the earliest proposal where `-unused-code-warnings` controls 32 & 60, and `-unused-type-warnings` controls 34:

- This allows users that desire to do that, to treat the type parts (34) as a first and independent step 
- while allowing things to stay simple (one single flag) for those who want to treat all warnings (32, 34, 60) in the same step.

This sounds good to me!

PR #494, opened 2024-05-29
Context_free.special_function: fix the parsing of indexing operators

Octachron:
Currently, indexing operators are misparsed by ppxlib in `Context_free.special_function` because the `Longident` parsing code stops on the first closing parenthesis in `(.!(;..))` and fails with an invalid argument because this first closing parenthesis is not at the end of the identifier.

This PR fixes this issue by finding the first pair of matched parentheses and add a test for the indexing operator cases.
Along the way, I added some more context on the `Invalid_argument` error.

Note that I am also tempted to propose a version of `special_function` that doesn't parse its argument and takes a `longident` directly to avoid this recurring issue.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 unreleased
 ----------
 
+- Fix `Longident.parse` so it also handles indexing operators such as
+  `.!()`, `.%(;..)<-`, or `Vec.(.%())` (#494, @octachron)
+
 0.32.1 (2024-04-23)
 -------------------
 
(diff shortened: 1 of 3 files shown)


NathanReb (commented):
Looks good, thanks for taking the time to fix this!

Could you add a changelog entry mentioning the fix?

Octachron:
I have added a change entry added, with a off-by-one fix and some new tests for unbalanced parentheses (e.g "(").

NathanReb (approved):
Tyvm!

PR #496, opened 2024-05-30
Context_free: add a special_function' variant

Octachron:
Parsing `Longindent.t`s is quite unpractical because `Longident.t` are used as an IR for different and distinct syntaxic classes. This means that interpreting the string argument of the `Context_free.special_function` is less trivial than it appears.

To circumvent this issue, this commit takes the view that in complex cases the user themself can still easily specify the identifier that they intended to use.

Thus it sounds sensible to provides a `special_function'` variant function which takes directly a `Longident.t` argument for advanced uses.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@ unreleased
 - Fix `Longident.parse` so it also handles indexing operators such as
   `.!()`, `.%(;..)<-`, or `Vec.(.%())` (#494, @octachron)
 
+- Add a `special_function'` variant which directly takes a `Longident.t`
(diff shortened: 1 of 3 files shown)


NathanReb (approved):
Looks good, thanks!

NathanReb:
I missed that the DCO check's unhappy. If you can fix this, I'll merge right after!

Octachron:
The identity confusion and the location of the change entry should be fixed now.

NathanReb:
I just fixed the formatting, will merge when the CI's green.

PR #499, opened 2024-06-10
update stale bits of CONTRIBUTING.md

v-gb:


CONTRIBUTING.md
```diff
@@ -18,21 +18,11 @@ below for details.
...
-points:
-
-- all the modified code is correctly indented according to ocp-indent
-- `make` succeeds
-- `make test` succeeds
+Before submitting a PR, please run `dune build @install @runtest @fmt`
+on your machine. In addition to normal dependencies, you'll need to
+install [cinaps][cinaps] (to keep up-to-date some parts of the code
+that are auto-generated and committed in the repository), and
+`ocamlformat`.
```
NathanReb, line 25: `ocamlformat` needs to be installed in a specific version, how about mentioning this here?
```suggestion
`ocamlformat` (make sure to install the version matching the info in
[`.ocamlformat`](https://github.com/ocaml-ppx/ppxlib/blob/main/.ocamlformat))
```

NathanReb (commented):
Thanks for taking the time to update this!

CONTRIBUTING.md
v-gb: Ok, but I think we should be giving commands to run, not text that people have to interpret and figure how to make happen. I did that.

NathanReb: Yes but as you noticed, commands such as these are often forgotten about and become obsolete.

We'll have to consider adding `ocamlformat` to the opam file somehow to be able to provide a reliable command. I'll look into this.

NathanReb (approved):

CONTRIBUTING.md
v-gb: Text gets stale, but stale commands is much better than stale descriptions, because stale commands will fail immediately (which is then clearly a bug in the repo, whereas stale descriptions waste time to understand how to translate them into commands, only for that to fail, and even then it's not clear if you understood what to do right.

Obviously, the ideal solution would be for dune+opam to make `dune build` just grab whatever the build needs. Maybe the `dune pkg` work will make that happen, not sure. In the meantime, I agree that a `Makefile` command is a bit more likely to be kept up-to-date.

NathanReb: I'm hoping we can use this https://github.com/ocaml/opam/issues/4959 to simply make an opam command work here. I'll merge what we already have in the meantime!

PR #501, opened 2024-06-12
Add a trunk-build github action

NathanReb:
Adds an action that builds with the latest supported trunk. We'll need this for incremental trunk support and I'd like to see it run on #487.

I heavily reused the work of @shym from #410, thank you for this contribution!

### DIFF
--- a/.github/workflows/trunk-build.yml
+++ b/.github/workflows/trunk-build.yml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Pick up a robust cache prefix
         id: setup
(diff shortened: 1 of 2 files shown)


NathanReb:
The trunk build fails atm of course because our opam bounds don't allow for OCaml 5.3.

.github/workflows/trunk-build.yml
```diff
@@ -0,0 +1,51 @@
+name: Build on `trunk`
+
+on:
+  push:
```
patricoferris, line 4: Do we want this to also run on just _any_ push to _any_ branch? Or only on pushes to `main` and pulls?

patricoferris (approved):
This looks good :))

.github/workflows/trunk-build.yml
NathanReb: I'd say only on push events to `main`, I'll update the action!

NathanReb: For PRs I'd keep any branch though, does that sound good?

PR #504, opened 2024-06-30
Fix issue 503

### LINKED ISSUE #503: Badly formed locations for `Pexp_fun` after 5.2 leads to Merlin confusion
Hi! Thanks for your work on ppxlib ❤️ 

We have stumbled upon this problem. Consider this file `input.ml`:

```ml
let make ~foo ~bar = foo ^ bar
```

Before 5.2, the locations for this file were the following (I cleaned up the `ocaml.ppx.context` part as it's not relevant for the issue and formatted them):

[locations_5_1.txt](https://github.com/user-attachments/files/16045500/locations_5_1.txt)

In 5.2, the locations are as follow:

[locations_5_2.txt](https://github.com/user-attachments/files/16045501/locations_5_2.txt)

Here's the diff between both files for clarity:

```diff
174,175c174,175
<               ((pos_fname input.ml) (pos_lnum 1) (pos_bol 0) (pos_cnum 30)))
<              (loc_ghost true)))
---
>               ((pos_fname input.ml) (pos_lnum 1) (pos_bol 0) (pos_cnum 18)))
>              (loc_ghost false)))
```

This location corresponds to the `Pexp_fun` ast node that corresponds to the `~bar` argument. The change of location is problematic, because this AST node locations ends up being 14-18, while it has children that exceed that range, like a `Pexp_apply` which is immediate child with locations 21-30. Badly formed locations like this leads to issues with Merlin, in our experience.

Is this something possible due to the 5.1 -> 5.2 migration? is there anything we could do on our side to work around this problem? See related issue: https://github.com/reasonml/reason-react/issues/840

jchavarri:
Fixes #503.

In the first commit, notice the difference in the ending location of the `Pexp_fun` with the labelled `bar`.

In 5.1:
https://github.com/ocaml-ppx/ppxlib/blob/4661a93e6029f86bd3e0d3fa6d519fe4c658a6dc/test/501_migrations/one_migration.t#L1726

In 5.2:
https://github.com/ocaml-ppx/ppxlib/blob/4661a93e6029f86bd3e0d3fa6d519fe4c658a6dc/test/502_migrations/one_migration.t#L719

The second commit updates `Migrate_502_501` so that the ranges of locations of `Pexp_fun` stay consistent with the behavior in 5.1.


NathanReb (commented):
hanks for taking the time to report and fix this bug!

If I understand it correctly you should apply the same change to the `Pexp_newtype` node below as well!

The tests for the 501<->502 migrations are pretty unreadable in their current state unfortunately and I'd rather avoid adding more of those. I suggest you instead take an approach closer to how we test the `ocaml.ppx.context` migration (see the [`driver.ml`](https://github.com/ocaml-ppx/ppxlib/blob/main/test/driver/ocaml-ppx-context-load-path-migration/driver.ml), [`dune`](https://github.com/ocaml-ppx/ppxlib/blob/main/test/driver/ocaml-ppx-context-load-path-migration/dune) and [`run.t`](https://github.com/ocaml-ppx/ppxlib/blob/main/test/driver/ocaml-ppx-context-load-path-migration/run.t) files). The idea is that you'd move both your tests to their own folder and write a custom no-op (as in, no actual transformaion) driver that run specific migrations and could here, display the location of specific bits of the AST, such as the `Pexp_fun` with the `bar` argument on the standard output.

I'm happy to help with the refactoring of those tests if needed so please let me know!

jchavarri:
Hi @NathanReb thanks for the quick review :)

I applied the following changes:

- In https://github.com/ocaml-ppx/ppxlib/pull/504/commits/b8e11706d460c20929286af2f3dea4d217c41ba2, I reverted the fix and added a small exe that checks the locations integrity for the given AST nodes
- In https://github.com/ocaml-ppx/ppxlib/pull/504/commits/6cc652ce3c322c6c913f9325fb75e9d02e40c53a, I re-applied the fix and can see how the tests errors are gone

I also removed the previous location tests.

Please lmk if something like this is what you had in mind.

NathanReb (approved):
This is exactly what I had in mind, thanks for taking the time to refactor it and add the changelog entry!

PR #506, opened 2024-07-05
Fix -locations-check behaviour

NathanReb:
It used to be locked behind the `-check` flag as well, despite what the documentation stated. We might want to do the same for the check-on-extension as well.

It's worth noting that all checks are disabled by default so there's no problem with `-no-check` not disabling the location check anymore.


pitag-ha (approved):
Sounds very good, thanks!

In my opinion,  `-location-check` is underused at the moment. Having it locked into `-check` made things very confusing on top of it.

PR #508, opened 2024-07-05
Simplify 502 migration location check test

NathanReb:
The test can be greatly simplified by using the driver's native location check instead!


NathanReb:
This is based on top of #506

PR #510, opened 2024-07-15
Make `-unused-code-warnings` also controls warning 34 suppression items

mbarbin:
Instead of only controlling the generation of items to suppress warnings 32 and 60, makes this flag (and the respective Deriving arg) also control the item generated to suppress warning 34 (unused type declaration).

This is based on a recent discussion with @NathanReb in #490 .

### DIFF
--- a/src/deriving.ml
+++ b/src/deriving.ml
@@ -253,7 +253,11 @@ module Generator = struct
       List.concat lerr
       |> List.map ~f:(fun err -> ext_to_item ~loc:Location.none err [])
     in
-    List.concat l1 @ [ { items = lerr; unused_code_warnings = false } ]
+    List.concat l1
+    @
+    match lerr with
+    | [] -> []
+    | _ :: _ as lerr -> [ { items = lerr; unused_code_warnings = false } ]
 end
 
 module Deriver = struct
@@ -712,8 +716,13 @@ let wrap_sig ~loc ~hide list =
    | Main expansion                                                  |
    +-----------------------------------------------------------------+ *)
 
-let types_used_by_deriving (tds : type_declaration list) : structure_item list =
-  if keep_w32_impl () then []
+let types_used_by_deriving (tds : type_declaration list) ~unused_code_warnings :
+    structure_item list =
+  let unused_code_warnings =
+    allow_unused_code_warnings
+      ~ppx_allows_unused_code_warnings:unused_code_warnings
+  in
+  if keep_w32_impl () || unused_code_warnings then []
   else
     List.map tds ~f:(fun td ->
         let typ = Common.core_type_of_type_declaration td in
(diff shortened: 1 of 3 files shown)


src/deriving.ml
```diff
@@ -253,7 +253,11 @@ module Generator = struct
       List.concat lerr
       |> List.map ~f:(fun err -> ext_to_item ~loc:Location.none err [])
     in
-    List.concat l1 @ [ { items = lerr; unused_code_warnings = false } ]
+    List.concat l1
+    @
+    match lerr with
+    | [] -> []
+    | _ :: _ as lerr -> [ { items = lerr; unused_code_warnings = false } ]
```
mbarbin, line 260: @NathanReb my reasoning here was :

1. Inserting `unused_code_warnings=false` impacts what the new code does since after these items are merged with the rest of the one generated by a ppx, `types_used_by_deriving` doesn't distinguish it from ppx generated ones, and believes one of the ppx intends not to allow the warnings. Thus, it is better to omit this extra element in the list if the `lerr` list to be inserted is empty (and result in the same generated items overall).

2. In case `lerr` is not empty, (this part I am more fuzzy on), I am guessing this is a mechanism designed in ppxlib to avoid stopping on the first error reported by one ppx. I made the hypothesis that in this case the entire ppxchain is going to be aborted in the end due to processing errors, and thus the impact of `unused_code_warnings` can be ignored in this code path (when errors are emitted).

src/deriving.ml
```diff
@@ -712,8 +716,13 @@ let wrap_sig ~loc ~hide list =
...
 
-let types_used_by_deriving (tds : type_declaration list) : structure_item list =
-  if keep_w32_impl () then []
+let types_used_by_deriving (tds : type_declaration list) ~unused_code_warnings :
+    structure_item list =
+  let unused_code_warnings =
+    allow_unused_code_warnings
+      ~ppx_allows_unused_code_warnings:unused_code_warnings
+  in
+  if keep_w32_impl () || unused_code_warnings then []
```
NathanReb, line 725: We should probably drop the `keep_w32_impl ()` here, it's clearly a bug to control the unused type warnings via this flag. I'll remove it in a subsequent PR!

src/deriving.ml
NathanReb: I would suggest a slightly different approach here. I think it makes things clearer if we determine whether we should silence warning 34 from the list of generators rather than from the wrapped generated items.

Getting it directly from the field that determines whether they opted out is safer than getting it from a field that derives from that. I can imagine a scenario where changes in the code could inadvertently break this behaviour, as suggests this specific change for instance.

Does that sound good to you?

NathanReb (commented):
Thanks for taking the time to add this feature, really appreciate your contributions!

Except for the way we determine if all derivers opted out, it looks good to me!

CC @ceastlund, you might be intrested in taking a look. I kind of changed my mind and decided to control warning 34 with the same flag and `Deriving.make` argument as you initially suggested. We'll still add a separate flag that allows only disabling the warning 34 silencing as some users (such as @mbarbin) are interested in getting this quick win out of the way without having to fix the other warnings.

src/deriving.ml
mbarbin: This sounds good to me. After the set of changes we intend to include in the release, their will be a clear replacement for this (`-unused-type-warnings`). It is not likely to break existing usage, as fixing this bug could only remove warnings rather than adding new ones (right?).

NathanReb: Yep that sounds about right!

src/deriving.ml
mbarbin: Oh, of course yes. I missed that. Thanks a lot! Done.

NathanReb (approved):
Looks good to me, thanks a lot!

PR #512, opened 2024-07-16
Fix extension removal bug in `Code_path.main_module_name`

### LINKED ISSUE #403: `Code_path.main_module_name` is sensitive to line directives
`Code_path.main_module_name` is sensitive to line directives, while it should directly use the name of the input file. Reported by @sim642 in https://github.com/ocaml-ppx/ppx_deriving/pull/263#discussion_r924308810

NathanReb:
Fixes #403.

I'm not sure whether `Code_path` should ignore line directives or not but I do believe it should properly strip extensions. We can always make a different decision about line directive later if do causes trouble beyond this.

### DIFF
--- a/src/code_path.ml
+++ b/src/code_path.ml
@@ -10,9 +10,14 @@ type t = {
   in_expr : bool;
 }
 
+let remove_all_extensions basename =
+  match String.split_on_char ~sep:'.' basename with
+  | [] -> assert false
+  | name :: _ -> name
+
 let top_level ~file_path =
   let main_module_name =
-    file_path |> Stdlib.Filename.basename |> Stdlib.Filename.remove_extension
+    file_path |> Stdlib.Filename.basename |> remove_all_extensions
     |> String.capitalize_ascii
   in
   {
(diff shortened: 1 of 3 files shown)


src/code_path.ml
```diff
@@ -10,9 +10,14 @@ type t = {
   in_expr : bool;
 }
 
+let remove_all_extensions basename =
+  match String.split_on_char ~sep:'.' basename with
+  | [] -> assert false
```
patricoferris, line 15: Just noting that `Stdlib.Filename.remove_extension` doesn't raise in the case where there is no extension, it just passes the filename through. Not sure if we want to do the same here or not.

patricoferris (approved):

src/code_path.ml
NathanReb: This won't raise as this is an invariant of `String.split_on_char`, see: https://ocaml.org/manual/5.2/api/String.html#extract.

I'll add a comment here, you're right that it was a bit bold to just dump an `assert false` here!

patricoferris: The more you know! Thanks :))

PR #516, opened 2024-08-19
Add ocamlformat as dev dependency

NathanReb:
With opam's 2.2.0 release, we now have a new `--with-dev-setup` flag and corresponding opam file var which we can use to declare dev dependencies such as ocamlformat!


patricoferris (approved):
Cool! Been waiting for the `dev` flag for a while ! :)

PR #517, opened 2024-08-22
Add Pp_ast module and ppxlib-pp-ast executable for pretty-printing ppxlib ASTs

NathanReb:
This adds pretty-printing functions to our API along with a simple executable that can be used to pretty-print input `.ml`, `.mli`, expressions, patterns or core types.

It's API is similar to `ppx_tools`'s `dumpast` except that it can also read binary ASTs of any version which I assume could come in handy in certain situation.

The printing style is also similar to dumpast:
- It prints `expression`s directly as their `pexp_desc` field and does so for any other such record with a `_desc` field
- It does not print locations, positions or attributes to avoid cluttering the output
- It prints `'a loc` types directly as `'a`

In the future we can add a bit of configuration there to control what gets printed or not but it is already pretty useful as it is.

I'd even consider using this with our `-dparsetree` driver option instead of the current sexp printer.


NathanReb:
Just added a config option to allow showing attributes as it turns out I needed this to investigate a migration bug.

bin/pp_ast.ml
```diff
@@ -0,0 +1,122 @@
...
+let show_attrs = ref false
+
+let set_input fn =
+  match !input with
+  | None -> input := Some fn
+  | Some _ -> raise (Arg.Bad "too many input files")
+
+let set_kind k =
+  match !kind with
+  | Some _ -> raise (Arg.Bad "must specify at most one of --impl or --intf")
```
patricoferris, line 67: Think this should be

```suggestion
  | Some _ -> raise (Arg.Bad "must specify at most one of --str, --sig, --exp, --pat or --typ")
```

src/pp_ast.mli
```diff
@@ -0,0 +1,27 @@
...
+
+  val make : ?show_attrs:bool -> unit -> t
+  (** Create a custom pretty-printing config.
+
+      Default values are the ones that are used when no configuration is passed
+      to the pretty-printers defined below.
+
+      @param show_attrs
+        controls whether attributes are shown or hidden. It defaults to [false].
+        When set to [true], records such as [expression] that have a [desc]
```
patricoferris, line 15: I'm a bit confused by this explanation, in particular

> When set to [true], records such as [expression] that have a [desc] field will only be printed if the list of attributes is non-empty

```suggestion
        When set to [true], attribute fields on records such as [expression] that have a [desc]
```

And even if the attributes are empty, it still prints them (i.e. it prints `[]` and not `__attrs`)

bin/pp_ast.ml
```diff
@@ -0,0 +1,122 @@
+open Ppxlib
+
+module Kind = struct
+  type t = Signature | Structure | Expression | Pattern | Core_type
+
+  let to_utils_kind = function
+    | Structure -> Ppxlib__Utils.Kind.Impl
```
patricoferris, line 7: Do you think Utils should be properly exposed behind a `Ppxlib.Private` module?

patricoferris (approved):
This looks great! I think waiting for the migration bug fix before merging might help uncover more useful features to add (e.g. you needed the `--show-attrs` configuration flag).

bin/pp_ast.ml
NathanReb: Yeah that might be a good idea!

src/pp_ast.mli
NathanReb: Yeah this is very poorly phrased indeed. What I was trying to say was:

1. If you don't pass `--show-attr=true`, then records with a `_desc` field are directly printed as the value of said field.
2. If you pass `--show-attr=true`, records with a `_desc` field and an empty list of attributes are still printed as the value of the `_desc` field.
3. If you pass `--show-attr=true`, records with a `_desc` field and a non empty list of attributes are printed as a regular record would.

> And even if the attributes are empty, it still prints them (i.e. it prints [] and not __attrs)

That's true but the behaviour is governed at the record level, i.e. if you look at how `expression`, it's defined using this function https://github.com/ocaml-ppx/ppxlib/pull/517/files#diff-cc3c0a60fe8fd76db7bb4c7d196c81e1a969784fbc96201434a5c3017aa0578aR104 which handles how to print `_desc` field containing records based on the value of `show_attrs` and the presence of `attributes`.

I'd gladly take a better phrasing to describe this behaviour. It might be best to describe it at a higher level of documentation for `ppxlib-pp-ast`, especially since one could be surprised not to see those records "skipped" regardless of the `--show-attr` specific behaviour.

I'll try to do that!

NathanReb:
I'll try to address the documentation related concerns and the exposure of private API parts for internal use but then I think this is good to go!

NathanReb:
I added loc printing and relevant configuration!

NathanReb:
I'm also wondering whether the tool should be installed as part of ppxlib's main package or a separate, ppxlib-tools one.

NathanReb:
I improved the documentation a bit, let me know what you think @patricoferris !

patricoferris:
The new documentation looks great! 

> I'm also wondering whether the tool should be installed as part of ppxlib's main package or a separate, ppxlib-tools one.

`ppxlib-tools` would just be to attach some executables too?

NathanReb:
> ppxlib-tools would just be to attach some executables too?

Yep, in order to keep `ppxlib` minimal as not every ppx user would be interested in `ppxlib-pp-ast`.

NathanReb:
In addition, we could also extend the set of dependencies for such tools to things like `cmdliner` for instance, whereas we try to keep the set of dependencies for `ppxlib` to a bare minimum.

NathanReb:
This should be the final round, I extracted `ppxlib-pp-ast` to a separate package and rewrote the tool itself using `cmdliner`.

I also added the ability to read source code directly from the command line which can be extremely convenient to quickly understand how to represent simple expressions or types for instance. This feature was part of `ppx_tools`' `dumpast` and I quickly realized it wasn't very convenient to go through a file or `stdin` for the simplest use cases.

I also took the time to write a bit of documentation in the manpage so hopefully the tool will be easier to use.

bin/pp_ast.ml
```diff
@@ -0,0 +1,219 @@
...
+      match t with
+      | Stdin -> Stdppx.In_channel.input_all stdin
+      | File fn -> Stdppx.In_channel.(with_file fn ~f:input_all)
+      | Source s -> s
+    in
+    Lexing.from_string all_source
+
+  let from_string = function
+    | "-" -> Stdin
+    | s when Sys.file_exists s -> File s
```
patricoferris, line 35: Do we have something lying about for doing `file_exists` checks that isn't this? IIRC `Sys.file_exists` converts _any_ error into `false` (e.g. a permission error) which is confusing for an end user maybe ? If we don't have anything easily at hand then this is fine.

patricoferris (approved):
This looks great to me! Happy for formatting bugs to be follow up PRs if they do arise, this would already be very useful for debugging migration issues!

test/ppxlib-pp-ast/dune
```diff
@@ -0,0 +1,5 @@
+(cram
+ (package ppxlib-tools)
+ (deps
+  ../../bin/pp_ast.exe
```
patricoferris, line 4:
```suggestion
%{bin:ppxlib-pp-ast}
```

I think we can change this -- if you do a `dune clean` and `dune runtest` directly then this won't pull in the right binary for the tests without explicitly depending on it like this.

NathanReb: Good catch! I'll give that a spin

bin/pp_ast.ml
NathanReb: We might have something in `stdppx`, I'll take a look!

NathanReb: From a quick look it doesn't seem like we have something satisfying lying around, we're already using `Sys.file_exists` in a few places.

I suggest we leave it at that for this PR but we can eventually add a better version to `stdppx` later on if needed.

Issue #435, opened 2023-06-25
`Ast_builder.Default.pexp_tuple` on empty list


sim642:
`Ast_builder.Default.pexp_tuple`, etc. currently have a special case for the one-element list:
https://github.com/ocaml-ppx/ppxlib/blob/6baf858275937bfafa05c8ce958f0e75ced00262/src/ast_builder.ml#L91-L99
I wonder if it makes sense to also have a special case for the empty list, degrading to `()` instead. Right now calling them with the empty list yields an error from deeper down that a tuple needs at least 2 elements.

Or maybe there's a reason this generalization hasn't been made? The existence `_opt` versions of these functions hints at the need to distinguish the empty list case.

pitag-ha:
Thanks for the issue! I wasn't sure myself, so I've just had a look at the compiler Parsetree module. There are few Parsetree invariants that aren't enforced by the type checker. They're written down as comments in the compiler Parsetree module. There's one for `Pexp_tuple`: [n>=2](https://github.com/ocaml/ocaml/blob/trunk/parsing/parsetree.mli#L342). That explains the behavior of `ppat_tuple` and why `ppat_tuple_opt` exists.

It would be quite valuable to add documentation about that in `Ast_builder` though. Do you want to do that?

PR #518, opened 2024-08-30
Ast_builder documentation

patricoferris:
Related to: https://github.com/ocaml-ppx/ppxlib/issues/324, https://github.com/ocaml-ppx/ppxlib/issues/435

Much of `Ast_builder` is undocumented. The names of functions are derived from the things they construct from `Ast.ml` (mostly AST nodes from `Parsetree`). The parsetree is well-documented. This PR pulls through that documentation directly to `Ast_builder`.

In order for this to work well, the `gen_ast_builder` script now produces module types for the interface to `Ast_builder` (one for functions with `~loc` arguments called `S_located` and one without, corresponding to the original `S`).

Before:

Screenshot 2024-08-30 at 14 53 03

After:

Screenshot 2024-08-30 at 14 49 12

_Known problems_:

Some of the doc attributes pulled from the parsetree contain odoc references which clash with the function names defined in the generated code e.g. `type label_declaration` and `val label_declaration` clash for the documentation comment attached to `ptyp_poly`.

---

If there is agreement to add this we could do the same for `Ast_pattern` I'm sure.

### DIFF
--- a/src/gen/gen_ast_builder.ml
+++ b/src/gen/gen_ast_builder.ml
@@ -60,11 +60,9 @@ struct
     | [] -> M.ctyp "%s" typ_name
     | params ->
         let params =
-          List.fold_left params ~init:"" ~f:(fun acc (ctyp, _) ->
-              Format.asprintf "%a" A.ctyp ctyp ^ ", " ^ acc)
+          List.map params ~f:(fun (ctyp, _) -> Format.asprintf "%a" A.ctyp ctyp)
         in
-        let params = String.sub params ~pos:0 ~len:(String.length params - 2) in
-        M.ctyp "(%s) %s" params typ_name
+        M.ctyp "(%s) %s" (String.concat ~sep:", " params) typ_name
 
   let gen_combinator_for_constructor
       ~wrapper:(wpath, wprefix, has_attrs, has_loc_stack) path ~prefix
--- a/src/gen/gen_ast_builder.ml
+++ b/src/gen/gen_ast_builder.ml
@@ -248,6 +248,26 @@ let dump fn ~ext printer x =
   Format.fprintf ppf "%a@." printer x;
   close_out oc
 
+let floating_comment s =
+  let doc =
+    PStr
+      [
+        {
+          pstr_desc =
+            Pstr_eval
+              ( {
+                  pexp_desc = Pexp_constant (Pconst_string (s, loc, None));
+                  pexp_loc = loc;
+                  pexp_loc_stack = [];
+                  pexp_attributes = [];
+                },
+                [] );
+          pstr_loc = loc;
+        };
+      ]
+  in
+  Sig.attribute (Attr.mk { txt = "ocaml.text"; loc } doc)
+
 let generate filename =
   (*  let fn = Misc.find_in_path_uncap !Config.load_path (unit ^ ".cmi") in*)
   let types = get_types ~filename in
@@ -320,7 +340,7 @@ let generate filename =
                 Bytes.set bs 0 (Char.uppercase_ascii @@ Bytes.get bs 0);
                 String.concat ~sep:" " (Bytes.to_string bs :: rest)
           in
-          (M.sigi "(** {2 %s} *)" label :: items) @ acc)
+          (floating_comment (Format.asprintf "{2 %s}" label) :: items) @ acc)
         (mod_sig_items located) []
     in
     let items =
--- a/src/gen/gen_ast_builder.ml
+++ b/src/gen/gen_ast_builder.ml
@@ -120,11 +120,9 @@ struct
         let body =
           if fixed_loc then body else M.expr "fun ~loc -> %a" A.expr body
         in
-        let str =
-          M.stri "let %a = %a" A.patt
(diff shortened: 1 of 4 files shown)


src/gen/gen_ast_builder.ml
```diff
@@ -100,11 +178,41 @@ struct
            body
        in*)
     let body =
-      if List.mem "loc" ~set:funcs && not fixed_loc then
+      if List.mem "loc" ~set:(List.map ~f:snd funcs) && not fixed_loc then
```
NathanReb, line 181: This appears to be checked for at least twice, how about extracting:
```ocaml
let has_loc_field =
  List.exists (function (_, "loc") -> true | _ -> false) funcs
in
```

src/gen/gen_ast_builder.ml
```diff
@@ -117,11 +225,11 @@ struct
...
                 common_prefix (List.map cds ~f:(fun cd -> cd.pcd_name.txt))
               in
-              List.map cds ~f:(fun cd ->
-                  gen_combinator_for_constructor ~wrapper path ~prefix cd))
+              List.map cds
+                ~f:(gen_combinator_for_constructor ~wrapper path ~prefix td))
       | Ptype_record lds ->
           let prefix = prefix_of_record lds in
-          [ gen_combinator_for_record path ~prefix lds ]
+          [ gen_combinator_for_record path ~prefix td td.ptype_attributes lds ]
```
NathanReb, line 232: We could just pass `td` here and let `gen_combinator_for_record` extract the attributes from it couldn't we?

src/gen/gen_ast_builder.ml
```diff
@@ -140,6 +248,26 @@ let dump fn ~ext printer x =
...
+                  pexp_loc = loc;
+                  pexp_loc_stack = [];
+                  pexp_attributes = [];
+                },
+                [] );
+          pstr_loc = loc;
+        };
+      ]
+  in
+  Sig.attribute (Attr.mk { txt = "ocaml.text"; loc } doc)
```
NathanReb, line 269: Can't we simplify this with `M.sigi`?

NathanReb (commented):
This looks great! Awesome addition!

I have a few minor code comments but other than that I think it's a great idea and a very useful addition to the documentation.

On a slightly unrelated note, I wonder if we shouldn't change the way we generate those files to promote the generated sources so they can be inspected and we are notified whenever they are updated rather than all of it being silent as it is now. That's of course a matter for a separate PR but I'm still curious to know what you think about it!

src/gen/gen_ast_builder.ml
```diff
@@ -11,8 +48,27 @@ end) =
...
+      | "desc" :: _ ->
+          String.sub ~pos:0 ~len:(String.length typ_name - 5) typ_name
+      | _ -> typ_name
+    in
+    match typ.ptype_params with
+    | [] -> M.ctyp "%s" typ_name
+    | params ->
+        let params =
+          List.fold_left params ~init:"" ~f:(fun acc (ctyp, _) ->
+              Format.asprintf "%a" A.ctyp ctyp ^ ", " ^ acc)
```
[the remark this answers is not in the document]
patricoferris: Thanks for catching this -- I guess there's no type in Ast.ml with more than one type variable!

src/gen/gen_ast_builder.ml
```diff
@@ -66,31 +122,53 @@ struct
...
-          A.expr body
+        let str =
+          M.stri "let %a = %a" A.patt
+            (pvar (function_name_of_id ~prefix cd.pcd_name.txt))
+            A.expr body
+        in
+        let return_type = core_type_of_return_type return_type in
+        let typ =
+          List.fold_right cd_args ~init:return_type ~f:(fun cty acc ->
+              M.ctyp "%a -> %a" A.ctyp cty A.ctyp acc)
```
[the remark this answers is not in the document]
patricoferris: Yeh? I was just copying the existing code for constructing the body for the implementation (just above)?

src/gen/gen_ast_builder.ml
patricoferris: Yep

src/gen/gen_ast_builder.ml
patricoferris: The parser seemed to feel on OCaml 4.04 and 4.05 when parsing a floating documentation comment no matter what way I expressed it (`(** {2 ...} *)`) or with the attribute

src/gen/gen_ast_builder.ml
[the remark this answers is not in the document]
NathanReb: Hmm right, I guess we can leave this out for now for the sake of consistency but we should come back to this in the future!

src/gen/gen_ast_builder.ml
NathanReb: Ah ok, good to know! Let's keep it that way then!

NathanReb (approved):
Looks good to me! Let's follow up on promoting the generated files and optimizing the parse/print loops in a few places!

I'll open issues for both!

NathanReb:
@patricoferris I think you can go ahead and merge this!

PR #519, opened 2024-09-04
Fix roundtrip check to support 5.1 <-> 5.2 migrations

NathanReb:
Printing to source code and parsing with the compiler before running the upward migration is expected to fail because
`fun x y -> z` and `fun x -> fun y -> z` parse to the same AST on 5.01- but have different representations on OCaml 5.02+.

We therefore need to first migrate it to the compiler version before printing it so that it is properly annotated with attributes by the downward migration, allowing the upward one to preserve the original AST.

### DIFF
--- a/src/code_matcher.ml
+++ b/src/code_matcher.ml
@@ -114,6 +114,21 @@ struct
   let parse_string s =
     match M.parse (Lexing.from_string s) with [ x ] -> x | _ -> assert false
 
+  (* To round trip our AST we convert it to the compiler's version, print it as
+     source using the compiler pretty-printers, parse it back using the
+     compiler's parser and migrate it back to our version.
+
+     Skipping the first migration can lead to errors because some subtleties may
+     be lost by older parsers. For instance in OCaml 5.02 [fun x y -> z] and
+     [fun x -> fun y -> z] have different representation but in OCaml 5.01 they
+     both parse to the same AST. Running the migration to the compiler AST first
+     anotates the AST using attributes allowing the final migration to preserve
+     such differences. *)
+  let round_trip ast =
+    let compiler_ast = M.to_compiler ast in
+    remove_loc
+      (parse_string (Format.asprintf "%a@." M.pp_compiler compiler_ast))
+
   let rec match_loop ~end_pos ~mismatch_handler ~expected ~source =
     match (expected, source) with
     | [], [] -> ()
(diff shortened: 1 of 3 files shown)


NathanReb:
CC @patricoferris !

patricoferris:
Definitely seems to fix the diffing in the 5.2 AST branch (https://github.com/ocaml-ppx/ppxlib/pull/514/checks?check_run_id=29662622891 -- the errors now are because of something I need to potentially promote it seems). The `Astlib.Compiler_pprintast.structure_item` is unbound on some platforms it seems/compiler versions

NathanReb:
Yeah I was kind of expecting `Pprintast.structure_item` not to exist on older compilers, I'll fix that!

NathanReb:
> the errors now are because of something I need to potentially promote it seems
I got those errors when testing on top of the 5.2 bump as well. My guess is that the deriving_inline bits in `ast.ml` were not updated because of the roundtrip error but I might be misunderstanding what's going on here.

patricoferris:
> The other side are those weird formatting changes, such as removed spaces before = which I'm not quite sure where they are coming from.

I thought this too, but this is actually always been the case! I tested on older versions of the codebase -- the `=` and extra bracketing is _removed_ by a subsequent `dune build @fmt --auto` and any `dune build @lint` won't be retriggered afterwards :))

patricoferris:
Oh and the `fun x -> fun y ->` is _converted_ back to `fun x y ->` by OCamlformat!

patricoferris:
Ah that's the problem, sorry for all the messages, that's what is broken here. The traversal and ocamlformat are fighting each other, so a lint followed by a format undoes the work of the lint.

NathanReb:
We do still have a roundtripping issue though don't we? Can we go ahead and merge this PR?

src/code_matcher.ml
```diff
@@ -131,7 +133,9 @@ struct
         let y = remove_loc y in
         if Poly.( <> ) x y then (
           let round_trip =
-            remove_loc (parse_string (Format.asprintf "%a@." M.pp x))
+            let compiler_x = M.to_compiler x in
```
patricoferris, line 136: Could we leave a comment here for the future ppxlib devs about why exactly we do this -- something like your PR comment would be great?

patricoferris (approved):
I think we should leave some comment about the logic in this section, but otherwise this is good to go :))

src/code_matcher.ml
NathanReb: Sure thing!

PR #524, opened 2024-09-23
Simple pp output

patricoferris:
This is just an idea to share based on #517 -- changing the migrations `run.t` to use this was very useful for debugging _more_ issues with the 5.2 AST bump cc @NathanReb. 

We discussed (offline) the possibility of changing `-dparsetree` to output the `Pp_ast` output. We would need to understand the downstream use-cases of `-dparsetree` before doing that.

### DIFF
--- a/src/driver.ml
+++ b/src/driver.ml
@@ -913,6 +913,7 @@ let load_source_file fn =
 
 type output_mode =
   | Pretty_print
+  | Pp_simple
   | Dump_ast
   | Dparsetree
   | Reconcile of Reconcile.mode
@@ -1149,6 +1150,14 @@ let process_file (kind : Kind.t) fn ~input_name ~relocate ~output_mode
           Ast_io.write oc
             { input_name; input_version; ast }
             ~add_ppx_context:true)
+  | Pp_simple ->
+      with_output output ~binary:false ~f:(fun oc ->
+          let ppf = Stdlib.Format.formatter_of_out_channel oc in
+          let ast = add_cookies ast in
+          (match ast with
+          | Intf ast -> Pp_ast.signature ppf ast
+          | Impl ast -> Pp_ast.structure ppf ast);
+          Stdlib.Format.pp_print_newline ppf ())
   | Dparsetree ->
       with_output output ~binary:false ~f:(fun oc ->
           let ppf = Stdlib.Format.formatter_of_out_channel oc in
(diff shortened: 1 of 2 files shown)


patricoferris:
I think the "1,117" lines removed really speaks volumes for adding something like this to the driver. It saved me a lot of time debugging the migrations :))

test/501_migrations/one_migration.t
```diff
@@ -4,1127 +4,159 @@ test checks whether parsing on 5.0.0 (result of test running on 5.0.0) is the sa
 parsing on 5.1.0 and then migrating down to 5.0.0 (result of test running on 5.1.0).
 
 The test is mostly useful for debuggung problems in a full round-trip. Since Ppxlib's
-`dparsetree` option doesn't compactify or strip locations, its output is very long.
+`pp-simple` option doesn't compactify or strip locations, its output is very long.
```
NathanReb, line 7: That doesn't hold anymore as we do remove locations now!

test/501_migrations/one_migration.t
```diff
@@ -4,1127 +4,159 @@ test checks whether parsing on 5.0.0 (result of test running on 5.0.0) is the sa
...
-          ((pos_fname file.ml) (pos_lnum 2) (pos_bol 25) (pos_cnum 25)))
-         (loc_end
-          ((pos_fname file.ml) (pos_lnum 2) (pos_bol 25) (pos_cnum 40)))
-         (loc_ghost false))))))
-    (pstr_loc
-     ((loc_start ((pos_fname file.ml) (pos_lnum 2) (pos_bol 25) (pos_cnum 25)))
-      (loc_end ((pos_fname file.ml) (pos_lnum 2) (pos_bol 25) (pos_cnum 40)))
-      (loc_ghost false)))))
+  $ ./identity_driver.exe -pp-simple file.ml
+  [ Pstr_attribute
```
NathanReb, line 93: Ah that's good to know, I didn't think about floating attributes! I'll open a PR to fix that!

NathanReb (commented):
The output is incredibly easier to read and more compact, I wasn't expecting that much of an improvement over the sexp representation!

As you mentioned, I'll try to see if we can simply change the output of `-dparsetree` without having to introduce a new flag but if it turns out we can't this would be a worthy addition to the driver!

src/driver.ml
```diff
@@ -1343,6 +1353,14 @@ let standalone_args =
       Arg.Unit (fun () -> set_output_mode Dump_ast),
       " Dump the marshaled ast to the output file instead of pretty-printing it"
     );
+    ( "-pp-simple",
```
NathanReb, line 1356: Maybe `-pp-ast` would be clearer if we retain the new flag option.

patricoferris:
Closing in favour of #530

PR #525, opened 2024-09-24
Filter out floating attributes in Pp_ast

NathanReb:
This will prove particularly useful for our internal use as it won't print `[@@@ocaml.ppx.context]` unless asked to.


patricoferris (approved):
Thanks @NathanReb ! :)

PR #526, opened 2024-09-24
Add json mapper for pp_ast

pedrobslisboa:
## Description

Add `json` mapper on `pp_ast` to improve usage in tools such as AST Explorer.
The idea is to increment the @NathanReb PR https://github.com/ocaml-ppx/ppxlib/pull/517

@NathanReb, your opinion on how to structure it better is very welcome.
@jchavarri @davesnx thank you for the help.

## How

To use it, we just need to pass `--json` to `ppxlib-pp-ast.` It works with all other flags, such as `--show-attrs` and `--show-loc`.

## How to test

- In this branch build the project
- Create a test.ml on the root and add some content
- Run `ppxlib-pp-ast test.ml` and check the result
- Run `ppxlib-pp-ast --json test.ml` and check the result
- Use others flags to see the result with `--json`


NathanReb:
Yes, as Patrick said, ppxlib should not depend on external libraries as it otherwise prevents them from being able to depend on ppxlib.

The `Yojson` dependent parts should go into `ppxlib-pp-ast` or just a seperate tool shipped with `ppxlib-tools`.

We could add a json lifter to ppxlib but given it's fairly easy to write one this could be done directly in `ppxlib-tools`.

Did you try to load the resulting JSON into AST explorer? It would be nice to have a working usecase for this before we merge this feature!

pedrobslisboa:
@NathanReb Hey ya!

> Did you try to load the resulting JSON into AST explorer? It would be nice to have a working usecase for this before we merge this feature!

I've just pushed a PR on astexplorer refmt to add the ppxlib pp_ast working with this json feature: https://github.com/jchavarri/astexplorer-refmt/pull/1
This same feature could be used on ocaml-platform (Which I'm also working on to migrate to pp_ast). This print is a local demo of the extension running with pp_ast and json: 
Captura de Tela 2024-12-06 às 18 56 46


About the yojson, I'm working on removing it

NathanReb:
That's awesome! Let me know when this is ready for review!

pedrobslisboa:
@NathanReb BTW, I think now it's ready for review

NathanReb:
My impression is that we could simplify this quite a bit by having a lifter exposed that returns a Yojson compatible value.

It could be built on top of the `simple_val` lifter, inheriting all the config in the process, although I'd imagine that when converting an AST to JSON, we don't care as much about minimizing the output but rather to make it as close to the original AST and let consumer tool do the work.

That way, we can have `ppxlib-pp-ast` use said lifter from `ppxlib` and use Yojson to actually print it out as valid JSON.

How does that sound?

NathanReb:
Sorry for the delay by the way, the end of year holidays + the OCaml 5.03 release took quite a lot of my time!

src/pp_ast.mli
```diff
@@ -60,28 +60,46 @@ module Config : sig
           be. *)
 end
 
+type simple_val =
```
NathanReb, line 63: You're not responsible for this as I'm the one who came up with that name originally but since we're going to expose it, I think we should come up with something better.

I think some commonly used name for this kind of type is `repr` as in representation. If you have a suggestion I'd gladly hear it as I'm not super inspired here as the original name suggests :smile:

src/pp_ast.mli
```diff
@@ -60,28 +60,46 @@ module Config : sig
...
+  | Int32 of int32
+  | Int64 of int64
+  | Nativeint of nativeint
+  | Record of (string * simple_val) list
+  | Constr of string * simple_val list
+  | Tuple of simple_val list
+  | List of simple_val list
+  | Special of string
+
+type printer = Format.formatter -> simple_val -> unit
```
NathanReb, line 80: I guess we could remove this alias in favor of a simpler `repr pp` don't you think?

src/pp_ast.mli
```diff
@@ -60,28 +60,46 @@ module Config : sig
...
 type 'a pp = Format.formatter -> 'a -> unit
-type 'a configurable = ?config:Config.t -> 'a pp
+type 'a configurable = ?config:Config.t -> ?printer:printer -> 'a pp
```
NathanReb, line 82: For future reference: this could be considered a breaking change here.

Do you think it would make sense to make the `?printer` a part of the config directly? That way this functions API does not change and we just add a new optional argument to `Config.make`.

NathanReb: We could eventually split the full config into a lifter config and a printing config.

NathanReb (commented):
This looks good! I think you're right, that's the simplest and most flexible solution here.

Could you add a couple tests in `test/ppxlib-pp-ast/`, a changelog entry and update ppxlib-tools deps to reflect the changes?

pedrobslisboa:
> This looks good! I think you're right, that's the simplest and most flexible solution here.
> 
> Could you add a couple tests in `test/ppxlib-pp-ast/`, a changelog entry and update ppxlib-tools deps to reflect the changes?

Ops, I rebased it and forgot to add back the tests. Sorry

src/pp_ast.mli
pedrobslisboa: > For future reference: this could be considered a breaking change here.

I thought that adding this feature with backward compatibility would avoid that to be considered breaking change 😞 

Adding it to the Config like this couldn't also be considered a breaking change, but for the Config? 
image

src/pp_ast.mli
pedrobslisboa: `repr` looks like a good name, though.

src/pp_ast.mli
NathanReb: I think `Config.make` will pretty much always be applied fully as opposed to the printer which could be partially applied with the `?config` argument so it can then be used as a regular `'a pp`.

You are right that strictly speaking it is the same kind of change but given the nature of the two functions, I think it's far less likely to break user code that way.

What do you think?

pedrobslisboa: That makes sense to me

pedrobslisboa:
When you said, "update ppxlib-tools deps to reflect the changes?", what did you mean by that? I didn't get it.

Besides that, I think I updated the remaining content. LMK if anything is missing

NathanReb:
> When you said, "update ppxlib-tools deps to reflect the changes?", what did you mean by that? I didn't get it.

Now that ppxlib-tools depends on `yojson`, we need to declare this dependency in ppxlib-tools' opam file. We use dune to generate our opam file so what you need here is to update the relevant `(package ...)` stanza in `dune-project` to add `yojson` to the list of dependencies there.

The CI's green because we happen to also depend on it via the benchmark package but installing ppxlib-tools via opam would fail.

NathanReb:
I'm trying to wrap up the 5.3 compat thing and I'll review and merge right after, sorry for the delay!

pedrobslisboa:
Hey, @NathanReb, long time no see. Life happened to me as well, with some tasks to work on. But I'm back to review the PRs I left.

I update this one with the main.

pedrobslisboa:
> Hey, @NathanReb, long time no see. Life happened to me as well, with some tasks to work on. But I'm back to review the PRs I left.
> 
> I update this one with the main.

I just noticed that the cram tests were not being executed on my local project, I need to identify why, and then I'll fix the errors and push

NathanReb:
@pedrobslisboa nice, thanks! I'm trying to wrap up the 5.4 compat with @patricoferris but I'll review this shortly I promise!

src/pp_ast.mli
```diff
@@ -60,28 +111,27 @@ module Config : sig
...
-  type 'a printer
-
-  val structure : structure printer
-  val structure_item : structure_item printer
-  val signature : signature printer
-  val signature_item : signature_item printer
-  val expression : expression printer
-  val pattern : pattern printer
-  val core_type : core_type printer
+  type 'a ast_printer
```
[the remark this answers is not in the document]
pedrobslisboa: I just made the rebase and changed it.

NathanReb (approved):
Awesome, let's merge!

PR #527, opened 2024-09-24
Feat/pp ast pprintast

pedrobslisboa:
## Description

@NathanReb after read your feature and PR I thought about do this, what do you think?

Add the sprint function to the pp_ast. The idea is to provide an easy way to debug and improve dx on tests and documentation.

Before, the only way to get a string from it was by creating an AST, like `[%expr 40 + 2]` and applying `Pp_ast.expression (Format.formatter_of_buffer buffer) [%expr 40 + 2]` and then collect the value from the buffer, for a simple print it was necessary `Pp_ast.expression (Format.std_formatter) [%expr 40 + 2]`
Now, `pp_ast` has a sprint function that delivers the ast structure from a string. 

## Usage

```ocaml
let () = Pp_ast.sprint "42" |> print_endline
(* Pexp_constant (Pconst_integer ( \"42\", None)) *)
```

It was also added to Ppxlib.Pprintast

```ocaml
let () = Pprintast.sprint "42" |> print_endline
(* Pexp_constant (Pconst_integer ( \"42\", None)) *)
```

### DIFF
--- a/bin/pp_ast.ml
+++ b/bin/pp_ast.ml
@@ -1,23 +1,5 @@
 open Ppxlib
 
-module Kind = struct
-  type t = Signature | Structure | Expression | Pattern | Core_type
-
-  let to_utils_kind = function
-    | Structure -> Ppxlib_private.Utils.Kind.Impl
-    | Signature -> Ppxlib_private.Utils.Kind.Intf
-    | _ -> assert false
-end
-
-module Ast = struct
-  type t =
-    | Str of structure
-    | Sig of signature
-    | Exp of expression
-    | Pat of pattern
-    | Typ of core_type
-end
-
 module Input = struct
   type t = Stdin | File of string | Source of string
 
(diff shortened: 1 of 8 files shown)


NathanReb:
Do you have a concrete usecase in mind for this? One where `ppxlib-pp-ast` couldn't be used and this would have to be done programmatically.

pedrobslisboa:
> Do you have a concrete usecase in mind for this? One where ppxlib-pp-ast couldn't be used and this would have to be done programmatically.

@NathanReb, I think this is just for debugging and DX usage. I use Pprintast.string_of_expression a lot, and I'm missing something like it, but for printing the last.

I think `ppxlib-pp-ast` covers all cases for CLI usage, but it's not there for code usage, right?
The `Pp_ast.expression` could work for it and for my cases. But it's strange to provide like this:
```ocaml
let ast_string expr = 
  let buffer = Buffer.create 256 in
  let formatter = Format.formatter_of_buffer buffer in
  let _ = Pp_ast.expression formatter expr
  Format.pp_print_flush formatted ();
  Buffer.contents buffer
```
But I need to work with an Ast instead of a string.

Looking at your comment, I see I'm focusing on my use case, which I can solve locally. I don't know if it could be a feature anyone else would like.

NathanReb:
If you'd just like to get a string you can use `Format.asprintf`:
```ocaml
let ast_string exp =
  let config = Pp_ast.Config.make () in
  Format.asprintf "%a" (Pp_ast.expression ~config) exp
```

I realize that the config optional argument makes it a bit annoying here as otherwise you could just write:
```ocaml
Format.asprintf "%a" Pp_ast.expression exp
```

I haven't really thought the API through. I'll come up with something so that the printer types are compatible with `%a` format strings.

Would that be good enough for you?

NathanReb:
I'm curious to know in which cases you'd need the parsing though. Could you point to the project where you'd like to use the 'parse source code' + 'turn the AST into a string' function?

PR #530, opened 2024-09-26
Replace -dparsetree sexp output with a pretty-printed AST

NathanReb:
This is an attempt to replace `-dparsetree` output format to something more readable.

I'll trigger a rev-deps build on opam to see if that would break anything.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -21,6 +21,10 @@ details.
   a `ppxlib-pp-ast` executable in a new separate `ppxlib-tools` package
   (#517, @NathanReb)
 
+- Change `-dparsetree` from a sexp output to a pretty printed AST, closer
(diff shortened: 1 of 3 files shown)


NathanReb:
See https://github.com/ocaml/opam-repository/pull/26623 for rev-deps.

NathanReb:
I couldn't find anything in the revdeps build indicating that anyone's relying on the sexp output but you know what it's like to look through opam-ci's ppxlib revdeps builds. If you'd like to have a look @patricoferris please let me know!

patricoferris (approved):
Looked through the CI as best I could and looks good to me. Also did a little poking on Github search and sherlocode. I think we make the change and patch downstream breakages which I really can't imagine would be too many :)

PR #536, opened 2024-11-06
Fix out of date test

NathanReb:
The test wasn't properly updated following #525 and #530, not sure how it went unoticed.


NathanReb:
Seems like our tests aren't 5.3 proof :/. If that sounds good to you, I'll take care of that in a separate PR.

Issue #338, opened 2022-04-30
"Corrected code doesn't round-trip" with deriving_inline


sim642:
I have defined a deriver named "leq" with ppxlib, but I guess any ppxlib-based deriver would have the same issue.
First, I use it in a test file as follows:
```ocaml
type t = L1.t [@@deriving_inline leq][@@@end]
```
Second, running dune's build and promote changes it to the following:
```ocaml
type t = L1.t [@@deriving_inline leq]
let _ = fun (_ : t) -> ()
let rec (leq : t -> t -> bool) =
  let __0 () = L1.leq in
  ((let open! ((Ppx_deriving_runtime)[@ocaml.warning "-A"]) in __0 ())
    [@ocaml.warning "-A"])[@@ocaml.warning "-39"]
let _ = leq
[@@@end]
```
This promoted file now refuses to build by giving the following error:
```
Error: ppxlib: the corrected code doesn't round-trip.
This is probably a bug in the OCaml printer:
<no differences produced by diff>
diff: /tmp/build_f87c44_dune/ppxlibb4e945: No such file or directory
diff: /tmp/build_f87c44_dune/ppxlib38b9eb: No such file or directory
```

Third, I discovered that removing the `let _` definitions (whatever they are there for) and manually changed the file to the following:
```ocaml
type t = L1.t [@@deriving_inline leq]
let rec (leq : t -> t -> bool) =
  let __0 () = L1.leq in
  ((let open! ((Ppx_deriving_runtime)[@ocaml.warning "-A"]) in __0 ())
    [@ocaml.warning "-A"])[@@ocaml.warning "-39"]
[@@@end]
```
Then the round-tripping error disappears, but dune build again suggests a promotion to add the two definitions back, leading back to the second code snippet above.

Therefore I cannot get deriving_inline to a stable state, where it builds without round-trip errors and doesn't keep suggesting additional promotions.

panglesd:
Thanks for the report!

Time allocated to ppxlib has become a scarce ressource, however I'll do my best to investigate this issue. Sorry in advance for the dealy!

pitag-ha:
Hey, thanks for reporting and sorry for the delay! I've just tried to reproduce the problem and am not able to reproduce it. Are you using a mix between `ppx_deriving` and `ppxlib` to define your `leq` deriver? I'm wondering because of the open of the `ppx_deriving` runtime (which might be related to the round-trip problem, but I'm not sure).

sim642:
[...]

PR #537, opened 2024-11-06
Add Submodules with simpler APIs to Pp_ast

NathanReb:
This adds a functor and a `Default` module, similar to what's done in `Ast_builder`. One can use the functor to get pre-configured formatters or use the formatters provided by the `Default` module which are preconfigured with the default configuration.

This provides modules whose pretty printers can be used directly with `"%a"` format strings as they don't expect an optional argument first.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,7 +19,7 @@ details.
 
 - Add ppxlib's AST pretty-printing utilities in `Ppxlib.Pp_ast` and
   a `ppxlib-pp-ast` executable in a new separate `ppxlib-tools` package
-  (#517, @NathanReb)
+  (#517, #525, #537, @NathanReb)
 
 - Change `-dparsetree` from a sexp output to a pretty printed AST, closer
   to what the compiler's `-dparsetree` is.
--- a/src/pp_ast.ml
+++ b/src/pp_ast.ml
@@ -302,21 +302,32 @@ class lift_simple_val =
       | NoInjectivity -> Constr ("NoInjectivity", [])
   end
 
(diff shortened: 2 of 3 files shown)


patricoferris (approved):
This looks fine to me.

Is there any benefit to doing this the `Ast_builder` way fully by extracting the type interface and using a `'a with_loc` and `'a without_location` type (except for with and without a `config`) ?

NathanReb:
Yes that sounds like a good idea! I'll factor out the pretty-printer type!

I'll go ahead and make the test suite 5.3 compatible first though!

NathanReb:
I pushed the changes with the new alias types. I didn't go for `with/without_config` as I found those names a bit ambiguous. One could read `with_config` either as with a config argument or with an embedded configuration. I used `configured` and `configurable` instead as I found those would leave less room for interpretation.

NathanReb:
This is due to the quoter returning different IDs in one case and the other. The lint build only generates the deriving_inline code and therefore gets different IDs than the regular builds that also run the rest of the PPX-es.

I need to refresh my memory on how `deriving_inline` is supposed to work but it's a little odd that it's ran buy the regular builds AND the lint. I would expect it to only be ran by the lint builds so that it's not a dependency of the regular builds anymore.

mbarbin:
@NathanReb did you mean to reply to #338 ? (your comment makes more sense to me in the context of this other issue)

patricoferris (approved):

NathanReb:
> @NathanReb did you mean to reply to #338 ? (your comment makes more sense to me in the context of this other issue)

Indeed! My bad, I'll move the message!

Issue #509, opened 2024-07-10
Question on writing deriver for class


Wenke-D:
It appears impossible to get the attache class AST using **context free transformation**.

```OCaml
class apple = object
   ...
end [@@inspect]
```
in the ppxlib, `Deriving.add` do not have the optional arg to get `class_declaration list`.

Did I miss something or it is indeed impossible.

Thanks !

Wenke-D:
The objective is to inspect the methods and fields the class, then generate some functions outside.

NathanReb:
Yes indeed it seems it's not supported. Off the top of my head I can't think of a reason why we wouldn't add support for this!

PR #538, opened 2024-11-09
Support class type declarations in derivers

patricoferris:
This PR adds support for class _type_ declarations, which I mistook #509 for. All the same, seems like a good feature to add given ppxlib didn't support it.

It reuses some code from type declarations as they are pretty similar but there is Recursive flag for class types (are the just always recurive?). This bit could perhaps use some more work.

### DIFF
--- a/src/deriving.mli
+++ b/src/deriving.mli
@@ -105,10 +105,12 @@ with type deriver := t
 
 val add :
   ?str_type_decl:(structure, rec_flag * type_declaration list) Generator.t ->
+  ?str_class_type_decl:(structure, class_type_declaration list) Generator.t ->
   ?str_type_ext:(structure, type_extension) Generator.t ->
   ?str_exception:(structure, type_exception) Generator.t ->
   ?str_module_type_decl:(structure, module_type_declaration) Generator.t ->
   ?sig_type_decl:(signature, rec_flag * type_declaration list) Generator.t ->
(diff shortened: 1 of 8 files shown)


src/deriving.mli
```diff
@@ -105,10 +105,12 @@ with type deriver := t
 
 val add :
   ?str_type_decl:(structure, rec_flag * type_declaration list) Generator.t ->
+  ?str_class_type_decl:(structure, class_type_declaration list) Generator.t ->
```
NathanReb, line 108: This change could potentially break some rev deps, although I deem it quite unlikely.

I think it's fair to treat this as non breaking from our perspective given how the `add` and `add_alias` function are used in practice but still wanted for us to keep this in mind before we release this!

NathanReb (commented):
This looks good to me! I can't think of a good reason why not to add this.

I'm curious as to why you've added the `Class_decl` context and handled it everywhere without exposing it. Is this an omission or is there a particular reason why you did this that I missed?

I think only allowing it for class types is fair given that's what we do for modules as well, although I could see how it would be useful to allow deriving values from class/module declarations but I think it's a topic for another time.

src/deriving.mli
patricoferris: Agreed. 

As a litmus test, I tested this change against 50 revdeps without any failures due to this. There were failures because of flakey tests and no solves.

Issue #532, opened 2024-10-11
`Location.none` is different than the compiler's


ncik-roberts:
Both the compiler and ppxlib define `Location.none` as `let pos = { dummy_pos with pos_fname = "_none_" } in { loc_start = pos; loc_end = pos; loc_ghost = true }`.

However, ppxlib defines `dummy_pos` as `{ pos_fname = ...; pos_lnum = 1; pos_bol = 0; pos_cnum = -1 }` [^1], whereas the compiler defines `dummy_pos` as `{ pos_fname = ...; pos_lnum = 0; pos_bol = 0; pos_cnum = -1 }` [^2]. `pos_lnum` is different.

This mismatch causes at least 1 bug in ppxlib. For example, this code that attempts to show a reasonable location in the event that an attribute's location is `none` will fail to detect the case where the _compiler_ inserted the `none`-locationed attribute (e.g. a doc comment):

https://github.com/ocaml-ppx/ppxlib/blob/8a0cb7122d7d454c20d732621795d910018d1b66/src/common.ml#L146-L157

At Jane Street, we've patched ppxlib's `Location.none` to be the same as the compiler's with no issues.

[^1]: https://github.com/ocaml-ppx/ppxlib/blob/8a0cb7122d7d454c20d732621795d910018d1b66/src/location.ml#L10-L19
[^2]: https://github.com/ocaml/ocaml/blob/b8f23dd5af09a3d9e6f891e2ac7f0d4baee81e1b/stdlib/lexing.ml#L25-L30

patricoferris:
Thanks @ncik-roberts ! 

I'll run this through some reverse dependency checks to see if anything comes up but I think we could align with the compiler's values.

patricoferris:
Hey @ncik-roberts,

Whilst writing some tests for this portion of code I was wondering under which circumstances you ran into this issue -- it seems to be fairly hard to trigger given most of the checks against `Location.none` are happening behind the `-checks` flag? The standard `ocaml.doc` attributes are then ignored as they are reserved? Maybe some JS libraries are manipulating doc comments? Having a concrete example would be very useful if possible?

ncik-roberts:
Yes, I think it's simply that a user ppx was calling the `loc_of_attribute` mentioned in my original post and raising an error (with `Location.raise_errorf` or equivalent) with the returned location. I think you can repro it on the following OCaml program:

```
(** my doc comment *)
let x = 3
```

[...]

PR #540, opened 2024-11-25
Match compilers dummy loc

patricoferris:
A fix as per #532.

I tested reverse dependencies (I think most of them, for their latest version) and they mostly installed `--with-test`. Any that didn't seemed unrelated.

### DIFF
--- a/src/location.ml
+++ b/src/location.ml
@@ -8,7 +8,7 @@ type t = location = {
 }
 
 let in_file name =
-  let loc = { pos_fname = name; pos_lnum = 1; pos_bol = 0; pos_cnum = -1 } in
+  let loc = { pos_fname = name; pos_lnum = 0; pos_bol = 0; pos_cnum = -1 } in
   { loc_start = loc; loc_end = loc; loc_ghost = true }
 
 let set_filename loc fn =
@@ -17,6 +17,7 @@ let set_filename loc fn =
   { loc with loc_start; loc_end }
 
 let none = in_file "_none_"
+let is_none v = Poly.( = ) v none
 
 let init lexbuf fname =
   let open Lexing in
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 unreleased
 ----------
  
+- Change `Location.none` to match the compiler's `Location.none` as of OCaml
+  4.08. This fixes a bug in `loc_of_attribute` (#540, @ncik-roberts, @patricoferris)
+
 - Bump ppxlib's AST to 5.2.0 (#514, @patricoferris)
 
 - Add a `-raise-embedded-errors` flag to the driver. Setting this flag raises the first
--- a/src/common.ml
+++ b/src/common.ml
@@ -149,7 +149,7 @@ let loc_of_attribute { attr_name; attr_payload; attr_loc = _ } =
      from older asts. *)
   (* "ocaml.doc" attributes are generated with [Location.none], which is not helpful for
      error messages. *)
-  if Poly.( = ) attr_name.loc Location.none then
+  if Location.is_none attr_name.loc then
     loc_of_name_and_payload attr_name attr_payload
   else
     {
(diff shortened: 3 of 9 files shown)


NathanReb (approved):
Looks good to me!

This probably needs a changelog entry!

NathanReb:
Also, rebasing it will reduce the diff as the quoter tests don't print out the whole AST anymore!

patricoferris:
Turns out this actually changed in OCaml compiler versions, so this is not the proper fix.

- `<= 4.07` it matches what we have, see for example https://github.com/ocaml/ocaml/blob/4.07/parsing/location.ml
- `>= 4.08` it switched to used `Lexing.dummy_pos`

The fix might be fairly easy (using Cinaps) but the problem becomes all of the expect tests. And whilst we could in theory do the `[%%expect_in]` stuff I think our tests become less and less useful the longer they get and they're filled with `none` locations. It would be great to run `Toploop.execute_phrase` and then parse values and use `Pp_ast` on them... Any ideas @NathanReb ?

NathanReb:
I guess we could use [%%ignore] here, similar to what I did for the `Quoter` tests:
```ocaml
let x = [%expr ...]
[%%ignore]

let () = Pp_ast.Default.expression x
[%%expect {|
...
|}]
```

Would that work for you?

patricoferris:
It actually turns out I don't need that @NathanReb -- the files are only enabled for new enough compilers that it doesn't matter. But it is good to have that trick up my sleeve should we need it. Happy to make that change here to help future proof these tests (but I don't think it is likely for those bits to change too much).

astlib/location.mli
```diff
@@ -52,6 +52,10 @@ val set_input_lexbuf : Lexing.lexbuf option -> unit
 val none : t
 (** An arbitrary value of type [t]; describes an empty ghost range. *)
 
+val dummy_pos : Lexing.position
```
NathanReb, line 55: I'm not sure we should expose this as Astlib is supposed to represent the minimum stable API we'd need from the compiler and it seems that something like `in_file` would be enough here wouldn't it?

src/location.ml
```diff
@@ -8,7 +8,7 @@ type t = location = {
 }
 
 let in_file name =
-  let loc = { pos_fname = name; pos_lnum = 1; pos_bol = 0; pos_cnum = -1 } in
+  let loc = { L.dummy_pos with pos_fname = name } in
```
NathanReb, line 11: I think this is equivalent to the compiler's `in_file`. If we used it instead, we wouldn't need the cinaps bits for dummy_pos anymore.

NathanReb (commented):
If I understood the issue correctly, the important part is to agree with the compiler on what is `Location.none` right?

If so then we could probably simply stick to using the compiler's functions to build `none` locs.

This won't work when reading AST emitted by a different compiler (which we in theory support) but even that could be fixed by versioning the none loc.

As I'm writing this I wonder if the right fix isn't the following:
1. Stick to having our stable `Location.none` and choose a convention for it
2. Have a function that detects whether a location is a valid `none` loc from any version we support
3. Use the function from `2` instead of poly `=` in our codebase

What do you think?

patricoferris:
The current errors we're seeing on 4.04 and 4.05 are due to newer compiler versions checking of a line number is "valid" or not (i.e. `> 0`) and replacing it with `1` if not... https://github.com/ocaml/ocaml/blob/7a9d83047450e337ff31f3d73e5c2d9af66bd363/parsing/location.ml#L214

On the older compilers, the location is used directly, hence we are getting `0` with this new change. Possible solutions are to:

 - manually change the line number location
 - revert back to making `Location.none` match the compiler's `Location.none`

any thoughts @NathanReb ?

NathanReb:
Given these are errors we produce ourselves and they point to an actual file, I guess we could directly have those errors point to line 1. Maybe we can factor this out in an internal `Location.raise_error_in_filef: file: string -> ...` kind of function.

How does that sound?

NathanReb:
That being said, I'm particularly against using the matching compiler's `Location.none` if you think that's the better approach!

patricoferris:
@NathanReb -- as we discussed, since ppxlib only supports compilers 4.08+ we no longer need to handle multiple possible values for `Location.none` which greatly simplifies this PR. We're making a bit of a bet on it not changing in the future but I think that's ok.

NathanReb (approved):
Perfect, thanks!

I also think it's acceptable to bet on `Location.none` not changing again. If it does we can experiment with the other suggestions, e.g. migrating `Location.none` between versions or have `Location.is_none` accept a range of representations.

PR #544, opened 2024-12-09
[WIP] Add ppx-by-example

pedrobslisboa:
> [!WARNING]
> This is a work-in-progress branch
>Be careful on reviewing it as it's not ready for it

## Description

This is the first PR of the movement of adding the examples of ppx-by-example to ppxlib

## Todo

- [ ] Add test section
- [ ] Separe this branch into smaller ones

PR #545, opened 2024-12-15
Add AST section to examples

pedrobslisboa:
## Description

This is the first part of adding the ppx-by-example to ppxlib
I will divide it into PR so we can review it better, focusing on the explanation and typos without having tons of files to review. It is really necessary that this section is easy to understand and is aligned with ppxlib.
@NathanReb Can you create a `ppx-by-example` base branch? So I can point this one to it, and after all merges (Other sections) on this branch, we move with the entire `ppx-by-example` branch to main?

## AST

This PR adds the AST sections AST / Building ASTs / Destructing ASTs

## How to test it

- At this branch run 
  ```bash
   make doc
   open _build/default/_doc/_html/ppxlib/index.html
  ```
- Access [Examples](file:///Users/pedrolisboa/study/ppxlib/_build/default/_doc/_html/ppxlib/examples.html)
- Access each section: 
  - AST
  - Building an AST
  - Destructing an AST


Makefile
pedrobslisboa, line 1: @NathanReb I've made some improvements on Makefile, can you check out if it makes sense?

The main feature is to add `doc-dev`, but I also added the `help` command

pedrobslisboa:
@patricoferris Tagging you as you reacted to this [PR](https://github.com/ocaml-ppx/ppxlib/pull/544)

doc/dune
```diff
@@ -1,2 +1,11 @@
...
  (package ppxlib))
+
+(rule
+ (alias doc)
+ (deps
+   (glob_files ./images/*))
+ (action
+   (progn
+     (system "mkdir -p %{project_root}/_doc/_html/ppxlib/assets/images")
+     (system "cp -R ./images/ %{project_root}/_doc/_html/ppxlib/assets/images/"))))
```
pedrobslisboa, line 11: Btw, I didn't know any other way to get assets on odoc. Do you guys happen to know?

NathanReb:
@pedrobslisboa just pushed a `ppx-by-example` branch, sorry for the delay!

doc/dune
pedrobslisboa: @NathanReb @patricoferris I know that Odoc v3 supports it; maybe there is no alternative other than due to a hack similar to this one. Do you guys know a better hack?

panglesd: Indeed, this hack will unfortunately not work for the documentation [generated for ocaml.org](https://ocaml.org/p/ppxlib/latest/doc/index.html).

The proper way to do it is to:
- install the image files in the `odoc-pages` folder of the doc (see [this example](https://github.com/ocaml/odoc/blob/master/doc/dune#L55) for how to do it with dune)
- Include it with [the odoc3 syntax](https://ocaml.github.io/odoc/odoc/odoc_for_authors.html#media), in your case `{image!filename.png}`

The tooling support for for odoc 3 is still under construction, so `dune` won't yet be able to build the documentation with the images... however, [`odoc_driver` can](https://ocaml.github.io/odoc/odoc-driver/index.html) and soon dune and ocaml.org will be able to do it! You would be one of the first users of images the odoc 3 way!

(Another way to do it is to use permalinks to the image eg `{image:https://github.com/ocaml-ppx/ppxlib/<hash>/filename.png}`)

I also wonder if turning some images into text wouldn't be better for accessibility.

pedrobslisboa:
> I just updated the ppx-by-example branch and rebased it on top of `main`.
> 
> Can you please rebase this @pedrobslisboa ? I'll review straight away!

Rebased and ready for review.

pedrobslisboa:
🏓

doc/example-ast-building.mld
```diff
@@ -0,0 +1,173 @@
...
+Beyond simple expressions, you may need to build more complex ASTs, such as [let] expressions.
+
+{2:example-1-constructing-a-let-expression-with-ast_builder Example 1: Constructing a Let Expression with {!Ppxlib.Ast_builder}}
+
+{{:https://github.com/ocaml-ppx/ppxlib/tree/main/examples/1-AST/a%20-%20Building%20AST/building_ast.ml#L40-L60} 🔗 Sample Code}
+
+{[
+let let_expression =
+  let expression =
+    Ast_builder.Default.pexp_constant ~loc:Location.none
```
patricoferris, line 134: If we can, I think we want to try and suggest best practices and not have write `~loc:Location.none`, how about this becomes a `let let_expression ~loc =` or we open a module where `~loc` is defined which is usually what we suggest doing?

doc/example-ast-destructing.mld
```diff
@@ -0,0 +1,171 @@
...
+You’ll need to destructure the AST representing the extension point ([%one]) to replace it with [1].
+There are several ways to destructure an AST. We’ll explore three methods:
+
+- {b AST Structure Pattern Matching}
+- {b Using {!Ppxlib.Ast_pattern} High-Level Destructors}
+- {b Using Metaquot}
+
+{1:ast-structure-pattern-matching AST Structure Pattern Matching}
+
+The most fundamental method for destructuring an AST in PPXLib is by directly matching on the AST’s structure.
```
patricoferris, line 44: Looking through the current documentation and README, it seems to me that we tend to write Ppxlib rather than PPXLib. Perhaps we could even change all occurrences of PPXLib into `{! Ppxlib}`.

doc/example-ast-destructing.mld
```diff
@@ -0,0 +1,171 @@
...
+
+You can further simplify it:
+
+{[
+let match_int_payload =
+  let open Ast_pattern in
+  pstr (pstr_eval (eint (int 1)) nil ^:: nil)
+]}
+
+Using [eint] instead of [pexp_constant] and [pconst_integer] provides better type safety. The [int] wildcard captures the integer value.
```
patricoferris, line 122:
```suggestion
Using [eint] instead of [pexp_constant] and [pconst_integer] provides better type safety. The [int] combinator captures the integer value.
```

I'm not entirely sure about the claim of "better type safety", I think `eint` is just a short-hand for the same code: https://github.com/ocaml-ppx/ppxlib/blob/b3c240e020aba81b7da26ac3c00c2df6710dfda7/src/ast_pattern.ml#L226 ?

doc/example-ast-destructing.mld
```diff
@@ -0,0 +1,171 @@
...
+- {b [Pconst_integer]} becomes [pconst_integer]
+
+{2:example-2-simplifying-matching-with-eint Example 2: Simplifying Matching with [eint]}
+
+{{:https://github.com/ocaml-ppx/ppxlib/tree/main/examples/1-AST/b%20-%20Destructing%20AST/destructuring_ast.ml#L40-L49} 🔗 Sample Code}
+
+You can further simplify it:
+
+{[
+let match_int_payload =
```
patricoferris, line 117: Can we further simplify this to:

```ocaml
let match_int_payload =
  let open Ast_pattern in
  single_expr_payload (eint (int 1))
```

PR #548, opened 2024-12-19
Fix packaging

NathanReb:
I ran into some errors while testing out the repo before attempting the release, this should fix those:
1. Correctly allocate tests to their respective packages. Our main test suite now is affected to `ppxlib` so they aren't run when running something like `dune build -p ppxlib-tools` or `opam install -t ppxlib-tools`. This prevent some test dependencies to leak from `ppxlib` to `ppxlib-tools`.
2. Bump our dune lang version. Initially opam CI reported that we had to update dune lang to 2.9 but while solving `1`, I actually ended up having to bump it to 3.8. This enabled some warnings that I guess where not part of the default build profile before and I ended up fixing those as well.


NathanReb:
It seems we can't currently allocate cinaps stanzas to specific packages so I ended up also adding a test deps on cinaps in ppxlib-tools for now. I reported this upstream: https://github.com/ocaml/dune/issues/11232.

If there's a better, easier fix, I'll of course patch this before the release!

NathanReb:
Not sure I understand what's wrong with the bench build, are you any more familiar with it than I am @patricoferris ?

NathanReb:
Ok the question rather was how did it even work before but I fixed it. It needed an explicitly declared dependency on the `drivers/` folder. I guess bumping the dune lang changed dune's behaviour and it stopped copying those files to `_build` directory without any explicit dependency over them.

patricoferris (approved):
Thanks @NathanReb -- sorry for only getting around to this now, post holidays. This seems like a good clean-up to me

NathanReb:
No worries @patricoferris ! I was on holidays myself as well!

PR #552, opened 2025-01-13
5.3 effect syntax

hhugo:


NathanReb:
I like the idea, if that sounds good to you I'll make it more aligned with our other migration specific extensions and add tests to turn it into a full PR.

hhugo:
> I like the idea, if that sounds good to you I'll make it more aligned with our other migration specific extensions and add tests to turn it into a full PR.

Feel free to take it from there.

NathanReb:
I added a test and tweaked a couple simple things.

I still have a couple concerns though before we can merge this:
1. it won't work if someone's using the driver's source output as it just straight up prints the internal AST version without running the migration back to the compiler's version
2. I wonder how it interacts with `deriving_inline` and other correction generating ppx-es. My feeling is that it should be fine as long as one doesn't try to directly get the deriver to embed code that uses the effect syntax but it will likely result in the corrected output containing the extension otherwise.

I'll test this a bit more. If it turns out I was right, I think we could add a flag to the driver to force it to migrate the AST and use the compiler's printer to prevent that from happening in the few places where it could cause troubles.

PR #554, opened 2025-01-13
Update ocamlformat 0.27.0

NathanReb:
CC @Julow!


Julow (approved):
The diffs look like what we expect.

PR #555, opened 2025-01-15
Add --use-compiler-pp flag to standalone driver

NathanReb:
This flag allows user to force the driver to print AST as source code using the installed compiler printers rather than its own.

This can be useful in situation where one is using the driver source output on older OCaml version to prevent outputting incompatible syntax. In addition it can prevent the driver to output uninterpreted extensions when migrating down newer compiler features before we bumped the internal AST.

### DIFF
--- a/src/reconcile.ml
+++ b/src/reconcile.ml
@@ -291,7 +291,20 @@ let reconcile ?styler (repls : Replacements.t) ~kind ~contents ~input_filename
               copy_input pos ~up_to:repl.start.pos_cnum ~line ~last_is_text
                 ~is_text
             in
-            let s = Replacement.text ~use_compiler_pprint repl in
+            let s =
+              try Replacement.text ~use_compiler_pprint repl
+              with Astlib.Compiler_pprintast.Unavailable ->
+                let loc =
+                  {
+                    (Location.in_file input_filename) with
+                    loc_start = repl.start;
+                    loc_end = repl.stop;
+                  }
+                in
+                Location.raise_errorf ~loc
+                  "Ppxlib.Reconcile: Cannot print this AST fragment using the \
+                   compiler printers with OCaml < 4.14"
+            in
             let line =
               match target with
               | Output Using_line_directives ->
--- a/src/reconcile.ml
+++ b/src/reconcile.ml
@@ -28,6 +28,64 @@ module Context = struct
     | Floating_attribute Signature_item -> Pprintast.signature_item
     | Floating_attribute Class_field -> Pprintast.class_field
     | Floating_attribute Class_type_field -> Pprintast.class_type_field
+
+  let compiler_printer : type a. a t -> Stdlib.Format.formatter -> a -> unit =
+   fun ctx ppf a ->
+    let open Extension.Context in
+    let open Attribute.Floating.Context in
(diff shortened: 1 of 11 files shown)


NathanReb:
I'm struggling a little bit to write a test for this. Ideally we'd like a situation where the tests will work even after we bump our AST.

That means that we'd like to try and generate new syntax nodes that either can't be migrated down or that are printed out as different source code once migrated down. I initially thought I could work something out with the unit functor parameter but it turns out it's just a representation change, the syntax was already supported before it seems.

Ideally we'd want something that we can test with 4.08 minimum since we're going to drop support for older compilers.

If anyone as a suggestion I'd gladly take it!

NathanReb:
Ok, this should be ready for review. I found a decent way to test it, even though I'm no completely satisfied but I think we'll have to settle for this for now.

I used the named existential syntax which only exist from 4.13 onward:
```ocaml
match x with
| C (type a) ... -> ...
```

The test runs on 4.12 and lower and shows that by default we successfully output the new syntax because our internal printers and AST are 4.14 at the time of writing. If we use the flag, it tries to migrate this before printing and fails as we expect. I would have prefered a test with a successful outcome but I couldn't figure out a nice printing change for that. Maybe I missed something there so feel free to make any suggestion!

Another thing to note is that the full range of Pprintast functions is only exposed since 4.14 so to make this compatible with older compilers I exposed dummy versions of those functions on old compilers. This is very unlikely to be hit by the vast majority of users as the only commonly used correction based code gen is `deriving_inline` which only requires structure and signature printers. I'll make it so the error clearly points out that the problem is the combination of OCaml version + `--use-compiler-pp`.

NathanReb:
I think this should be ready for review @patricoferris, if you have some spare cycles!

test/driver/compiler-pp/run.t
```diff
@@ -0,0 +1,65 @@
+The --use-compiler-pp flag can be used when using the driver's source code
+output, either directly are when generating a .corrected file, to force
```
patricoferris, line 2:
```suggestion
output, either directly when generating a .corrected file or to force
```

src/reconcile.ml
```diff
@@ -230,7 +291,20 @@ let reconcile ?styler (repls : Replacements.t) ~kind ~contents ~input_filename
...
+                let loc =
+                  {
+                    (Location.in_file input_filename) with
+                    loc_start = repl.start;
+                    loc_end = repl.stop;
+                  }
+                in
+                Location.raise_errorf ~loc
+                  "Ppxlib.Reconcile: Cannot print this AST fragment using the \
+                   compiler printers with OCaml < 4.14"
```
patricoferris, line 306: Is this 4.14 going to have to be updated to 5.2 with the AST bump? If so, is there anyway for us to make this easier to remember?

patricoferris (approved):
Thanks Nathan -- this is pretty useful for debugging things too.

src/reconcile.ml
NathanReb: No, it's 4.14 as in the installed compiler version has to be at least 4.14 for this to work so this won't change for as long as we support OCaml < 4.14!

PR #556, opened 2025-01-22
Drop support for OCaml < 4.08

NathanReb:


NathanReb:
I'm a bit surprised as when I run the tests locally I get the following diff:
```
$ dune runtest
File "test/driver/exception_handling/run.t", line 1, characters 0-0:
diff --git a/_build/.sandbox/7bf643a7b13fe8b4ac90a5ba1f2309f0/default/test/driver/exception_handling/run.t b/_build/.sandbox/7bf643a7b13fe8b4ac90a5ba1f2309f0/default/test/driver/exception_handling/run.t.corrected
index 3fc684ec..cc421bae 100644
--- a/_build/.sandbox/7bf643a7b13fe8b4ac90a5ba1f2309f0/default/test/driver/exception_handling/run.t
+++ b/_build/.sandbox/7bf643a7b13fe8b4ac90a5ba1f2309f0/default/test/driver/exception_handling/run.t.corrected
@@ -143,10 +143,10 @@ when the -embed-errors flag is not passed
 
   $ echo "let _ = [%gen_raise_exc] + [%gen_raise_exc]" > impl.ml
   $ ./extender.exe impl.ml
-  Fatal error: exception Failure("A raised exception")
+  Fatal error: exception (Failure "A raised exception")
   [2]
   $ ./extender.exe -embed-errors impl.ml
-  Fatal error: exception Failure("A raised exception")
+  Fatal error: exception (Failure "A raised exception")
   [2]
 
  In the case of derivers
@@ -155,7 +155,7 @@ when the -embed-errors flag is not passed
   $ echo "type b = int [@@deriving deriver_raised_exception]" >> impl.ml
   $ echo "type b = int [@@deriving deriver_raised_exception2]" >> impl.ml
   $ ./deriver.exe -embed-errors impl.ml
-  Fatal error: exception Failure("A raised exception")
+  Fatal error: exception (Failure "A raised exception")
   [2]
 
  In the case of Constant types
@@ -203,10 +203,10 @@ when the -embed-errors flag is not passed
 
   $ echo "let _ = [%gen_raise_exc] + [%gen_raise_exc]" > impl.ml
   $ ./whole_file_exception.exe impl.ml
-  Fatal error: exception Failure("An exception in a whole file transform")
+  Fatal error: exception (Failure "An exception in a whole file transform")
   [2]
   $ ./whole_file_exception.exe -embed-errors impl.ml
-  Fatal error: exception Failure("An exception in a whole file transform")
+  Fatal error: exception (Failure "An exception in a whole file transform")
   [2]
```

@patricoferris could you try to build this branch locally on a 4.08 switch to see if you get the same resulsts?

patricoferris:
I got the same problem,  but I had to jump through a lot of hoops first. It would seem dependencies we have for our benchmarks are giving us grief here: https://ocaml.ci.dev/github/ocaml-ppx/ppxlib/commit/ccfc1007f7f6bbc3482621b103bb5c990bd502f9/variant/debian-12-4.08_opam-2.3#L245-245

After I installed `base` on `4.08.1` I managed to reproduce your issues @NathanReb. But now, after I remove `base` it is still failing like we see here... I don't know why the CI is not failing...

patricoferris:
To recreate the CI _success_:

```sh
$ opam sw create 4.08.1
$ opam pin . -yn --with-version=dev
$ opam install ppxlib ppxlib-tools
# This doesn't have the test deps
$ opam install cinaps 
# No findlib apparently 
$ opam install ocamlfind 
$ dune runtest
No such package: base
$ dune runtest
```

Then I went for a `opam install base` and ran `dune runtest` and got the same errors.

NathanReb:
ah right so base actually is an undeclared test dependency of ours but the relevant tests fail silently when it's not installed.

Thanks for your help, I'll look into it!

NathanReb:
Ok circling back to this. I had initially misunderstood your comments.
1. The `No such package: base` comes from a different test, I removed the `#require` that was causing this as it was not necessary anyway.
2. The test failure I copied above does not happen if base is not installed initially but it consistently appears once it has been installed even if we remove it. I do still see the errors after removing base myself. This is quite puzzling.

NathanReb:
Indeed, starting from a fresh switch worked and the tests pass but as soon as I install base it fails again, even though we don't link base.

NathanReb:
I'm not sure how base is messing up with the exception printing here but it so specific and only relevant to one test that I don't think it's worth investigating this any further, especially since everything works perfectly well when installing our deps and test deps only.

Let's merge!

PR #557, opened 2025-01-24
Re-enable CI builds for 4.07-4.09

NathanReb:
This is an attempt at re-enabling the CI builds for OCaml 4.07 to 4.09. Thanks to ocaml-ci maintainers we now know that this was caused by ppxlib-bench not being co-installable with the other packages on those versions.

I'll try different solutions here.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -64,7 +64,7 @@ format.")
 than ppxlib itself. Also the benchmark runner has dependencies that ppxlib doesn't
 have.")
  (depends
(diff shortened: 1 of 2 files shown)


NathanReb:
Screenshot from 2025-01-24 11-14-49
Aleluia!

NathanReb:
I'm expecting some test to break as I ran into expect test failures when manually testing on 4.08 recently.

PR #558, opened 2025-02-05
5.3 AST bump

patricoferris:
There are only two significant changes to the AST from 5.2 to 5.3: the addition of a dedicated effects AST node and constant nodes getting a location. The latter will likely be the most impacting change to downstream dependencies of ppxlib.

Of course, we need to get #514 in first, but this is ready for when we want to make the next change and catch up with ocaml/ocaml, from there we should be able to have a more regular AST bump cadence (depending on how the parsetree changes).

### DIFF
--- a/ast/ast_helper_lite.ml
+++ b/ast/ast_helper_lite.ml
@@ -17,7 +17,7 @@
 open Stdlib0
 module Location = Astlib.Location
 module Longident = Astlib.Longident
-open Astlib.Ast_502
+open Astlib.Ast_503
 
 [@@@warning "-9"]
 
@@ -52,16 +52,21 @@ let protect_ref =
 let with_default_loc l f = protect_ref (R (default_loc, l)) f
 
 module Const = struct
-  let integer ?suffix i = Pconst_integer (i, suffix)
-  let int ?suffix i = integer ?suffix (Int.to_string i)
-  let int32 ?(suffix = 'l') i = integer ~suffix (Int32.to_string i)
-  let int64 ?(suffix = 'L') i = integer ~suffix (Int64.to_string i)
-  let nativeint ?(suffix = 'n') i = integer ~suffix (Nativeint.to_string i)
-  let float ?suffix f = Pconst_float (f, suffix)
-  let char c = Pconst_char c
+  let mk ?(loc = !default_loc) pconst_desc = { pconst_desc; pconst_loc = loc }
+  let integer ?loc ?suffix i = mk ?loc (Pconst_integer (i, suffix))
+  let int ?loc ?suffix i = integer ?loc ?suffix (Int.to_string i)
+  let int32 ?loc ?(suffix = 'l') i = integer ?loc ~suffix (Int32.to_string i)
+  let int64 ?loc ?(suffix = 'L') i = integer ?loc ~suffix (Int64.to_string i)
+
+  let nativeint ?loc ?(suffix = 'n') i =
+    integer ?loc ~suffix (Nativeint.to_string i)
+
+  let float ?loc ?suffix f = mk ?loc (Pconst_float (f, suffix))
+  let char ?loc c = mk ?loc (Pconst_char c)
 
   let string ?quotation_delimiter ?(loc = !default_loc) s =
-    Pconst_string (s, loc, quotation_delimiter)
+    (* Should we have a different loc for the constant ? *)
+    mk ~loc (Pconst_string (s, loc, quotation_delimiter))
 end
 
 module Attr = struct
--- a/ast/ast_helper_lite.ml
+++ b/ast/ast_helper_lite.ml
@@ -65,7 +65,6 @@ module Const = struct
   let char ?loc c = mk ?loc (Pconst_char c)
 
   let string ?quotation_delimiter ?(loc = !default_loc) s =
-    (* Should we have a different loc for the constant ? *)
     mk ~loc (Pconst_string (s, loc, quotation_delimiter))
 end
 
--- a/src/pp_ast.ml
+++ b/src/pp_ast.ml
@@ -205,6 +205,10 @@ class lift_simple_val =
         ~lift_record:super#module_expr ~desc:mod_.pmod_desc
         ~attrs:mod_.pmod_attributes mod_
 
+    method! constant c =
+      self#lift_record_with_desc ~lift_desc:self#constant_desc
+        ~lift_record:super#constant ~desc:c.pconst_desc ~attrs:[] c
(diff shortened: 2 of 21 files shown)


notxvilka:
Hi, any chance for this to be merged soon?

NathanReb:
This is next on our list but we're first making sure the ecosystem had enough time to adapt to the 5.2 bump.

ast/ast_helper_lite.ml
```diff
@@ -52,16 +52,21 @@ let protect_ref =
...
+
+  let nativeint ?loc ?(suffix = 'n') i =
+    integer ?loc ~suffix (Nativeint.to_string i)
+
+  let float ?loc ?suffix f = mk ?loc (Pconst_float (f, suffix))
+  let char ?loc c = mk ?loc (Pconst_char c)
 
   let string ?quotation_delimiter ?(loc = !default_loc) s =
-    Pconst_string (s, loc, quotation_delimiter)
+    (* Should we have a different loc for the constant ? *)
```
NathanReb, line 68: I think the constant loc should include the delimiters, same as the surrounding expr, so yes, I'm guessing it should have a different loc!

NathanReb, line 68: I think the constant loc should include the delimiters, same as the surrounding expr, so yes, I'm guessing it should have a different loc!

test/ppxlib-pp-ast/basic.t
```diff
@@ -48,7 +53,10 @@ But it can also be used to pretty a single expression:
   Pexp_apply
     ( Pexp_ident (Lident "+")
     , [ ( Nolabel, Pexp_ident (Lident "x"))
-      ; ( Nolabel, Pexp_constant (Pconst_integer ( "2", None)))
+      ; ( Nolabel
+        , Pexp_constant
+            { pconst_desc = Pconst_integer ( "2", None); pconst_loc = __loc}
```
NathanReb, line 58: We should probably add a special case to `Pp_ast` here as well, so that unless specified otherwise, the `pconst_desc` is printed directly, without the wrapping record.

traverse_builtins/ppxlib_traverse_builtins.ml
```diff
@@ -12,6 +12,9 @@ class map =
   let any x = x in
   object
     method int : int T.map = any
+    method int32 : int32 T.map = any
```
NathanReb, line 15: These seem like a worthy addition but I think I missed why we need them now, could you provide some context?

NathanReb (commented):
Looks great! I have a few minor remarks but other than that it should be good to go!

traverse_builtins/ppxlib_traverse_builtins.ml
patricoferris: Hmm I have no idea, I can actually remove them and everything still builds! Perhaps I was working on something else at the time and thought I needed them. Happy to remove them from this PR :) ?

bin/pp_ast.ml
```diff
@@ -97,6 +97,12 @@ let loc_mode =
   in
   named (fun x -> `Loc_mode x) Cmdliner.Arg.(value & vflag `Short [ full_locs ])
 
+let full_consts =
```
NathanReb, line 100: I don't think this is necessary, since the only other field is the location, we can make the `show-loc` flag control whether we skip the constant record or not, similar to what we do with over records with a `_desc` field.

traverse_builtins/ppxlib_traverse_builtins.ml
NathanReb: Yeah I guess if we don't use them we might as well remove them!

src/pp_ast.ml
```diff
@@ -205,6 +216,23 @@ class lift_simple_val =
         ~lift_record:super#module_expr ~desc:mod_.pmod_desc
         ~attrs:mod_.pmod_attributes mod_
 
+    method! constant_desc (c : constant_desc) =
+      if config.full_consts then super#constant_desc c
+      else
+        match c with
+        | Pconst_integer (i, None) -> self#int (int_of_string i)
```
NathanReb, line 223: I had something simpler in mind, i.e. still print the `_desc` part as we used but only skip the wrapping record.

I think printing the actual constant `_desc` is useful for users that need to discover how they are represented. What do you think?

patricoferris: Yep, I've just used the existing logic for this

NathanReb (approved):
Perfect, let's merge!

NathanReb:
I'm happy to help debug the homebrew issue though, might need a bit of context but it seems ppxlib isn't properly built there.

notxvilka:
> I'm happy to help debug the homebrew issue though, might need a bit of context but it seems ppxlib isn't properly built there.

This is the error that updating to OCaml 5.3 produced:
```
  Error: # env-file             /private/tmp/opamroot20250501-87349-oia9xr/log/ppxlib-8069-d57ac3.env
  # output-file          /private/tmp/opamroot20250501-87349-oia9xr/log/ppxlib-8069-d57ac3.out
  ### output ###
  # (cd _build/default && /private/tmp/opamroot20250501-87349-oia9xr/default/bin/ocamlc.opt -w -9 -g -bin-annot -I astlib/.astlib.objs/byte -I /private/tmp/opamroot20250501-87349-oia9xr/default/lib/ocaml-compiler-libs/common -I /private/tmp/opamroot20250501-87349-oia9xr/default/lib/ocaml/compiler-libs -no-alias-deps -open Astlib__ -o astlib/.astlib.objs/byte/astlib__Ast_501.cmo -c -impl astlib/as[...]
  # File "astlib/ast_501.ml", lines 245-299, characters 2-44:
  # Error: This variant or record definition does not match that of type
  #          Parsetree.pattern_desc
  #        An extra constructor, Ppat_effect, is provided in the original definition.
  
  
  
  <><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
  ┌─ The following actions failed
  │ λ build ppxlib 0.36.0
  ```

NathanReb:
This looks like the build isn't correctly setup here somehow. It seems to assume it is building on OCaml 5.1 and therefore declares type equalities between the 5.1 copy of the AST types and the compiler ones, which does not work since the compiler is 5.3.

Can you provide a bit more context there on how those builds are setup?

ppxlib.0.35.0 and ppxlib.0.36.0 build perfectly well on a 5.3 switch. Might be a problem with dune as well since we rely on it to get the compiler version in our `astlib/dune` rules.

PR #559, opened 2025-02-05
Add flag `-raise-embedded-errors` to `Driver`

jaymody:
Adds a flag `-raise-embedded-errors` to the shared args in `Driver`, which will raise the first `ocaml.error` node found in the processed AST.

### Motivation
It's useful to be able to run a standalone ppx driver for it's side affects (e.g. registering correction files), without actually making use of it's output directly.

For example, given the program `program.ml`:

```ocaml
type t = int [@@deriving_inline my_deriver]
```

One might run:

```bash
$ my_ppx.exe program.ml -null
```
With the objective of registering a correction file while ignoring any output. However, since errors are embedded into the AST, the error `[%%ocaml.error "ppxlib: [@@@deriving.end] attribute missing"]` never gets surfaced to the users. Now it can be, with `-raise-embedded-errors`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 unreleased
 ----------
 
+- Add a `-raise-embedded-errors` flag to the driver. Setting this flag raises the first
+  `ocaml.error` embedded in the final AST.
+
 - Export `Ast_pattern.fail`. (#563, @ceastlund)
 
 - Make `Ast_traverse.sexp_of` more concise, and add a test. (#561, @ceastlund)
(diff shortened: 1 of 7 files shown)


NathanReb (approved):
It's a bit annoying to have to add yet another flag for this but I guess for now there's no other way around it.

I'd like to have a unique way to produce errors via ppxlib that would let the driver decide whether it embeds them or raise them straight away but that would require the whole ecosystem to update so it won't happen overnight.

NathanReb:
Could you rebase and add a changelog entry?

The changelog check is not picking it up because the PR is based on a version where I had not re-introduced the `unreleased` changelog section.

NathanReb (approved):
Perfect, thanks!

PR #561, opened 2025-02-07
Concise Ast_traverse.sexp_of

ceastlund:
As tests in the first commit show, `Ast_traverse.sexp_of` is ridiculously verbose. The output is overwhelmed with empty fields, the fields of location records, and so forth. This feature applies several shorthands to the sexp that make it much more manageable, while still uniquely expressing a value.

1. Longidents are rendered as a single string, unless parsing fails to round-trip.
2. `Location.none` is elided as `()`.
3. Other locations are rendered as strings, with suffix `<ghost>` if `loc_ghost=true`.
4. Records drop fields containing only the sexp `()`.
5. Records with only one remaining field named `txt` or `*_desc` reduce to that fields' value.

After these changes, the test output goes from so long OCaml won't print it to only a few times larger than the `Pprintast` output.


NathanReb (approved):
Looks good to me!

PR #562, opened 2025-02-10
Update README.md: Mention the Dune dependency mechanism `(env_var MY_VAR)`

lukstafi:
Mention the Dune dependency mechanism `(env_var MY_VAR)`, that makes `ppx_get_env` more reasonable.


lukstafi:
Sorry I messed up signing.

PR #563, opened 2025-02-11
Export Ast_pattern.fail.

ceastlund:
Without this function it was hard to write new pattern matchers using `of_func`.


src/ast_pattern.mli
```diff
@@ -210,3 +210,6 @@ type context
 
 val of_func : (context -> Location.t -> 'a -> 'b -> 'c) -> ('a, 'b, 'c) t
 val to_func : ('a, 'b, 'c) t -> context -> Location.t -> 'a -> 'b -> 'c
+
+val fail : Location.t -> string -> _
+(** Call from [of_func]'s argument when the pattern does not match. *)
```
NathanReb, line 215: Might be worth mentioning the `string` argument should describe the expected AST's shape.

NathanReb (approved):
This looks good! 

Just a comment on the functions documentation but besides that it's good to go!

PR #569, opened 2025-03-25
Fix location inserted when migrating constants from 5.2 to 5.3

NathanReb:
We should never insert `Location.none` and should at least aim at having a loc that points to the right file. Inserting such nodes is bound to produce very unhelpful errors for end users.

This fixes the 5.2 -> 5.3 migration so that when copying constants from expressions, patterns and interval patterns, the constant location is a ghost location pointing the closest surrounding expression or pattern in the AST. Most of the time, this will be the correct location for the constant, except in interval patterns.

This should prevent users from getting unusable error messages.

This will also fix the test failures in #558

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 unreleased
 ----------
 
+- Fix 5.2 -> 5.3 migration of constants. Those used to always have a `none`
+  location which can lead to unhelpful error messages.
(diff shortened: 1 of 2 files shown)


patricoferris (approved):
Sorry for the slow review @NathanReb -- good catch!

PR #566, opened 2025-02-18
Use setup-ocaml@v3 for trunk build

patricoferris:
The darcs issue seemed related to https://github.com/ocaml/setup-ocaml/issues/923

I also bumped the version of the trunk compiler we are building against.

PR #570, opened 2025-04-04
Bump trunk CI build to 5.4.0+trunk

patricoferris:
We know have a dune 3.18.0 with 5.4.0 support (afaict). See the conversation in https://github.com/ocaml-ppx/ppxlib/pull/566


NathanReb:
There are a few missing promotions it seems!

patricoferris:
I think we're just waiting for the magic numbers to be bumped in OCaml.

NathanReb:
> I think we're just waiting for the magic numbers to be bumped in OCaml

I thought they were bumped on the development branch now.

It also seems the dune version we're using for the trunk CI build does not build.

astlib/longident.ml
```diff
@@ -1,5 +1,73 @@
...
+(*   the GNU Lesser General Public License version 2.1, with the          *)
+(*   special exception on linking described in the file LICENSE.          *)
+(*                                                                        *)
+(**************************************************************************)
+open Location
+
+type t = (*IF_AT_LEAST 504 Ocaml_common.Longident.t = *)
+    Lident of string
+  | Ldot of t loc * string loc
+  | Lapply of t loc * t loc
```
NathanReb, line 20: I think we should have a `Longident` module in `ast_504` similar to what we have for `Asttypes`. That way if the type is further modified, it'll be treated as a regular AST change.

I think we should still have a `Longident` module in `astlib` but it should only provide the utility functions on Longidents from the compiler, same as the `Parse`, or `Pprintast` modules do for other moving Ast types.

Since Astlib is supposed to provide the stable layer on top of the compiler-libs that's necessary for ppxlib, we should maybe make it provide the exact set of functions we need and nothing more.

What would you think about the following:
- Make Legacy_longident define only the type
- Make `Longident` provide the functions we use, conditionnally compiled to work with the current compiler
- Have a `Longident` module in `Ppxlib` that expose the functions corresponding to our internal AST.

As I write this I wonder if, in fact, we don't need any compiler functions on longidents but simply need the ones for our internal AST. Making a compat layer for functions unnecessary as we can simply have a local copy of the correct version.

test/expect/expect_test.ml
```diff
@@ -162,8 +162,10 @@ let main () =
         Oval_variant (name, Some (map_tree param))
     | Oval_string (s, maxlen, kind) ->
         Oval_string (s, (if maxlen < 8 then 8 else maxlen), kind)
-    | Oval_tuple tl -> Oval_tuple (List.map ~f:map_tree tl)
-    | Oval_array tl -> Oval_array (List.map ~f:map_tree tl)
+    (*IF_NOT_AT_LEAST 504 | Oval_tuple tl -> Oval_tuple (List.map ~f:map_tree tl) *)
+    (*IF_AT_LEAST 504 | Oval_tuple tl -> Oval_tuple (List.map ~f:(fun (l, v) -> (l, map_tree v)) tl) *)
```
NathanReb, line 166:
```suggestion
    (*IF_AT_LEAST 504 | Oval_tuple tl -> Oval_tuple (List.map ~f:(fun (label, v) -> (label, map_tree v)) tl) *)
```

test/expect/expect_test.ml
```diff
@@ -162,8 +162,10 @@ let main () =
...
+    (*IF_NOT_AT_LEAST 504 | Oval_array tl -> Oval_array (List.map ~f:map_tree tl) *)
+    (*IF_AT_LEAST 504 | Oval_array (tl, v) -> Oval_array ((List.map ~f:map_tree tl), v) *)
```
NathanReb, line 168:
```suggestion
    (*IF_AT_LEAST 504 | Oval_array (tl, mutable_) -> Oval_array ((List.map ~f:map_tree tl), mutable_) *)
```

NathanReb (commented):
Looks great! Thanks for taking the time to work on this.

From the top of your head, do you know which functions in the migration need to be reviewed?

kit-ty-kate:
the magic numbers have been bumped

patricoferris:
Also, we're waiting on a new version of dune that support 5.4.0 too.

kit-ty-kate:
the dune PR has been merged in opam-repository

PR #572, opened 2025-04-24
Update ubuntu version used for changelog related actions

NathanReb:
The actions weren't run anymore because we were using an old, fixed version of ubuntu.


patricoferris (approved):

patricoferris:
Perhaps it only takes effect at the moment of opening a PR, I'm going to merge and see if new PRs are fixed. Thanks !

PR #574, opened 2025-04-30
Add context-free rule type that replaces AST nodes with attributes

Skepfyr:
This adds a new kind of context-free rule that allows PPX authors to write PPXs that replace AST nodes that have their attributes on them. You can already achieve something like this with extension nodes, but the syntactic overhead for that is very high for some PPXs.

I wanted this specifically for [ppx_template](https://github.com/janestreet/ppx_template), which uses attributes to add suffixes to identifiers, adding an extension node is too heavyweight as the attributes need to get sprinkled all over your code.


src/attribute.ml
```diff
@@ -823,7 +823,7 @@ let collect =
         =
       let loc = Common.loc_of_attribute attr in
       super#payload payload;
-      Attribute_table.add not_seen name loc
+      Attribute_table.replace not_seen name loc
```
Skepfyr, line 826: To clarify this change: some attributes get duplicated in the AST (that is, they point to the same location) but we only want to include them once in `not_seen` so that `remove_seen` works as expected.

NathanReb:
I think this introduces a bit of confusion as to what the expanded code will look like. The clear split between attributes and extension in our rules so far also helps understand what part of the code stays, and what gets rewritten.

ppx_template does not have documentation yet. Could you give me an example of how it's used and what kind of syntactic overhead it introduces when using extension points?

NathanReb:
Reading the opam description of the package:
> This PPX rewriter is used internally at Jane Street to simulate polymorphism over various extensions to OCaml's type system, in cases where there is not yet built-in compiler support. The long-term goal is thus to delete it entirely.

I realize this is meant to be a temporary work around to current compiler limitations. I'm a bit more reluctant to add this to ppxlib if the ppx that justifies this feature is meant to be deleted in the end.

Are none of the work around available satisfactory at the moment?

One other option is to use a whole AST rewriting pass to deal with those attributes.

Skepfyr:
I don't quite understand what you mean by:
> I think this introduces a bit of confusion as to what the expanded code will look like. The clear split between attributes and extension in our rules so far also helps understand what part of the code stays, and what gets rewritten.

Are you saying that currently only extension nodes are rewritten and attributes only ever add code, and that split is useful?

ceastlund:
I think _in general_, it is useful for extensions to rewrite their contents and attributes to be metadata. I think for our use case and perhaps some others, attributes are more natural. In part, it's because we are rewriting the attached expression/etc. to _some version of_ itself, rather than to some new thing. So for example `module M : S [@mode portable]` reads more clearly as `module M : S (* where [S] has some [portable] notion to it *)`, whereas `module M : [%mode (S @ portable)]` reads as `module M : _ (* where [_] is in some way generated from the names [S] and [portable] *)`.

I'd like us to add the ability to register attribute transformers, but perhaps we should add caveats to the documentation explaining it should be used sparingly, and only for cases where one is, essentially, transforming likes-to-likes, or a thing to a version of itself.

As for `ppx_template` being temporary---I think any given usage of it is probably temporary, but I think the ppx itself will live on indefinitely, as different use cases cycle in and out of it. So we do want long-term support for this.

@NathanReb, how does this sound?

ceastlund:
We should also add some caveat in `context_free.mli`.

src/context_free.ml
```diff
@@ -440,6 +492,60 @@ let context_free_attribute_modification ~loc =
...
 
+let handle_attr_replace_once context attrs item base_ctxt : 'a option t =
+  let result =
+    List.find_map attrs ~f:(fun (Rule.Attr_replace.T a) ->
+        let rec get_attr_payloads : type l.
+            ('a, l) Rule.Attribute_list.t ->
+            (bool * l Rule.Parsed_payload_list.t) t = function
+          | [] -> return (false, Rule.Parsed_payload_list.[])
+          | x :: xs ->
+              (if Attribute.Context.equal context (Attribute.context x) then
```
NathanReb, line 503: If I'm not mistaken, this shouldn't be necessary. The type of `attrs` must match the type of `context`.

src/context_free.ml
```diff
@@ -8,6 +8,34 @@ module A = Attribute
 module AC = Attribute.Context
 
 module Rule = struct
+  module Attribute_list = struct
```
NathanReb, line 11: This and the `Parsed_payload_list` seem extremly convoluted to me and I'm not sure they bring a lot of value.

I might be wrong but it seems the same thing can be achieved using a single attribute and passing all the config in the payload, e.g. instead of:
```ocaml
let x = y [@prefix "pre"][@suffix "post]
```
one could write:
```ocaml
let x = y [@fix ~pre:"pre" ~suf:"post"]
```
or something along those lines.

It would also clarify things greatly from a reader perspective, as it's now obvious this is a single transformation handling both, the name of the attribute likely hinting to the name of the ppx responsible for handling it.

src/context_free.ml
```diff
@@ -440,6 +492,60 @@ let context_free_attribute_modification ~loc =
         "A context-free rule deleted or added attribues of a str/sig item",
       [] )
 
+let handle_attr_replace_once context attrs item base_ctxt : 'a option t =
```
NathanReb, line 495: Do you think this could be integrated into `map_node` together with extensions handling so that we have a single function responsible for rewriting a node until it reaches a fix point.

NathanReb:
The documentation looks fine, thanks for taking the time to add this!

src/context_free.ml
ceastlund: What's going on here, is that we have multiple attributes that are often used separately, but may be used together. We want each one to be succinct. When transformed as just a collection of attributes, they need to be combined to get the right result. But we don't want
```
x [@foo a], y [@bar b], z [@foo a] [@bar b]
```
to have to be written as
```
x [@foobar ~foo:a], y [@foobar ~bar:b], z [@foobar ~foo:a ~bar:b]
```
That's much more verbose. The fact that these things are transformed together is sort of an implementation detail, and I don't think it needs to creep into the notation.

src/context_free.ml
ceastlund: I think lifting helpers out is a good thing, no? We don't need everything to be one great big recursive ball of twine?

NathanReb: I'm not saying this function shouldn't exist. I'm suggesting that we handle node rewriting and recursion in a single function, rather than having to call this and `map_node` for every single method.

It also seems that the current way things are handled, we will expand attr-based rewriting first, until we reach a fix point, then extension rewriting, until we reach a fix point. That means that if an extension wanted to produce nodes with attr, it wouldn't work. There's now a transformation ordering issue, which we should not have with context-free rules.

What I'm suggesting here is that we have a `map_node` function responsible for determining if a node needs rewriting (be it an extension rule or attribute rule), applying the rule, recursively call itself on the result of the expansion.
I'm perfectly okay with attributes having priority over extensions but I don't think they should be handled entirely separately in a different phase.

src/context_free.ml
NathanReb: I think this is still achievable with the suggested API. One could register rules for the individual attributes for conciseness and have a main attribute that lets you combine them.

As you mention, the fact that you have to handle them together is an implementation detail, it probably shouldn't leak into ppxlib public API.

Would something like the following example be acceptable?

E.g. you'd write:
```ocaml
x [@foo a], y [@bar b], z [@foobar ~foo:a ~bar:b]
```

This would also likely make the ppx code a bit simpler as you don't have to handle the GADT lists in there anymore either.

Skepfyr: The issue is that people would use those shorter attributes together, which means that the ppx code would need to handle having multiple attributes that it cares about on one node. every ppx will have to write code that either tries to consume all of its other attributes or at least spots them and emits an error. The idea behind this code was to move that code into ppxlib so that the individual ppxs don't have to worry about it.

I agree the GADT lists are a tad complicated to look at, however they're surprisingly easy to use, which is what convinced me that this is probably fine, because they define `[]` and `( :: )` you can basically just pretend they're normal lists in the ppx code, but just get a compile error if it's the wrong length or has a wrong element type.

ceastlund: Perhaps we should (1) put the single-attribute case first and document that, then (2) present the multiple-attribute case in a submodule. The submodule's documentation can describe a use case and show how it works, so by the time the user sees the GADTs and the function signature, it's much clearer what it's all for.

src/context_free.ml
ceastlund: Ah, I had misunderstood. Yes, if we've put this at the wrong place in the expansion logic, we should move it.

src/context_free.ml
ceastlund: @NathanReb does that sound sufficient to you?

NathanReb: @ceastlund Yes that sounds good to me. @Skepfyr made a great argument there as well and I can see how neat it is to use.

@patricoferris also pointed out that we have an `Expert` module in `Extension`. We could do something similar here.

I'm also wondering if we shouldn't move parts of this to its own file as it describes a specific type of rules, just as extensions do.

Skepfyr: I've extracted it out into a submodule. I don't think creating an entirely separate file like extension or deriving is quite worth it because there's a lot less interface, but happy to move it if you think otherwise.

src/context_free.ml
NathanReb: Thanks, it looks much better! I think it's okay for signature and structure, I'll refactor those eventually!

Could you add a test that shows the recursion works correctly? I think this is fairly easy to break inadvertently so I'd feel much better if it were tested.

I.e. we need a test that shows that if `[%ext ...]` produces `x [@rewrite]`, the attribute based rewrite does happen.

src/context_free.ml
NathanReb: Looks good to me, thanks!

Skepfyr:
Do you know a solution to https://github.com/ocaml-ppx/ppxlib/pull/574#discussion_r2114186995? (I'm mostly linking it because GitHub makes it easy to lose some review comments)

src/attribute.ml
ceastlund: If we take it out, does something fail internally here at Jane Street? If so, can we turn that into an external reproduction? If not, should this fix just be a separate PR?

src/context_free.ml
NathanReb: Awesome thanks!

src/attribute.ml
NathanReb: If I understand the problem correctly our tests would need to run on OCaml <= 5.0, parse a source file with an attribute set as you described above, and have a ppx driver run on it and "consume" the attribute. Before your fix this should trigger the bug and we should get an unused attribute warning because of the second "shadow" attribute not being marked as seen. After your fix this should go away. Am I correct?

I think you can simply restrict the compiler version under which the test is run, using dune's `enabled_if` stanza field, as it is done [here](https://github.com/ocaml-ppx/ppxlib/blob/main/test/502_pexpfun/dune#L10) for example.

You make it sound like it's trickier than that so I might have missed something.

Either way, I think this would be best fixed in its own PR as @ceastlund suggests!

NathanReb: I'm fine merging this as is, the change seem sane. I was suggesting we add a test to prevent reintroducing it someday but if it's too cumbersome to write a meaningful test, let's go ahead and merge!

NathanReb (approved):
Thanks, let's merge!

PR #575, opened 2025-04-30
Remove ASTs from 402 to 407

patricoferris:
Since we only support OCaml >=4.08, this is all deadcode and only makes grepping through the codebase harder.


NathanReb (approved):
Thought I did that as part of #556 but I only removed the migrations! Thanks for fixing this!

Issue #546, opened 2024-12-17
Extend deriving API to work on modules


benbellick:
Hey @NathanReb. Thanks again for the awesome work here. 
In light of [this comment](https://github.com/ocaml-ppx/ppxlib/issues/529#issuecomment-2373997825), I wanted to see if it is possible after all (and not too difficult) to extend the deriving API to operate on modules, rather than only module types. 

FYI, if you're curious to see the repo where I am hoping to use this, it is [here](https://github.com/benbellick/ppx_deriving_decoders). 

As I mentioned in the prior conversation, I would love to be able to do the following: 
```ocaml
module X = struct 
  type t = int * string 
end [@@deriving decoders]
```
--via-ppx-->
```ocaml
module X = struct 
  type t = int * string 
  module Encode (D : Decoders.Decode.S) = struct 
    let t_decode = <SOME IMPLE USING D>
  end 
end
```

Please let me know if this is possible and if there is time for it! Happy to help also if there is an easy way to jump into the codebase.

patricoferris:
@benbellick this seems very reasonable to me. I don't think @NathanReb and I currently have the bandwidth for this, but an implementation is going to look very similar to #538 only instead of class type declarations you will want module implementations -- following that PR might be the easiest way to jump in.

patricoferris:
@benbellick this feature has been added in #576 and should be in the next release.

Note that, just like any deriving code, you can only append new things after the thing you are deriving on. So in this instance you cannot "inject" code into the module via the deriving on the module itself. I reckon, with the right amount of `include`s you could fashion a work around though.

Let me know if you need any help, cheers :)

PR #576, opened 2025-05-05
Add module bindings and module declarations to deriving

patricoferris:
In https://github.com/ocaml-ppx/ppxlib/issues/546 a feature request was made for `[@@deriving ...]` to work on module bindings (and perhaps their signature counterpart declarations). 

Some notes:
 - I had to disable the cinaps str_to_sig stuff as module bindings and declarations are not as nicely symmetric as other types.
 - Like other `[@@deriving]` features, derived code is placed after the node, therefore using this with module bindings will not place the code _inside_ the module definition but after it. I think some amount of `include`ing can work around this maybe (on the end user's side)?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,9 @@ unreleased
 - Add a new context-free rule type that replaces AST nodes that have the registered
   attributes attached to them. (#574, @Skepfyr)
 
+- Allow users to derive code from module bindings and module declarations
+  (#576, @patricoferris)
+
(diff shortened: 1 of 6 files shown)


NathanReb (approved):
This looks good to me!

I wasn't quite sure initially whether we should provide those or stick to module type declaration, symmetrically to what we do with types (i.e. [@@deriving] works on type declaration, not value bindings) but I guess module are a bit different and it does not really hurt to allow it as a ppx framework.

PR #580, opened 2025-05-23
Remove unused code from Code_path

davesnx:
I belive got added by accident here: https://github.com/ocaml-ppx/ppxlib/commit/3e0f8018f867bcc35357928e0383a8fd380f4943 as there's some inline expected tests in the same commit


patricoferris (approved):
Thanks @davesnx :)

Could you sign this commit for the DCO check?

davesnx:
Done @patricoferris

PR #583, opened 2025-06-25
Update pprintast

NathanReb:
Our local copy of pprintast for 5.3 was patched from the 5.2 version rather than simply copied from the compiler's own.

This updates our `pprintast` copy and adds a short comment documenting the update process.

### DIFF
--- a/astlib/pprintast.ml
+++ b/astlib/pprintast.ml
@@ -21,73 +21,19 @@
 (* Extensive Rewrite: Hongbo Zhang: University of Pennsylvania *)
 (* TODO more fine-grained precedence pretty-printing *)
 
+(* This file was copied from OCaml 5.3's pprintast.ml and modified in the
+   following ways:
+   - Added [open Ast_503] before other global opens
+   - Replaced [Lexer.is_keyword] with [Keyword.is_keyword] for compat with
+     Ocaml < 5.2.
+   - Added [class_signature] and [type_declaration] entry points at the end. *)
+
 open Ast_503
 open Asttypes
 open Format
 open Location
 open Longident
 open Parsetree
-
-module Option = struct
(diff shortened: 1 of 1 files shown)


patricoferris (approved):
Thanks @NathanReb -- this is definitely needed and rather tedious. The documentation looks good.

patricoferris:
It might not be a crazy idea to add a reminder at the top of https://github.com/ocaml-ppx/ppxlib/blob/main/ast/ast.ml too as this pprintast file is likely to need updating on an AST bump too.

hhugo:
Do you plan to release a 0.36.1 version with this fix ? The gen_js_api testsuite is currently failing with ppxlib.0.36

NathanReb:
I was planning on releasing this as part of 0.37 but I can cut a 0.36.1!

NathanReb:
> Do you plan to release a 0.36.1 version with this fix ? The gen_js_api testsuite is currently failing with ppxlib.0.36
Wait, now that I think about it, this can't be the issue here as this patch should only be necessary post 5.3 bump, so only on ppxlib's main branch.

Do you have a pointer to the test failures so I can try to debug those?

hhugo:
Running gen_js_api tests with 5.3+ppxlib.0.36 should give you the repro

NathanReb:
Some of those seem legit, like:
```diff
-        fun _ -> fun _ -> Obj.magic
+        fun _ _ -> Obj.magic
```
Is probalby just tied to the AST bump to 5.2 and the arity representation changes.

```
-let (_ : (string * int) to_js) =
+let _ =
   fun (x13 : (string * int)) ->
```
On the other hand might come from a bug, I'll need to investigate this a bit further and look at the how the tests are run.

Do you get the same output regardless of the compiler being used?

PR #587, opened 2025-07-10
Add dev tool to generate dummy migration

NathanReb:
This tool is inspired from the omp version of old.

It is invoked as follows:
```
dune exec -- dev/gencopy/gencopy.exe astlib/ast_X.ml astlib/ast_Y.ml > migrate_X_Y.ml
```
It will generates deep "copy" functions from ast version X to version Y. These are dummy function that are generated under the assumption that the ast types in X and Y are identical.

When adding support for new compiler versions, this should be used as a base to be editted, following the compiler errors that will point to AST nodes that actually differ between the two versions.

## Differences with the OMP tool

The OMP tool used the compiler libraries to load compiled interfaces and a proper typing environment and used that to recursively generate copy functions for any types it encountered, starting from a given set of entry point types. This ensured it wouldn't generate unnecessary functions.

It turns out that this method as one major draw back: the tool heavily relied on unstable libraries and therefore could only build on a fixed version of OCaml. It was also a bit more complex than this new approach which does not bother with typing env lookups.

This version is much simpler, it uses the ppxlib driver to parse the `ast_x.ml` source file to our internal AST and generates the copy functions from the type declarations, all in one big set of recursive value bindings.

In practice this makes almost no difference as we end up needing copy functions for every type anyway. Worst case scenario: unused functions will be picked up by the compiler and can be deleted manually.
One advantage the old approach had is that it was able to generate copy functions for external types such as location, which we don't do anymore. Once again, in practice it doesn't matter much as location has not changed and therefore can simply be copied using the identity function since it's not versioned.

One thing to note is that it completely changes the order in which the copy functions appear in a `migrate_X_Y.ml` file compare to what we used to have.

### DIFF
--- a/dev/gencopy/gencopy.ml
+++ b/dev/gencopy/gencopy.ml
@@ -107,21 +107,27 @@ module Copy = struct
     let vars = List.mapi (fun i _ -> ptuple_var i) typs in
     Ast_builder.ppat_tuple vars
 
-  let copy_var ~decl var_name =
+  let param_index ~decl var_name =
+    let rec find_index i p l =
+      match l with
+      | [] ->
+          failwith
+            (Printf.sprintf "Could not find var '%s in %s" var_name
+               decl.ptype_name.txt)
+      | hd :: _ when p hd -> i
+      | _ :: tl -> find_index (i + 1) p tl
+    in
     let is_var (typ, _) =
       match typ.ptyp_desc with
       | Ptyp_var name -> String.equal name var_name
       | _ -> false
     in
-    let index = List.find_index is_var decl.ptype_params in
-    match index with
-    | None ->
-        failwith
-          (Printf.sprintf "Could not find var '%s in %s" var_name
-             decl.ptype_name.txt)
-    | Some i ->
-        let name = from_var_name i in
-        Ast_builder.pexp_ident (loc (Lident name))
+    find_index 0 is_var decl.ptype_params
+
+  let copy_var ~decl var_name =
+    let index = param_index ~decl var_name in
+    let name = from_var_name index in
+    Ast_builder.pexp_ident (loc (Lident name))
 
   let copy_t modname =
     Printf.sprintf "copy_%s" (String.uncapitalize_ascii modname)


dev/gencopy/gencopy.ml
```diff
@@ -0,0 +1,369 @@
...
+    let name =
+      match ptype_name with
+      | "t" -> String.lowercase_ascii context.inside
+      | _ -> ptype_name
+    in
+    Printf.sprintf "copy_%s" name
+
+  let arrow_type args ret =
+    let rev_args = List.rev args in
+    let rec aux acc args =
```
patricoferris, line 53: Could this be simplified to a `fold_left`: 

```ocaml
List.fold_left (fun acc hd -> Ast_builder.ptyp_arrow Nolabel hd acc) ret rev_args
```

dev/gencopy/gencopy.ml
```diff
@@ -0,0 +1,369 @@
...
+    let vars = List.mapi (fun i _ -> ptuple_var i) typs in
+    Ast_builder.ppat_tuple vars
+
+  let copy_var ~decl var_name =
+    let is_var (typ, _) =
+      match typ.ptyp_desc with
+      | Ptyp_var name -> String.equal name var_name
+      | _ -> false
+    in
+    let index = List.find_index is_var decl.ptype_params in
```
patricoferris, line 123: We're gonna need something for the older compiler we support for this function.

patricoferris (commented):
This looks great @NathanReb, I had a quick pass and it seems good to me. I agree that some tests would be very useful, as well as testing it against some previous migrations.

Re:`copy_location` -- I think that makes sense to me, some hardcoded parts are probably nothing to worry about given how slow moving the Parsetree is (except in the ways we don't want it to change!)

dev/gencopy/gencopy.ml
NathanReb: I can probably also remove the `List.rev` and make it a `List.fold_right` straight away!

PR #589, opened 2025-08-04
Fix Ast_builder's `value_binding` to generate the correct `pvb_constraint`

NathanReb:
I'd like to add some basic tests for this before we merge but it's basically a copy of the code that's handling the same conversion in the 5.0 -> 5.1 migration.


NathanReb:
I couldn't figure out a proper test for the coercion case and I suspect the logic to convert coercion expression is faulty or at the very least too restrictive.

I'm marking this as ready for review as I don't have a lot of time to spend on ppxlib atm and would still like to be able to release 0.36.2 and 0.37 shortly.

I'll open an issue regarding the conversion to the coercion constraint!

PR #593, opened 2025-08-28
Expose type equality of `Location.Error.t` between Astlib and Ppxlib.

ceastlund:
This provides access to some slightly lower level accessors.


NathanReb (commented):
Looks good to me! This probably deserves a changelog entry though!

NathanReb (approved):
Indeed! My bad!

PR #594, opened 2025-08-28
Add `@@@ppxlib.inline.end`, deprecate `@@@deriving.end`

ceastlund:
Replace `deriving.end` with `inline.end` so the attribute makes more sense in the context of, e.g., `@@expand_inline`. Deprecate the old name for the attribute, with a command-line flag to allow it.

### DIFF
--- a/src/code_matcher.ml
+++ b/src/code_matcher.ml
@@ -3,13 +3,25 @@ open! Import
 module Format = Stdlib.Format
 module Filename = Stdlib.Filename
 
-(* TODO: make the "deriving." depend on the matching attribute name. *)
+let allow_deriving_end = ref false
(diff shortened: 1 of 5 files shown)


panglesd:
I'm not sure how to read [this `[@@@deriving.end]` search](https://github.com/search?q=%40%40%40deriving.end+NOT+is%3Afork&type=code): it says 4.2k files, but only displays 5 pages of roughly 20 files per page...

(Note an ocaml.org tutorial mentioning `[@@@deriving.end]` that will have to be updated).

Sherlodoc finds [less results](https://sherlocode.com/?q=%5B%40%40%40deriving.end%5D).

Issue #597, opened 2025-09-29
Internal ppxlib migration attributes are sometimes left in the AST


patricoferris:
Ppxlib sometimes uses AST attributes to encode information needed to perform a correct round-trip migration. For example, to correctly roundtrip function arity, we make use of `ppxlib.migration.stop_taking` to indicate that we should "stop taking" function arguments:

https://github.com/ocaml-ppx/ppxlib/blob/a5be80d4c9f0ae9dc2afeb203c72b4df2a1dc0b3/astlib/migrate_502_501.ml#L156

In the corresponding, opposite migration (in this example, upwards from 501 to 502) there is code to make use of this attribute, and importantly _remove it_. 

However, this opposite migration does not always happen. For example, ppxlib 0.35.0 has an internal AST from OCaml 4.14.0. If ppxlib is used on a compiler 5.2.0 and above, only the downward migration will happen which introduces the `ppxlib.migration.stop_taking` attribute which is then never removed as we do not migrate the AST back up through 501->502.

We need to remove any extra information ppxlib encodes in the AST for migrations. There are at least three options:

1. Use a separate data-structure to encode extra information to help with migrations but that leaves the AST itself intact. 
2. Remove the extra information until it is needed. For example, for a vast majority of ppxlib use-cases, 0.35.0 does not need the extra migration information (unless someone is manually migrating the AST back up).
3. Have some function that removes all of this extra information and enforce some policy that users will have to call this function in order to have a _clean_ AST (this could perhaps be enforce via the type system?).

### Test Case

Using a 5.2.0 (or higher) compiler with ppxlib.0.35.0 installed, create a standalone executable:

```ocaml
let () = Ppxlib.Driver.standalone ()
```

Call it on the following `test.ml` file:

```ocaml
let f a = fun b -> a + b
```

Which produces the following:

```ocaml
let f a = ((fun b -> a + b)[@ppxlib.migration.stop_taking ])
```

Thanks to @n-osborne for setting me on this issue.

patricoferris:
Thanks @davesnx -- do you have a concrete example somewhere, cram tests for a particular project?

PR #598, opened 2025-09-30
Add remove functions for migration metadata

patricoferris:
Functions to help with #597. It is not quite clear to me when/if we should default to calling these except perhaps for the `use_compiler_pp` option.

The main reason for that is that the output of the ppx might be consumed by another ppx like tool (for example ppxlib-pp-ast) which makes use of the internal Ast_io function for re-reading the ppxlib output. Crucially, these functions then re-migrate the code to the internal AST at which point it will _need_ those attributes saved in the original ppxlib output.


NathanReb (approved):
Can you add a changelog entry ? Other than that it's good to go!

PR #599, opened 2025-09-30
Revert 5.3 AST Bump

patricoferris:
This reverts the bump of the internal AST to 5.3.

The only slight difference is that I copied the approach @NathanReb took to _adding_ the 5.3 pprintast module but applied it to the 5.2 pprintast module instead.


NathanReb (commented):
Could you also add the `0.36.1` and `0.36.2` changelog sections from our `0.36` branch as we now re-imported the features back into the main branch?

PR #601, opened 2025-10-08
Fix pprintast handling of raw identifiers

NathanReb:
This is imports upstream patches ocaml/ocaml#13604 and ocaml/ocaml#14279 that were done post 5.2 and improve pprintast's handling of raw identifiers.

It should fix the issue that fstar folks ran into with the 0.36.2 release, see the original discussion: https://github.com/ocaml/opam-repository/pull/28610#issuecomment-3362301901.

CC @mtzguido @mtelvers

### DIFF
--- a/test/pprintast/raw_identifiers/test.ml
+++ b/test/pprintast/raw_identifiers/test.ml
@@ -37,6 +37,27 @@ let stri =
 [%%ignore]
 
 Format.asprintf "%a" Pprintast.structure_item stri
+;;
 [%%expect{|
 - : string = "let f = (mod)"
+|}]
+
+let stri2 =
+  let open Build in
+  pstr_value
+    Nonrecursive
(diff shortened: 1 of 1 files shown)


NathanReb:
cc @Octachron let me know if I missed something here!

test/pprintast/raw_identifiers/test.ml
```diff
@@ -0,0 +1,42 @@
...
+;;
+[%%ignore]
+
+Format.asprintf "%a" Pprintast.expression expr
+;;
+[%%expect{|
+- : string = "10 mod 3"
+|}]
+
+(* [let f = (mod) *)
```
Octachron, line 28: I would propose to also test `(mod)` (or `lsl` for a change)  in pattern position `let f (lsl) = (mod)`.

PR #605, opened 2025-10-21
Fix Stdppx.Result.map_error

patricoferris:
reported by @ceastlund


NathanReb (approved):
Thanks for fixing this!

PR #606, opened 2025-10-21
Support OCaml 5.5

patricoferris:


NathanReb (approved):
Thanks!

NathanReb:
We need to bump the upper bound in our dune-project/opam files to get a bit of feedback from the CI.

We'll also need to update `ast/supported_version/supported_version.ml`, `astlib/config/gen.ml` and to add identity migrations and the relevant `ast_505.ml`.

I'm happy include these extra steps in a separate PR.

NathanReb (approved):
Looks good to me! Still curious to hear your feedback on the gencopy workflow to see if we can improve it somehow.

I guess one thing we could change is that it should probably dump a comment at the beginning of the file, hinting that it was initially generated with that tool and later modified. Having that comment point to the dev/README.md would help maintainers discover it if they're not aware of its existence.

PR #607, opened 2025-10-23
Add support for 5.4 labeled tuples

NathanReb:
This adds support for 5.4 labeled tuples and comes with different things:
1. Downward migrations no longer fail and instead encode labeled tuples into an extension point
2. Such extension points are detected when migrating from 5.3 to 5.4 and are correctly translated into their rightful AST representation.
3. `Ast_builder` and `Ast_pattern` have 3 new functions each to build/destruct encoded labeled tuple types, expressions and patterns.

The actual handling of the extension point is done in a single module in astlib: `Encoding_504`. It provides constructor/destructor based on the 5.3 AST (used in the migration code) and on the 5.2 AST (used in Ast_builder/Ast_pattern).
To make sure they don't desync in the future, I functorized the encoding/decoding. This required writing quite a bit of boiler plate and therefore would only advise this when we actually want to provide a stable API to handle such encoded nodes and would not bother otherwise.
E.g., when adding support for `Bivariant` type parameters, I will only provide encoding/decoding functions based on the 5.3 AST to be used in the migration code.

### DIFF
--- a/astlib/encoding_504.ml
+++ b/astlib/encoding_504.ml
@@ -1,6 +1,7 @@
 module Ext_name = struct
   let ptyp_labeled_tuple = "ppxlib.migration.ptyp_labeled_tuple_504"
   let pexp_labeled_tuple = "ppxlib.migration.pexp_labeled_tuple_504"
+  let ppat_labeled_tuple = "ppxlib.migration.ppat_labeled_tuple_504"
 end
 
 let invalid_encoding ~loc name =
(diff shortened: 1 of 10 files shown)


patricoferris:
Thanks Nathan, even that small example would be great in some documentation. Given this is, until our next major release bump, supposed to be used by authors I think we need to make it easy for ppx authors to onboard :)

Issue #610, opened 2025-10-27
Provide Mod.apply_unit


craff:
Currently it triggers a warning, from 5.03 apply_unit must be used and there could be a fallback for older version.

PR #611, opened 2025-10-27
Add Mod.apply_unit

craff:
Solves #610. Just modified ast_helperlite.mli? And tested on my code.

### DIFF
--- a/.ocamlformat-ignore
+++ b/.ocamlformat-ignore
@@ -47,6 +47,7 @@ test/driver/non-compressible-suffix/test.ml
 test/driver/transformations/test.ml
 test/driver/transformations/test_412.ml
 test/driver/transformations/test_510.ml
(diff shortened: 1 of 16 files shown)


patricoferris:
Hi @craff, thanks for this adding this. 

Is there any reason you are not using the functions supplied by `Ast_builder`, in this case I think it would be`pmod_apply_unit`?

craff:
I made an attemp to use Ast_builder and gave up (to much work). Here is a list of painful things:

- no optional argument, for ~loc but many other. For instance Const.string s to be replaced by Pconst_string s with 3 arguments. 
- Mod. syntax is shorter an allow for local open
- Vb module is also nive as it groups many functions. 
- ...

craff:
I also rebased the PR, but it makes commit appear as part of the PR ?

PR #612, opened 2025-11-25
Remove myself (pitag-ha) as a GH code owner

pitag-ha:
Heya, as I'm not active on ppxlib anymore, I'm removing myself as a GH code owner, so that I won't automatically be added as a reviewer by GH anymore. I think this is the only place where I was listed as a maintainer. If there are other places, please let me know.

Btw, IMO, we should edit CODE_OF_CONDUCT.md to ask folks to reach out to you, @NathanReb and @patricoferris instead of @panglesd, @ceastlund  and me in the very unlikely event that there's a code of conduct violation to report. Does that sound good to you?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @ceastlund @NathanReb @panglesd @patricoferris @pitag-ha
+* @ceastlund @NathanReb @panglesd @patricoferris


patricoferris:
Yes! Thank you Sonja, @panglesd would you also like to be removed from the CODEOWNERS file? 

Re:code-of-conduct -- I think that makes sense to me, I can open a separate PR for that and we can discuss there.

Issue #618, opened 2026-01-07
missing some compiler error locations when using Dune + ppxlib + OCaml 5.4


gasche:
There is a regression in the interaction between Dune, ppxlib and OCaml 5.4.0, which results in some compiler errors being reported with missing location information. This is fairly disruptive in practice (as it breaks "jump-to-error" workflows.)

I believe that the issue is the one discussed in https://github.com/ocaml/ocaml/pull/12991, but this compiler-side issue is fairly difficult to find so I thought that opening another issue where users might look for them could be helpful.

### Reproduction steps

In an empty temporary directory:

```
echo > a.mli
echo "let () = NonExistingModule.run ()" > a.ml
echo "(lang dune 3.0)" > dune-project
echo "(library (name test) (preprocess (pps ppx_here)))" > dune
dune build
```

Observed output:

```
File "_none_", line 1:              
Error: Unbound module NonExistingModule
```

Expected output:

```
File "a.ml", line 1:              
Error: Unbound module NonExistingModule
```

(This works properly with OCaml 5.3 and older compiler versions.)

anmonteiro:
Thanks for confirming, that sounds like a different issue, then.

PR #619, opened 2026-01-10
Fix longident locs

patricoferris:
Previously, in our migration from the 503 AST to the 504 AST, we were adding `Location.none` locations to all of the segments of any long identifiers. This PR starts to try to fix that by reconstructing the locations of `Ldot` segments. 

Partially fixes: #618

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,7 +5,8 @@ unreleased
   an error instead (#613, @ceastlund, @patricoferris)
 - Ignore extensions inside attributes for the unused extension check
   (#616, @Skepfyr)
-- Fix Longident locations for 504 and above (#619, @patricoferris)
+- Fix a bug that inserted `Location.none` into `Longident`s when using OCaml
+  5.4 and above (#619, @patricoferris)
 
 0.37.0
 ------
--- a/test/504_migrations/longident-locs/run.t
+++ b/test/504_migrations/longident-locs/run.t
@@ -10,7 +10,7 @@ If we run the driver on the following source file:
 
 then the non-existing module should have a sensible error location.
 
-  $ ocamlc -ppx "./driver.exe --as-ppx" test.ml test.ml.pp
+  $ ocamlc -ppx "./driver.exe --as-ppx -locations-check" test.ml test.ml.pp
   File "test.ml", line 1, characters 9-30:
   1 | let () = NonExistingModule.foo ()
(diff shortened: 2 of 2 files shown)


patricoferris:
Thanks for the ping @gasche, I'll take a look this week, but there are two main issues why this PR is not undrafted.

1. We are not doing anything smart for reconstructing the locations of `Lapply`s -- unfortunately I don't think there is much we can do about that given we have no way of knowing in the original source code how much white space there could be (e.g. `type t = int Map.Make     (String).t` is valid).
2. In the case where we migrate downwards we could in theory stash the true locations and reapply them in the migration back up, but I'm not sure that is really worth the hassle.

I'll need to take a closer look, for example, I just discovered that the following is broken:

```ocaml
module F (A : sig type t end) = struct
  module G (B : sig type t end) = struct
    type t = A.t * B.t
  end
end

type t = { v : F(Option).G(Int).t }
```

Which produces:

```
File "example/main.ml", line 7, characters 15-29:
7 | type t = { v : F(Option).G(Int).t }
                   ^^^^^^^^^^^^^^
Error: Modules do not match: (module Stdlib__Option) is not included in
       sig type t end
     ...
     Type declarations do not match:
       type 'a t = 'a option = None | Some of 'a
     is not included in
       type t
     They have different arities.

```

giltho:
Naive question because I don't think I understand  how AST migration is done internally but if ppxlib doesn't bump its internal AST anytime soon, how will modular explicits be supported in 5.5?

giltho:
I meant the first one! (Though this is mostly out of curiosity, thank you for taking the time 😊)

> We will encode modular explicits into the AST which will preserve them as code is migrated downwards

Is the part of the AST that contains the modular implicit kept opaque? That means I wouldn't be able to use ppxes in "new" AST nodes I guess. Though, given it reuses first-class module syntax maybe that's not an issue

NathanReb:
I think cutting a release from the main branch would be mostly okay. We can add 5.4 and 5.5. full support incrementally thanks to our new approach. I agree a single 5.4 support release would have been best but I think it's fine, especially since what's left, if I'm not mistaken is the bivariance stuff, which most people will never run into.

patricoferris:
@giltho this is now released in ppxlib.0.38.0 :)

gasche:
Excellent news, thanks!

giltho:
Thank you very much!!!

PR #634, opened 2026-04-03
Use dev version of dune for trunk build

NathanReb:


NathanReb:
How naive of me, setup-ocaml handles dune installation for some reason.

patricoferris:
> How naive of me, setup-ocaml handles dune installation for some reason.

Argh, that's annoying. I couldn't find an options for setup-ocaml to workaround this without us having to pin dune explicitly in our own opam files.

NathanReb:
Yeah, not quite sure when the compatible release of dune is scheduled. Might be worth patching setup-ocaml as this could prove useful to other projects as well!

NathanReb:
This is not required anymore! Closing!

PR #640, opened 2026-04-22
Better migration errors

patricoferris:
The new error messages provide more context and make sure to mention that ppxlib raised the error. My `Format`-knowledge is not great so perhaps there is something better than `pp_print_text`. Also, very happy to discuss what the "better error messages" should be @NathanReb !

Part of improving #626

### DIFF
--- a/astlib/error.ml
+++ b/astlib/error.ml
@@ -1,0 +1,24 @@
+let raise_error_txt ~loc = Location.raise_errorf ~loc "%a" Format.pp_print_text
+
+let migration_error ~loc ~from ~to_ msg =
+  let s =
+    Format.sprintf
+      "ppxlib migration error (migrating ocaml.%s to ocaml.%s): %s\n\n\
+       Ppxlib was likely trying to migrate from %s to 5.2.0 and back again, \
+       but encountered an AST node that could not be encoded. Either remove \
+       the feature or open an issue at \
+       https://github.com/ocaml-ppx/ppxlib/issues."
+      from to_ msg Sys.ocaml_version
+  in
+  raise_error_txt ~loc s
+
+let invalid_encoding ~loc ~version msg =
+  let s =
+    Format.sprintf
+      "ppxlib invalid encoding: %s\n\n\
+       Ppxlib failed to decode a feature from the OCaml %s AST. If this does \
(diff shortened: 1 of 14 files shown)


astlib/error.ml
```diff
@@ -0,0 +1,24 @@
...
+      from to_ msg Sys.ocaml_version
+  in
+  raise_error_txt ~loc s
+
+let invalid_encoding ~loc ~version msg =
+  let s =
+    Format.sprintf
+      "ppxlib invalid encoding: %s\n\n\
+       Ppxlib failed to decode a feature from the OCaml %s AST. If this does \
+       not seem right, please do open an issue at \
```
NathanReb: How about also hinting at an interaction between the language feature and a ppx here?
Something along those lines:
```
This is likely caused by the use of a 5.3+ language feature within code directly interpreted by a ppx.
```

NathanReb: Maybe even add something like:
```
If you'd like this feature to be supported by the ppx involved, please reach out to the ppx maintainers.
If this does not seem right, please do open an issue at...
```

NathanReb (commented):
Looks good to me! Only have a minor suggestions on the decoding failure message.

The formatting looks good enough to me, we can always improve it later on.

astlib/error.ml
patricoferris: Added both in https://github.com/ocaml-ppx/ppxlib/pull/640/commits/f7aea0815af5188e826de63d14734985ca9ec70a

NathanReb (approved):
Looks good to me, thanks!

NathanReb:
Argh, ocaml-ci not showing up here either!

NathanReb:
https://ocaml.ci.dev/github/ocaml-ppx/ppxlib/commit/f7aea0815af5188e826de63d14734985ca9ec70a It seems some tests on 5.4+ were not updated!

patricoferris:
Thanks @NathanReb! New CI run at: https://ocaml.ci.dev/github/ocaml-ppx/ppxlib/commit/ef0e7a3d35d47d7deae20411a938c6460280fd8b

NathanReb:
Everything except risk builds is green, let's merge!

PR #641, opened 2026-05-06
remove stdlib-shims

copy:
stdlib-shims is only needed for OCaml <= 4.07, but the minimum version supported by ppxlib is already 4.08.

Tested on a 4.08.1 and a 5.4.1 switch.

### DIFF
--- a/ast/dune
+++ b/ast/dune
@@ -5,7 +5,7 @@
 (library
  (name ppxlib_ast)
  (public_name ppxlib.ast)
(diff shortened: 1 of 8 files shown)


NathanReb:
Thanks for the PR! You'll need to sign off commits to make the DCO happy!

NathanReb (approved):
Looks good to me! Good catch and thanks for the fix!

PR #642, opened 2026-05-06
Update 5.6 support: primitive alias

NathanReb:
This PR adds support for 5.6 primitive alias feature.

In 5.6 it is now possible to define an alias on a primitive defined elsewhere. Primitives are now represented differently in the AST so it needs to be properly migrated in between 5.6 and 5.5 and for primitive aliases, properly encoded/decoded.

### DIFF
--- a/astlib/migrate_505_506.ml
+++ b/astlib/migrate_505_506.ml
@@ -545,13 +545,18 @@ and copy_value_description :
        Ast_505.Parsetree.pval_attributes;
        Ast_505.Parsetree.pval_loc;
      } ->
-  {
-    Ast_506.Parsetree.pval_name = copy_loc (fun x -> x) pval_name;
-    Ast_506.Parsetree.pval_type = copy_core_type pval_type;
-    Ast_506.Parsetree.pval_prim = List.map (fun x -> x) pval_prim;
-    Ast_506.Parsetree.pval_attributes = copy_attributes pval_attributes;
-    Ast_506.Parsetree.pval_loc = copy_location pval_loc;
-  }
+  match pval_prim with
+  | [] ->
+      {
+        Ast_506.Parsetree.pval_name = copy_loc (fun x -> x) pval_name;
+        Ast_506.Parsetree.pval_type = copy_core_type pval_type;
+        Ast_506.Parsetree.pval_attributes = copy_attributes pval_attributes;
+        Ast_506.Parsetree.pval_loc = copy_location pval_loc;
+      }
+  | _ ->
+      Location.raise_errorf ~loc:pval_loc
+        "Ppxlib migration error: value_description with pval_prim <> [] cannot \
+         be migrated from Ocaml 5.5 to 5.6"
 
 and copy_type_declaration :
     Ast_505.Parsetree.type_declaration -> Ast_506.Parsetree.type_declaration =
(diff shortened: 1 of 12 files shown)


NathanReb:
I'm opening as draft for now as it lacks the following:
1. encoding/decoding of primitive aliases
2. tests

In its current state though it builds on 5.6 and will correctly handle code that does not use primitive aliases so it might prove useful to people that want to give trunk a shot.

I'll add full support and tests in the next few days!

NathanReb:
I'm also hoping to see the trunk build finally go green!

NathanReb:
hmm weird, can't see the ocaml-ci build!

If it's too unstable we'll have to add more reliable github actions build I guess.

astlib/migrate_505_506.ml
```diff
@@ -545,13 +545,18 @@ and copy_value_description :
...
+  match pval_prim with
+  | [] ->
+      {
+        Ast_506.Parsetree.pval_name = copy_loc (fun x -> x) pval_name;
+        Ast_506.Parsetree.pval_type = copy_core_type pval_type;
+        Ast_506.Parsetree.pval_attributes = copy_attributes pval_attributes;
+        Ast_506.Parsetree.pval_loc = copy_location pval_loc;
+      }
+  | _ ->
+      Location.raise_errorf ~loc:pval_loc
```
patricoferris: I'll incorporate your suggestions in #640 and then we can use that here?

test/506_migrations/run.t
```diff
@@ -0,0 +1,38 @@
...
+  > EOF
+
+Those should be correctly represented in 5.6- :
+(This command won't show anything interesting once our AST is >= 5.6)
+
+  $ ./driver.exe test.mli
+  val x : int -> int
+  external y : int -> int = "some_c_fun"
+  $ ./driver.exe test.ml
+  external x : int -> int
```
patricoferris, line 28: Are value descriptions allowed in structures? They seem to parse ok, but later they raise an error.

Reading this code, I had a question: should we tag all migrations that change a node with some attribute indicating that it is a ppxlib migration? Even if it is not an encoding? 

I guess my concern here is that in 5.6 we have code that doesn't compile. But in 5.5 we do, and there's no mention of ppxlib in the source. What do you think?

patricoferris (approved):
Thanks for this Nathan, and sorry about the slow review. This look pretty good to me, just a question about how to handle migrations that don't need to encode but might change the meaning of the generated code.

test/506_migrations/run.t
NathanReb: > Are value descriptions allowed in structures?

Yes, only at parsing. I think it was meant for ppx-es to interpret those, if they aren't rewritten, you'll get a typing error.

> should we tag all migrations that change a node with some attribute indicating that it is a ppxlib migration? Even if it is not an encoding?

I think so yes, attributes we use to ensure round-tripping are inserted and later consumed by migrations so I think it make sense.

> I guess my concern here is that in 5.6 we have code that doesn't compile. But in 5.5 we do, and there's no mention of ppxlib in the source. What do you think?

Sorry I didn't quite get what you meant here, can you give a concrete example?

astlib/migrate_505_506.ml
NathanReb: Yes we should use the common migration error function but I must mention this is a bit of corner case here.

This function is never called when `pval_prim = []` because it is checked in `copy_structure_item_desc` and `copy_signature_item_desc`. The only situation where this could be triggered is if someone manually called this specific `copy_value_description`.

test/506_migrations/run.t
NathanReb: > Reading this code, I had a question: should we tag all migrations that change a node with some attribute indicating that it is a ppxlib migration? Even if it is not an encoding?

I think I originally misread this part. What happens here isn't really a change in the node but rather an oddity of this specific node (i.e. value description in .ml files that aren't primitives). These nodes exist solely for ppx-es to interpret and are simply not correctly printed out by `Pprintast` which always prints `Pstr_primitive vd` as `external ...` without checking whether `vd.pval_prim = []`.

You can actually spot that because there's no `= "<external-fun-name>"` at the end of the primitive description, which there should be were it an actual primitive.
In 5.6, this printing bug is gone because there's now a specific `Pstr_val of value_description` variant to represent it.

PR #643, opened 2026-05-21
Fix `Ast_builder.Default.pexp_apply` duplicating attributes

Skepfyr:
Fix a simple bug in pexp_apply where it would use the wrong expression as the base meaning it would duplicate attributes and get the locations wrong.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,9 @@
 unreleased
 ----------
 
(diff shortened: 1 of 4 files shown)


NathanReb (commented):
Thanks for the fix! Could you add a changelog entry?

NathanReb (approved):

Skepfyr:
Thanks!

PR #644, opened 2026-06-21
Remove extra copy_module_type from pmty_with migration from 504 to 503

smuenzel:
In `astlib/migrate_504_503.ml`, when migrating `Pmty_with`, the module type is copied twice. 

This can be especially bad with many nested `with` constraints, such as https://github.com/janestreet/base/blob/259754a4a8601a1c59a9dcf2baf008e2a90d54fd/src/import0.ml#L5-L42, seemingly leading to a double copying at every constraint.

This error was found with claude code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,9 @@ unreleased
 
 - Add support for OCaml 5.6 primitive aliases (#642, @NathanReb)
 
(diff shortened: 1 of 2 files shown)


NathanReb (approved):
Thanks for the fix!

PR #647, opened 2026-07-12
Fix typos.

ScythingSt:
There are 2 typos in the documentation file [examples.mld](https://github.com/ocaml-ppx/ppxlib/blob/c6bb0d5bacfd05efec942b6c793017663f69f466/doc/examples.mld#L9).

```code
ppx_deriving_accesors  (old)
ppx_deriving_accessors (new)
```

### DIFF
--- a/doc/examples.mld
+++ b/doc/examples.mld
@@ -6,9 +6,9 @@ This section is here to allow viewing complete examples of PPXs written using [p
 
 In order to see a fully working complete example of a PPX written using [ppxlib], that you can compile, modify and test, go to the {{:https://github.com/ocaml-ppx/ppxlib/tree/main/examples}examples} folder of ppxlib sources.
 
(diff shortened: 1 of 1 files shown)


NathanReb (approved):
Thanks for the fix!

NathanReb:
You'll need to add a **Signed-off-by** trailer to make the DCO check happy! You can do it with a simple `git rebase -s main`

ScythingSt:
Thank you for your suggestion!

I just noticed that even though I did sign off my real name manually in the previous commit, DCO requires me to use the same name as for Git, which was a pseudonym. Now everything should be correct.

NathanReb:
Ah my bad, didn't take the time to check whether the commits were actually signed off. Anyway, thanks for fixing this, merging!

PR #650, opened 2026-09-07
Fix migration of 5.4's `Pexp_pack`

dra27-js:
In 5.4, https://github.com/ocaml/ocaml/pull/13809 effectively changed where attributes appear on package types, and the migration back to 5.3 drops them:

```ocaml
let () =
  List.iter
    (fun source ->
       let structure = Ppxlib.Parse.implementation (Lexing.from_string source) in
       Format.printf "in:  %s@." source;
       Format.printf "out: %a@.@." Ppxlib.Pprintast.structure structure)
    [ "let f (module K : S with type t = int [@foo]) = ()"
    ; "let g (x : int [@foo]) = ()"
    ]
```

gives:

```
in:  let f (module K : S with type t = int [@foo]) = ()
out: let f ((module K)  : (module S with type t = int)) = ()

in:  let g (x : int [@foo]) = ()
out: let g (x : ((int)[@foo ])) = ()
```

for OCaml 5.4/5.5 before this patch and the correct `out: let f ((module K)  : (((module S with type t = int))[@foo ])) = ()` with it.


patricoferris:
Thanks for the report @dra27-js -- taking a look :-)

patricoferris:
Hmm we do drop the package attributes here:

https://github.com/ocaml-ppx/ppxlib/blob/b9646cc521d0b12f3ade492fcd06d979e27c8f98/astlib/migrate_504_503.ml#L460-L468

And your fix re-adds them in the two places where a `package_type` might be used: a `Pexp_unpack` and a `Ptyp_package`. 

It would be good to add a quick test case for this, something like this in `test/504_migrations`. 

```
+ Check package_type attributes roundtrip:
+
+  $ cat > test.ml << EOF
+  > module type S = sig type t end
+  > let f (module K : S with type t = int [@foo]) = ()
+  > let f (module K : S with type t = (int [@foo])) = ()
+  > type t = (module S [@foo])
+  >
+  > EOF
+
+  $ ./driver.exe test.ml
+  module type S  = sig type t end
+  let f ((module K)  : (((module S with type t = int))[@foo ])) = ()
+  let f ((module K)  : (module S with type t = ((int)[@foo ]))) = ()
+  type t = (((module S))[@foo ])
+  $ ./driver.exe --use-compiler-pp test.ml
+  module type S  = sig type t end
+  let f ((module K)  : (((module S with type t = int))[@foo ])) = ()
+  let f ((module K)  : (module S with type t = ((int)[@foo ]))) = ()
+  type t = (((module S))[@foo ])
```

Happy to do that for you if you like @dra27-js ?

NathanReb:
It would be good to ensure the attributes are correctly preserved through round trips as well! Happy to take a look at that if needed!

dra27-js:
Sure - feel free to push this branch (or I can have a look later in the week)

astlib/migrate_504_503.ml
```diff
@@ -403,11 +403,18 @@ and copy_core_type : Ast_504.Parsetree.core_type -> Ast_503.Parsetree.core_type
...
+    | Ast_504.Parsetree.Ptyp_package { Ast_504.Parsetree.ppt_attrs; _ } ->
+        copy_attributes ppt_attrs
+    | _ -> []
+  in
   {
     Ast_503.Parsetree.ptyp_desc = copy_core_type_desc ~loc ptyp_desc;
     Ast_503.Parsetree.ptyp_loc = loc;
     Ast_503.Parsetree.ptyp_loc_stack = copy_location_stack ptyp_loc_stack;
-    Ast_503.Parsetree.ptyp_attributes = copy_attributes ptyp_attributes;
+    Ast_503.Parsetree.ptyp_attributes =
```
patricoferris, line 416: I'm concerned that this doesn't roundtrip very well, and this is a fact of not being able to unmix the core type attributes from the package type attributes. For example:

```ocaml
type t = (module S [@inner])[@outer]
```

Has the following (summarised for clarity), AST on OCaml 5.4:

```
type_declaration "t" 
  ptype_manifest =
    Some
      core_type
        attribute "outer"
          []
        Ptyp_package
          package_type "S"
          []
            attribute "inner"
              []
```

When we pass it through ppxlib (using the fixes here) we get `type t = (((module S))[@outer ][@inner ])` with the following AST:

```
type_declaration "t"
  ptype_manifest =
    Some
      core_type
        attribute "outer"
          []
        attribute "inner"
          []
        Ptyp_package "S"
        []
```

Which is better than without the fix where we were just dropping attributes! But maybe is not ideal. I wonder if we shouldn't have an encoding for package types _with_ attributes that we can use instead in order to safely preserve these attributes when we migrate back from 5.3 to 5.4 (which should be doable via `Ptyp_extension` and `Pexp_extension`)? What do you think @NathanReb?

I have added some tests in https://github.com/ocaml-ppx/ppxlib/pull/650/commits/ed03c4dfa22242f3bc63dba7685183781d64ba89 @dra27-js.

NathanReb: I would avoid wrapping in an extension here since there's a valid migration (modulo attributes). If we do encode those with an extension, that means than building and preprocessing old code with new compiler won't work, which is what we're trying to avoid.

It's not the most statisfying of ways but could we split the attributes from the types and the attributes from package by inserting our own `[@ppxlib.migration.split_package_attributes]`?

We just add an extra attribute that everyone will ignore and when migrating back to 5.4, the migration knows to treat all attributes after this one differently.

Does that sound doable?

patricoferris: That sounds good to me @NathanReb -- if you're happy enough I can push a commit to that effect @dra27-js.

dra27-js: Go for it - I can double-check again, but we're not actually hitting this particular case anymore (I can possibly check against an older tree if it would helpful to test it against "real" code that's hitting it?)

patricoferris: https://github.com/ocaml-ppx/ppxlib/pull/650/commits/4ec408f00738254a191315e5098fd54e02e54f26 contains the "splitting" approach @NathanReb and the tests now roundtrip correctly.

astlib/migrate_504_503.ml
NathanReb, line 429: I think it would improve readability to have the split attribute inserted here:
```ocaml
copy_attributes ptyp_attributes @ (split_package_attr::package_attrs);
```

What do you think?

NathanReb (commented):
Can you add a changelog entry ?

Besides that, I only have one minor style comment, otherwise this looks good to me!

Thanks for taking care of this @dra27-js @patricoferris.

astlib/migrate_504_503.ml
patricoferris: I agree in principle, however, I was trying to make it so we only added the split attribute whenever there are actually _some_ package attributes? The logic should be able to deal with having a superfluous split attribute so there would likely be no harm in adding it.

Would a comment suffice here?

patricoferris:
> Can you add a changelog entry ?

Hmm it seems @dra27 already did and the PR CI is failing for other reasons: https://github.com/ocaml-ppx/ppxlib/actions/runs/34973743104/job/104396127893?pr=650!

astlib/migrate_504_503.ml
NathanReb: No you're right, let's keep it as it is!

NathanReb:
My bad, indeed there's already a changelog entry, I got tricked by the commit view I was sent to from my notifications.

That other check keeps failing and it's not that useful so I'll simply remove it. The changelog check itself is green.

NathanReb (approved):

patricoferris:
We'll have to wait on the dream of an "all green" CI @NathanReb... FreeBSD isn't happy, but seems like an upstream issue :shrug:

dra27-js:
Thank you both very much for the work on this!

Issue #633, opened 2026-03-26
`Ast_pattern` fails on attributes


n-osborne:
Inspecting a construction containing a legal attribute fails with:

`Fatal error: exception Attributes not allowed here`

Reproduction case:

```ocaml
open Ppxlib

let ___bug =
  let loc = Location.none in
  let ast = [%expr x[@bug]]
  and prop _ = true and on_error () = false
  and pat = Ast_pattern.(pexp_ident __)
  in
  Ast_pattern.parse pat loc ast ~on_error prop

let () = if ___bug then print_endline "OK" else print_endline "KO"
```

NathanReb:
Thanks for reporting this!

I'm a bit puzzled here, it seems that since the start of our github history, the generated patterns reject anything that has attributes.

I'd have to check but I assume the logic is that it's too easy to miss on important attributes when using `Ast_pattern` and that its use should be reserved for "controlled space" such as the payload of a ppx's extension, where you know exactly what attributes are present or not.

That being said I agree this is not an acceptable scenario on multiple accounts:
- It prevents users from writing more stable code using `Ast_pattern` in a lot of scenarios
- The failure mode is disastrous and can be extremely misleading to the end user (as opposed to the ppx author)
- We should provide variants that allow attributes and eventually that even allow you to capture/pattern match over them
- It seems a bit redundant with the unused attribute check that seems

I'll reached out to the original authors of ppxlib to make sure I have a complete understanding of the situation before deciding how to proceed here but this definitely needs some sort of fix!

Out of curiosity, can you describe in which context you encountered this bug? Why are you using `Ast_pattern`? Are the attributes here some you mean to consume yourself or are they unrelated to your ppx/program?

n-osborne:
> Out of curiosity, can you describe in which context you encountered this bug? Why are you using Ast_pattern? Are the attributes here some you mean to consume yourself or are they unrelated to your ppx/program?

It is in the context of a prototype for code instrumentation and using `Ast_pattern` here is for demonstration purpose. So that's pretty artificial I guess. Also I'm not at liberty to show the code.

The story is as follow:

- I have to instrument differently an expression depending on it containing an item of a certain type (think an identifier eg)
- A previous phase of the instrumentation could have introduced such an item
- But those should be ignored (they are not present in the original code on which the instrumentation should be based)
- I decide to watermark the items the instrumentation introduce with an attribute.
- The bug

[...]

PR #651, opened 2026-09-15
Migrate 5.06 Pexp_hole and Pmod_hole

patricoferris:
A migration PR for https://github.com/ocaml/ocaml/pull/15051.

I went for the slightly more complicated encoding scheme thinking we would expose holes via `Ast_pattern` at the very least, but I'm not sure how useful that would be in practice.

cc @NathanReb -- any ideas about how we could expose these nodes for ppx rewriter authors?


NathanReb:
This looks good, thanks! I agree we need at least an `Ast_pattern` interface here since these nodes are explicitly meant to be interpreted by ppx-es.

I think exposing the following in `Ast_pattern` would be enough here:
```ocaml
val pexp_hole : (expression, 'a, 'a) t
val pmod_hole : (module_expr, 'a, 'a) t
```

This allows ppx authors to detect holes via a forward compatible API (if we bump to 5.6+ AST, these will remain, they simply won't have to decode anything). Does that sound good to you?

NathanReb:
We can eventually add the `Ast_builder` counterpart on demand, these could eventually be useful in some advanced ppx interactions where one generates holes for another to interpret.

As I'm writing this I'm wondering whether we should add them straight away and be done with it, since you already wrote the functorized encoding/decoding logic, we might as well.

patricoferris:
Yep, I completely agree. Added in https://github.com/ocaml-ppx/ppxlib/pull/651/commits/8185b9aa4e468733deddfea9d2106e628475bec7.

NathanReb:
Might be worth looking into #633 again as I can definitely imagine ppx-es expecting wholes to have attributes attached that guides the expansion.

With our current API it would be a little tedious to deal with this as you'd have to strip the attributes before checking whether it is a hole.

test/encoding/5.6/api/dune
```diff
@@ -0,0 +1,14 @@
+(rule
```
NathanReb, line 1: There's already a `test/encoding/5.6` folder where you could add `api/`!

I think you could also move the hole migration test in `test/encoding/5.6/migrations/run.t` though I admit splitting migration tests between the natural migration and the encoding does necessarily make that much sense.

src/ast_builder.ml
```diff
@@ -288,6 +288,14 @@ module Default = struct
     in
     { ppat_desc; ppat_loc = loc; ppat_attributes = []; ppat_loc_stack = [] }
 
+  let pexp_hole ~loc =
```
NathanReb: should we add a `()` argument to allow for ulterior addition of optional arguments?

NathanReb (commented):
This looks good to me! I just have a couple small remarks + it needs a changelog entry!

Thanks for taking care of it!

According @Octachron this should be the last AST change to make it to 5.6 so we could potentially cut a 5.6 compatible release once this is merged!

test/encoding/5.6/api/dune
patricoferris: Ha! I was just pattern-matching on the `50X` folders :facepalm: -- thanks :-)

Octachron:
@NathanReb , the include functor changes could still make the cut before the feature freeze, so it would probably be still more prudent to wait for the freeze (that will happen soon).

NathanReb (approved):

